PR #3, opened 2015-06-06
Construct BUILD_DATE using date -u for OSX support

jbye:
I am unable to build shards on OSX due to `date --utc` being unavailable on OSX. Changed `date --utc` to `date -u` to get it to build (which should also work on linux)


ysbaddaden:
Thanks. I always prefer long parameter names over single chars for readability, but it looks like single chars are more portable.

Issue #15, opened 2015-07-22
Add support for c libraries.


technorama:
One possible way to add c library support:
1.  A build command (or list of commands) in shard.yml
2.  State tracking and output capture of commands for debugging (similar to rubygems).
3.  Additional library path injection to the crystal build process.  shards.yml could contain additional c library paths.  Minor modification of crystal may be necessary to read the list of additional library paths.

ysbaddaden:
Could you detail your problem and use case before jumping to potential solutions? I'd like to understand what's actually needed beforehand.

It may involve a bunch of problems, like checking for the presence of a library, downloading, compiling, making the lib available to Crystal when linking, etc.

technorama:
Use case #1:

I want to use an existing library but need a few C functions added.   As far as I know the only way to use them is by compiling a library and linking it using the crystal provided c bindings.


Use case #2 (similar to above):

I have a small library in my git repository that must be compiled before crystal build is run.  Crystal c bindings are used to access the library.

kostya:
what if always call `make shard` when install new shard, and if user create that task in Makefile, it would be executed (compile some static c libs or something)

ysbaddaden:
Thanks for the use cases. We could have some hook commands to be run when an action is performed. For example:

```yml
scripts:
  postinstall: "cd ext && make"
```

The most obvious is `postinstall` but there could be others, like pre/post upgrade.

ysbaddaden:
@technorama I just implemented the postinstall script in #19, tell me if it fits your needs.

ysbaddaden:
Closing since #19 got merged.

hotty316:
> One possible way to add c library support:
> 1.  A build command (or list of commands) in shard.yml
> 2.  State tracking and output capture of commands for debugging (similar to rubygems).
> 3.  Additional library path injection to the crystal build process.  shards.yml could contain additional c library paths.  Minor modification of crystal may be necessary to read the list of additional library paths.
>

PR #19, opened 2015-08-16
Run postinstall script after installing dependency

ysbaddaden:
Dependencies may now have a post install script hook that will be
run whenever the dependency is installed or updated:

    - scripts:
      - postinstall: "cd ext && make"

refs #15


ysbaddaden:
The postintall script is only run when installing the dependency, and it won't be run for the current project.

PR #21, opened 2015-08-26
use new process overload in GitResolver#run

### LINKED ISSUE #20: fails to build with crystal master
Building with latest crystal (from git master) fails due to changed Process::run overloads.

``` ruby
in ./src/resolvers/git.cr:172: no overload matches 'Process::run' with types String
Overloads are:
 - Process::run(cmd : String, args = nil, env = nil : Env, clear_env = false : Bool, shell = false : Bool, input = false : Stdio, output = fal
se : Stdio, error = false : Stdio, chdir = nil : String | ::Nil)
 - Process::run(cmd : String, args = nil, env = nil : Env, clear_env = false : Bool, shell = false : Bool, input = nil : Stdio, output = nil :
 Stdio, error = nil : Stdio, chdir = nil : String | ::Nil, &block)

        status = Process.run("/bin/sh", input: command, output: capture)
                         ^~~
```

jreinert:
fixes #20

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -169,10 +169,15 @@ module Shards
 
       Dir.chdir(path) do
         Shards.logger.debug command
-        status = Process.run("/bin/sh", input: command, output: capture)
+        output = nil
+        error = nil
+        Process.run("/bin/sh", ["-c", command]) do |process|
+          output = process.output.read
+          error = process.error.read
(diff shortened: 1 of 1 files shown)


ysbaddaden:
Thanks! But a bunch more things got broken too and I wanted to avoid the `$?` global, so I fixed everything in #22

PR #27, opened 2015-09-04
Lock file implementation

### LINKED ISSUE #12: Lock dependencies
Shards is missing a mechanism to lock the state of installed dependencies, so installs could be repeatable, always installing the same exact set of dependencies across developer machines, CI, ...

The format of the lock file is ~~undecided. I think it should contain~~ the flattened list of dependencies, along with their Git URL and either the version or Git commit (if a git tag/commit/branch refs is specified).

The lock file will impact the install command which must abide to the locked versions/refs and do its best to install new dependencies that match. The lock file may be seen as an initial set of requirements for dependencies —yet not be injected as dependencies, which could install removed dependencies.

The update command must not be impacted, but must generate the lock file.

ysbaddaden:
- [x] install command:
  - [x] load lock file before install command
  - [x] enforce locked versions
  - [x] eventually generate lock file (if missing or outdated)
- [x] update command:
  - [x] eventually generate lock file
- [x] versioned lock file for future expansions / refactors
- [x] integration tests
- [x] drop custom groups
- [x] add `--production` parameter to skip development dependencies 

closes #12 #31


ysbaddaden:
It looks like this is requiring Crystal HEAD.

src/commands/install.cr
```diff
@@ -1,31 +1,80 @@
...
-            Shards.logger.info "Using #{package.name} (#{package.version})"
+        if locks = @locks
+          install(manager.packages, locks)
+        else
+          install(manager.packages)
+        end
+
+        unless lock_file?
+          File.open(lock_file_path, "w") { |file| manager.to_lock(file) }
+        end
```
ysbaddaden: The install command must regenerate the lock file in case any new package was installed.

ysbaddaden:
Sadly, custom groups are behaving badly with the lock file.

Either we have to resolve everything, which means cloning every dependency (whatever group we want to install); or the lock file lists the shards for the selected groups only, which results in a bad lock file (somehow truncated).

I think I'll remove the ability to have and select custom groups and keep the development group only, until we can figure out how to properly deal with it.

jhass:
Yeah, bundler can only do it since it can fetch the whole graph via rubygems.org's dependency API.

PR #30, opened 2015-09-09
log commands that are executed

jessedoyle:
It's probably a good idea to log any executed commands so that the user can verify commands.

This pull request simply logs the command as such:

```
Postinstall cd ext && make
```

This patch also fixes a small issue when run without a spec file. Previously the word `Error` was displayed twice (I submitted a PR for this earlier, but the error logging behaviour has changed since then).


ysbaddaden:
Very good idea, indeed. Thanks!

PR #54, opened 2015-09-26
Fix a link

sstarr:
The link at the top of the README pointed at `https://crystal-lang.org` but the site isn't available on port 443 so it was timing out.


ysbaddaden:
Thanks!

PR #63, opened 2015-10-16
Updated to Crystal 0.9.0

asterite:
Fixed needed for shards to compile in 0.9.0.

Once this is merged and there's a new release we can use that new release in the homebrew formula.

If you want, you can give me and @waj write access to this repository, maybe with instructions on how to release, so we can do these kind of changes on every release and free you from this task :-)

Thanks!

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -180,12 +180,12 @@ module Shards
     private def run(command, path = local_path, capture = false)
       # Shards.logger.debug { "cd #{path}" }
 
-      Dir.chdir(path) do
+      Dir.cd(path) do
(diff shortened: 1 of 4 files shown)


asterite:
Well, we used our fork for the homebrew formula. This will become less and less often as we stabilize the language.

ysbaddaden:
It looks like integration tests got broken.

ysbaddaden:
The broken tests are related to https://github.com/manastech/crystal/commit/ad47e983e615a3abe1f129af52c1db643c13f809

Shards exists before the coroutine has a change to print the log messages (oops).

PR #66, opened 2015-10-18
Add note that it is built in with Crystal 0.9.0

adlerhsieh:
Update README.


jhass:
built-in is a bit misleading, how about "comes with the official Crystal packages since Crystal 0.9.0"?

adlerhsieh:
Yeap that makes more sense. Thanks. Updated.

ysbaddaden:
I will wait for Crystal v0.10.0 to add that note. The bundled shards [has a problem](https://github.com/ysbaddaden/shards/pull/63#issuecomment-148857947).

adlerhsieh:
Got it. I think to include both `shards` (Mac) and `crystal shards` (Linux) commands is better. I'll update the commit when 0.10.0 is out.

PR #76, opened 2015-11-03
[README] shards is distributed via brew together with crystal

mrackwitz:
Nice that they both come together meanwhile. I tried to install [the shards formula](https://github.com/manastech/homebrew-crystal/blob/master/shards.rb), but this is currently failing because of non-matching checksums. Perhaps that should be fixed as well or just pulled?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,8 @@ for more details about the `shard.yml` format.
 You can download a statically compiled 64bit executable of Shards for Linux or
 OS X from the [releases](https://github.com/ysbaddaden/shards/releases) page.
 
+Shards is already distributed with the Homebrew formula [`crystal-lang`](https://github.com/Homebrew/homebrew/blob/master/Library/Formula/crystal-lang.rb) on OS X.
+
 Alternatively you may download a source tarball from the same page (or clone the
 repository) then run `make` —or `make release` for an optimized build— and copy
 the newly generated `bin/shards` somewhere into your PATH. A good place is


waj:
@mrackwitz I just fixed the checksum: https://github.com/manastech/homebrew-crystal/commit/c97094bea356963c81a85fb02fda2b15e30f292b
Thanks for noticing!

ysbaddaden:
Maybe something simpler like:

> Shards is already distributed with the homebrew recipe on OS X:

and

> It's accessible through `crystal shards`.

Then eventually squash your commits as a single one? Thanks!

mrackwitz:
Sure, but how do you mean:
> It's accessible through `crystal shards`.

I see how there is a separate formula, which you can install standalone by running `brew install manastech/crystal/shards`. But it is included in the latest `crystal-lang` formula as a binary release as well. So what is even the recommended way of installing it?
Imho it would be best if `crystal-lang` could depend on the `shards` formula, but I don't even know whether brew supports dependencies across different taps and the name is already taken by a different project in the main tap / repo.

ysbaddaden:
I'm not using OS X so I'm not entirely sure how the brew recipes works. Is it installing Shards standalone, or embedded? In the later case it might be necessary to call `crystal shards` instead of just `shards` (again: I'm not sure).

jhass:
I thought it's `crystal deps`?

mrackwitz:
Coming back to this: It is `crystal deps` (instead of `crystal shards`), but as that only calls through to `shards` as standalone binary, I left it completely out as that's documented already in the rest of the README.

Issue #79, opened 2015-11-10
Github login with personal access token


ScotterC:
Can you point me to the code related to authenticating with github? I have 2 factor authentication and as far as I can tell entering personal access tokens for the password field aren't working

```
 ~/code/crystal-world : shards install
Updating https://github.com/kemalcr/kemal.cr.git
Username for 'https://github.com': ScotterC
Password for 'https://ScotterC@github.com': {entered personal access toke}
Failed to clone https://github.com/kemalcr/kemal.cr.git
```

Shards 0.5.3 (2015-10-30)
Crystal 0.9.1

ScotterC:
Exploring a bit more:

```
 ~/code/crystal-world : shards install -v
Updating https://github.com/kemalcr/kemal.cr.git
git clone --mirror --quiet -- 'https://github.com/kemalcr/kemal.cr.git' kemal
Username for 'https://github.com': ScotterC
Password for 'https://ScotterC@github.com':
Failed to clone https://github.com/kemalcr/kemal.cr.git
```

```
git clone --mirror --quiet -- 'https://github.com/kemalcr/kemal.cr.git' kemal
```
fails the same way

replacing with ssh url succeeds:
```
git clone --mirror --quiet -- 'git@github.com:kemalcr/kemal.git' kemal
```


What's the advantage of https urls over ssh?

ysbaddaden:
Please use `git: git@github.com:user/repo.git` instead of `github: user/repo` when cloning private libraries. The github resolver just rewrites the URL then the git resolver does all the job. It's only meant to cleanup dependencies in `shard.yml` for public libraries.

PR #80, opened 2015-11-10
SSH Github URL instead of HTTPS

ScotterC:
I don't think this is the 'right' solution but making the PR now to
discuss it. Maybe there's a right time to use ssh urls and a right time
to make https urls?

Related to #79

### DIFF
--- a/src/resolvers/github.cr
+++ b/src/resolvers/github.cr
@@ -7,7 +7,7 @@ module Shards
     end
 
     def git_url
-      "https://github.com/#{dependency["github"]}.git"
+      "git@github.com:/#{dependency["github"]}.git"
     end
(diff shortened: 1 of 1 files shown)


ysbaddaden:
I didn't know we could clone any public project throught SSH. Yet, what's the actual benefit? Does it work if we don't have any SSH key on GitHub (eg: on a server)?

ysbaddaden:
CI failed because of `:/` which should be `:` in the URI.

```
Failed to clone git@github.com:/ysbaddaden/minitest.cr.git
```

ScotterC:
Yup. This has to do more with my configuration than the protocol shards is using.

PR #85, opened 2016-01-16
Registry related commands

ysbaddaden:
- search &lt;query&gt; see #87 
- info &lt;package&gt; see #86

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -7,7 +7,7 @@ module Shards
     puts
     puts "Commands:"
     puts "    check"
-    #puts "    info <package>"
+    puts "    info <package>"
     puts "    init"
     puts "    install"
     puts "    list"
@@ -36,9 +36,9 @@ module Shards
         case args[0]? || DEFAULT_COMMAND
         when "check"
           Commands::Check.run(path)
-        #when "info"
-        #  display_help_and_exit(opts) unless args[1]?
-        #  Commands::Info.run(args[1])
+        when "info"
+          display_help_and_exit(opts) unless args[1]?
+          Commands::Info.run(path, args[1])
         when "init"
           Commands::Init.run(path)
         when "install"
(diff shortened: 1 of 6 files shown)


asterite:
:+1:

jayvdb:
ping @ysbaddaden , these are really important commands to have.

RX14:
@jayvdb Shards doesn't have a registry and it won't be getting one. You can search for shards using https://crystalshards.xyz/

straight-shoota:
@RX14 Closing this PR is not a good approach. The issues requesting this (#86 and #87) are still open, so you just declined this implementation. IMHO these are valid feature requests, even without a registry. And they even come from the main contributer.

Shards doesn't have a centralized registry. But it has resolvers with access to shard repositories (github, gitlab, bitbucket). It would be easy to be able to query these resolvers about available shards. Providing this with the CLI would be a nice alternative to going through an external website (context switch and all). This could well be implemented with an additional tool, but I don't see a reason why shards itself shouldn't be able to query resolvers for available shards.

The command interface could look like this:
```shell
# search in github resolver
shards github search foo
# search in all available resolvers
shards search foo
```

RX14:
@straight-shoota either way this is a PR from two and a half years ago, which interfaces with an API that doesn't exist. I don't think its getting merged any time soon.

I didn't close the issues, I just closed this PR.

straight-shoota:
You're right. But your comment sounded like an explanation to close the issues.

jayvdb:
Thanks.  I am happy to get clarification on how this will proceed if this PR isnt it.

For context, @coala is building a shard manager (wrapper) at https://gitlab.com/coala/package_manager/merge_requests/150 , and we need to know whether a shard is locally installed (`shard list`) and we'll soon want to be knowing if a latter package has been released, without updating it, which I assumed `shard info` would have given us.

PR #96, opened 2016-03-20
Use existing version of minitest in usage example

splattael:
As a newbie to Crystal I tried to use the usage example from shards' README to get started.

It seems that `minitest`'s latest version is `0.3.1` and not `1.0.0` as stated in the README.

This commit fixes this issue.


ysbaddaden:
Good idea :-)

PR #101, opened 2016-04-05
Fix spelling mistake in SPEC.md

athix:
Changed "It should be intended with 2 spaces." to "It should be indented with 2 spaces."


ysbaddaden:
Thanks!

Issue #99, opened 2016-04-04
Broken symlink when adding a path dependency


soveran:
I'd like to see if anyone can confirm this issue. The symlink is created from the root of the project, but it should be created from `./libs` instead (so that the path in the link is relative to the directory where the link will be created). The fix should be trivial (changing the working directory to `./libs` before creating the link).

ysbaddaden:
Maybe we should always use an absolute path?

soveran:
I think using relative paths in `shard.yml` has more potential for sharing it with project collaborators, whereas using absolute paths leaves that option out. I have a fix that works, but it requires the paths to be declared relative to `INSTALL_PATH` (instead of the project root) and it also requires the `INSTALL_PATH` directory to be created in order to read the dependency version (in the fix, I create that directory while running `read_spec`, which is a harmless but unexpected side effect).

ysbaddaden:
I said the link should be absolute, not that the path configuration has to be absolute. It could be relative to the project root, expanded from there and eventually created as an absolute link.

This way we'd avoid both caveats: unexpected relative path, and always creating the install path even when not installing.

soveran:
Gotcha, yes, it makes perfect sense :-)

```diff
-      src_path = File.join(local_path, "src")
+      src_path = File.expand_path(File.join(local_path, "src"))
```

ysbaddaden:
Looks good! Can you write an integration test and open a pull request? Otherwise I'll have a look soon.

soveran:
I'll give it a try tomorrow morning.

PR #102, opened 2016-04-08
Fix: expand path in symlink for path dependency

soveran:
Solves issue #99.


ysbaddaden:
Merci !

nuxlli:
Nice!

@ysbaddaden which expected to enter a version? `0.6.3`?

ysbaddaden:
Yes, it will be released with the next Crystal release.

PR #103, opened 2016-05-05
Fix for Crystal 0.16.0: use class vars instead of constants

asterite:
This is because constants are now initialized before "main" code, so changing ENV before requiring code doesn't affect a constant's value.

I ran the tests and they all pass :-)

Only merge if this is the solution you also had in mind.

### DIFF
--- a/src/commands/prune.cr
+++ b/src/commands/prune.cr
@@ -8,13 +8,13 @@ module Shards
         return unless lockfile?
         locks = Lock.from_file(lockfile_path)
 
-        Dir[File.join(INSTALL_PATH, "*")].each do |path|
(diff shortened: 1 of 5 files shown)


asterite:
Oh, good, setters for cache_path and install_dir is a much better solution: https://github.com/crystal-lang/shards/commit/ca0e505e83840aa9af5375d8071d4d1d7e93b652

Closing this :-)

ysbaddaden:
Duh. How stupid can I be. I didn't look at PR before fixing this!

asterite:
@ysbaddaden No worries at all, really. I tried to do it yesterday but failed (I was running the tests in an incorrect way), and today I tried again and was happy to have it work so I decided to send a PR, but your commit is better.

PR #111, opened 2016-07-01
upgrade shard template for manastech/webmock

songjiayang:
For now, manastech/webmock has renamed to manastech/webmock.cr, we should upgrade the template. Details to see: https://github.com/manastech/webmock.cr


ysbaddaden:
Thanks, but there is no reason to target the master branch.

songjiayang:
@ysbaddaden have reseted my commit, Thanks

ysbaddaden:
Thanks!

PR #115, opened 2016-08-08
Add libraries field

### LINKED ISSUE #113: Informational library dependencies field
I propose to add a, for now purely informational, field that lists the shared libraries/sonames a shard wishes to link to, in other words a canonical and standard way to list non-Crystal dependencies.

Each entry provides a set of alternatives. For easy entry each entry is a comma separated list opposed to a nested YAML array, whitespace is ignored.

```yml
name: xml_for_sqlite3
shared_libraries:
  - libsqlite3.so.3.8, libsqlite3.dylib.3.8, libsqlite3.so, libsqlite3.dylib
  - libxml2.so, libxml2.dylib
```

If we can establish this as a community standard, then we have a canonical way of looking up a shards dependencies we need to install. Later tools could be built upon it, such as displaying them on a website, checking whether a system has all available upfront or even tools that look up which packages provide them and install those packages.

jhass:
as discussed in #113

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -25,6 +25,9 @@ development_dependencies:
     git: https://github.com/ysbaddaden/minitest.cr.git
     version: "~> 0.1.0"
 
+libraries:
+  libgit2: ~> 0.24
+
 scripts:
   postinstall: make ext
 
@@ -119,7 +122,7 @@ dependencies:
 A version requirement (String).
 
 - It may be a version number.
-- It may be `*` if any version will do.
+- It may be `"*"` if any version will do.
 - The version number may be prefixed by an operator: `<`, `<=`, `>`, `>=` or `~>`.
 
 Examples: `1.2.3`, `>= 1.0.0` or `~> 2.0`.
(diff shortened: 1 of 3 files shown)


src/spec.cr
```diff
@@ -23,6 +23,25 @@ module Shards
       end
     end
 
+    class Library
+      property soname : String
```
ysbaddaden: Maybe just `name` instead of `soname`?

ysbaddaden:
Just one question, otherwise :ok_hand:

SPEC.md
```diff
@@ -192,6 +195,34 @@ development_dependencies:
     version: ~> 0.1.3
 ```
 
+### libraries
+
+A list of shared libraries the shard tries to link to (Hash).
+
+This field is currently purely informational, shards itself does not
+use it, however other tools might. It serves as a canonical way to discover
+a shards non Crystal dependencies, both for tools as well as humans.
```
ysbaddaden: maybe:
> The field is purely informational. It serves as a canonical way to discover non Crystal dependencies in shards, both for tools as well as humans.

SPEC.md
```diff
@@ -192,6 +195,34 @@ development_dependencies:
...
+
+A shard should only list libraries it directly links to, it should not include
+libraries that are only referenced by dependencies.
```
ysbaddaden: I would be more restrictive here:
>A shard **must** only list libraries it directly links to

src/spec.cr
jhass: I'm on the fence, `soname` future proofs a bit in case we ever want to allow `pkg-config` names too or the like. But if you have a strong preference for `name` I'm ok with changing it.

ysbaddaden: Let's keep `soname`, because keeping `name` for `pkg-config` is a _very_ good idea.

luislavena:
@ysbaddaden @jhass is there any other open point that is blocking this from merge?

I would like to start using the `libraries` field in our shard files to describe the dependencies (and potentially port our work from Ruby).

Thank you!

:heart: :heart: :heart:

ysbaddaden:
Thanks for bumping this up @luislavena !

PR #116, opened 2016-08-08
update notes on how the postinstall works for shards to help clear confusion

jwoertink:
I was pretty confused as to how this works, so I'm hoping to help clear up confusion with this.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -197,10 +197,16 @@ development_dependencies:
 Shards may run scripts automatically after certain actions. Scripts are mere
 shell commands, and can either download and/or compile a library for example.
 
-Only the `postinstall` hook is supported, and is run after installing or
-upgrading a dependency. More hooks may be added later on.
+#### postinstall
+
+The `postinstall` option is for running a command after your project is installed as a dependency of another project. This command does not run at the top level of your project. You should also note that the directory the script is ran from is `libs/some_shard/` directory of the project installing your shard. You should use this to compile additional extensions needed such as C libs, etc...
+
+As an example, if `guardian` was a dependency of a blog project, then running shards install in that blog project would execute the postinstall script right after the `guardian` dependency was installed. It would not run for the blog project itself.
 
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -197,10 +197,16 @@ development_dependencies:
 Shards may run scripts automatically after certain actions. Scripts are mere
 shell commands, and can either download and/or compile a library for example.
 
-Only the `postinstall` hook is supported, and is run after installing or
-upgrading a dependency. More hooks may be added later on.
+#### postinstall
+
+The `postinstall` option is for running a command after your project is installed as a dependency of another project. This command does not run at the top level of your project. You should also note that the directory the script is ran from is `libs/some_shard/` directory of the project installing your shard. You should use this to compile additional extensions needed such as C libs, etc...
```
jhass: > This command does not run when running `shards install` inside the project specifying the hook. 
perhaps? Also
> Note that the directory the script is ran from is `libs/some_shard` of the project installing your shard as a dependency.
And 
> You can use  this to compile additional extensions needed, such as C libs.

SPEC.md
```diff
@@ -197,10 +197,16 @@ development_dependencies:
...
+
+As an example, if `guardian` was a dependency of a blog project, then running shards install in that blog project would execute the postinstall script right after the `guardian` dependency was installed. It would not run for the blog project itself.
```
jhass: > project called blog, then running `shards install` in that project [..]. It would not run for that project itself.

jwoertink:
Cool. Thanks for the suggestions @jhass. I made those changes.

ysbaddaden:
Thanks for proposing this, the definition was indeed confusing. I fixed it myself using a positive definition rather than outlining what it's not, along with a shorter text, but still insisting on what actually happens. I hope any confusion will be gone, now.

Aside: the examples in `SPEC.md` are excerpts only, and not full fledged `shard.yml` examples.

Issue #10, opened 2015-06-23
Require semver as official versioning policy


ozra:
Officially stating that crystal modules/packages should follow semver versioning (http://semver.org) would greatly increase stability in dependency managing.

Version matching becomes much more deterministic and it's simple to lock in on a version specifically, allow just patch updates, feature updates, or even breaking updates at will.

The project structure of crystal solutions seems to be rather well set and thought through already, so this is the logical thing to do to avoid dependency hells.

Following the semver rules are both simple and rewarding, and there's no benefit to ad hoc versioning schemes.

ysbaddaden:
I like the way rubygems states this (buried in documentation):
> The RubyGems library itself does not enforce a strict versioning policy, but using an “irrational” policy will only be a disservice to those in the community who use your gems.
> http://guides.rubygems.org/patterns/#semantic-versioning

Shards does the same: enforce nothing, expect developers to gear toward a sane versioning.

ozra:
Ah, I see, too bad. I think this is one of those things most of us developers _don't_ spend much thought on and where a good outspoken common guideline can really benefit a package eco system, while leaving your mind free to what "really matters" (the app/lib code). Much more than even a common syntax in a jointly used language.

Well, as long as one can write something along the lines of `some_pkg @ 1.2.*` semver matching can be exploited fully (since specifying version ranges via lt(e)/gt(e) has next to no need when using semver, except for in edge cases where a developer has mistakenly released a version of a dep breaking the rules _and_ it actually causes problems [breaking changes without major increase, or new overloads gaining precedence with only patch increase, thus breaking] and needs to be worked around explicitly)

ysbaddaden:
That being said, semver is fully supported, and is the de facto form for release numbers. All the operators have been implemented: =, >, <, >=, <= or the ~> for instance.

I'm just not enforcing the x.y.z scheme, and do allow x.y just like a.b.c.d.e to be release numbers. There are no reasons to not support them, and they work with the above operators.

ozra:
It does make sense, that it is de facto, because, as you mention, it's not an impossibility that a specific project actually may benefit from, say, an additional semantic unit, using four parts.

PR #117, opened 2016-09-08
GitLab shorthand in spec

TheLonelyGhost:
Adds `gitlab: user/project` similar to the one for both github and bitbucket.

This also modifies spelling mistakes and overall formatting of the SPEC.md file, but that can be pulled out into a separate PR if need be.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -145,7 +145,7 @@ Example: `git: git://git.example.org/crystal-library.git`
 
 #### github
 
-A GitHub repository (String).
+A [GitHub](https://github.com) repository (String).
 
 The value is the `user/repository` sheme. Extends the `git` resolver, and acts
 exactly like it. One of the other attributes (`version`, `tag`, `branch` o
@@ -155,13 +155,23 @@ Example: `github: ysbaddaden/minitest.cr`
 
 #### bitbucket
 
-A Bitbucket repository (String).
+A [Bitbucket](https://bitbucket.org) repository (String).
 
 The value is the `user/repository` sheme. Extends the `git` resolver, and acts
 exactly like it. One of the other attributes (`version`, `tag`, `branch` o
 r `commit`) is required. When missing Shards will install the HEAD refs.
 
 Example: `bitbucket: tom/library`
+
+#### gitlab
+
+A [GitLab](https://gitlab.com) repository (String).
+
+The value is the `user/repository` scheme. Extends the `git` resolver, and acts
+exactly like it. One of the other attributes (`version`, `tag`, `branch` or
+`commit`) is required. When missing, Shards will install the HEAD refs.
+
+Example: `gitlab: thelonlyghost/minitest.cr`
 
 #### commit
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -65,14 +65,11 @@ Examples: `minitest`, `mysql2`, `battery-horse`.
 
 The version number of the library (String, required).
 
-- It should follow the [Semantic Versioning 2.0](http://semver.org/spec/v2.0.0.html) format.
+- It should follow the [Semantic Versioning](http://semver.org/) format.
 - It must contain digits.
 - It may contain dots and dashes but not consecutive ones.
 
 Examples: `0.0.1`, `1.2.3` or `2.0.0-rc1`.
-
-While Shards doesn't enforce it, following a rational versioning scheme like
-[Semantic Versioning](http://semver.org/spec/v2.0.0.html) is highly recommended.
 
 ### authors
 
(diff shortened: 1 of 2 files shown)


SPEC.md
```diff
@@ -128,15 +131,29 @@ Examples: `1.2.3`, `>= 1.0.0` or `~> 2.0`.
...
-The library will be linked from the local path. The `version` attribute
-isn't required but will be used if present to validate the dependency.
+The library will be linked from the local path. The `version` attribute is not
+required but will be used if present to validate the dependency.
 
 #### git
 
 A Git repository URL (String).
 
+The URL may be of [any protocol supported by the git command line client](https://git-scm.com/docs/git-clone#_git_urls_a_id_urls_a),
```
ysbaddaden: Maybe a shorter link:

> The URL may be of [any protocol](https://git-scm.com/docs/git-clone#_git_urls_a_id_urls_a) supported by the git command line client,

SPEC.md
```diff
@@ -128,15 +131,29 @@ Examples: `1.2.3`, `>= 1.0.0` or `~> 2.0`.
...
+which may include:
+
+- `ssh://`
+- `git://`
+- `https://`
+- `http://`
+- `ftps://`
+- `ftp://`
+- implied ssh protocol (e.g., `user@example.com:subdir/repo.git`)
```
ysbaddaden: Maybe a simpler definition (let's keep out FTP and HTTP):

> wich includes `git`, `ssh` and `https`.

SPEC.md
```diff
@@ -128,15 +131,29 @@ Examples: `1.2.3`, `>= 1.0.0` or `~> 2.0`.
...
+
+Just the same as the local git command line client, `ssh_config` and associated
+`netrc` settings will be honored for authentication-protected repositories.
```
ysbaddaden: We do rely on the `Git` CLI but we may eventually choose to use `libgit2` or another solution which may not support these. I'm not sure we should include these in the SPEC.

SPEC.md
```diff
@@ -65,21 +65,21 @@ Examples: `minitest`, `mysql2`, `battery-horse`.
 
 The version number of the library (String, required).
 
-- It should follow the [Semantic Versioning](http://semver.org/) format.
+- It should follow the [Semantic Versioning 2.0](http://semver.org/spec/v2.0.0.html) format.
```
ysbaddaden: Please do not target a specific version. Anyone is free to choose which version they want, and a new version will eventually come out, which will require updating the SPEC.

SPEC.md
```diff
@@ -65,21 +65,21 @@ Examples: `minitest`, `mysql2`, `battery-horse`.
...
 - It must contain digits.
 - It may contain dots and dashes but not consecutive ones.
 
 Examples: `0.0.1`, `1.2.3` or `2.0.0-rc1`.
 
 While Shards doesn't enforce it, following a rational versioning scheme like
-[Semantic Versioning](http://semver.org/) is highly recommended.
+[Semantic Versioning](http://semver.org/spec/v2.0.0.html) is highly recommended.
```
ysbaddaden: I forgot we had two references to semver. We should keep only one reference.

ysbaddaden:
The GitLab sugar is fine, and thank you for correcting my English :-)

I added a few propositions about some SPEC corrections. Once answered this is good to merge.

SPEC.md
TheLonelyGhost: Without this I'm concerned how others might guess at specifying different credentials depending on the source.

Would there be a more appropriate place to include this idiosyncrasy?

For all protocols (except ssh) we seem to assume it will be wide open to anonymous users. In reality, as is the case with `Bundler`, we integrate the native configuration files for each protocol. I would imagine forgoing these configuration files in `libgit2` would break far more functionality than it would help. Thoughts?

SPEC.md
TheLonelyGhost: Yep, went a little overboard here when the link to the git-clone definition and a _short_ summary would have sufficed.  😅

SPEC.md
TheLonelyGhost: Really? I would have argued the opposite due to the nature of the semver spec following semantic versioning.

Or is the plan for this project to always support only the latest semver spec?

ysbaddaden: We don't enforce semver (see #10 about that). This is merely a recommendation. The user may choose whatever scheme they want; e.g. [CalVer](http://sedimental.org/calver.html) is a great alternative.

SPEC.md
ysbaddaden: This is exactly why this shouldn't be in the SPEC. We'd then have to always support these solutions, add support for platform specific solutions, ...

SPEC.md
TheLonelyGhost: Ah, wasn't aware of that development decision. I suspected this might have merited its own issue to discuss it rather than be bolted onto this PR, but I suppose that already happened.

Given the discussion in #10 and that this isn't pertinent to the GitLab sugar, I'll propose the changes in a separate PR.

SPEC.md
TheLonelyGhost: Fair enough. Err on the side of fewer features and things like this can be documented in blog posts or other unofficial sources. That way there isn't additional responsibility to support these features.

Will change.

luislavena:
@ysbaddaden is there a chance we can get this into upcoming release? :smiley: 

Thank you!

ysbaddaden:
Merged as 79a0db6a3bc0145c71c12dc33f89feed42aaa556 and e94d256116993b69e4bf44f6df4b31be09977d15. Thank you!

PR #119, opened 2016-09-14
Lift semver constraint; add calver recommendation

ysbaddaden:
Please see [my comment] below(https://github.com/crystal-lang/shards/pull/119#issuecomment-261271199) for an up to date description.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -68,14 +68,14 @@ Examples: `minitest`, `mysql2`, `battery-horse`.
 
 The version number of the library (String, required).
 
-- It should follow the [Semantic Versioning](http://semver.org/) format.
 - It must contain digits.
 - It may contain dots and dashes but not consecutive ones.
 
-Examples: `0.0.1`, `1.2.3` or `2.0.0-rc1`.
+Examples: `1.2.3`, `2.0.0-rc1` or `2016.09`.
 
 While Shards doesn't enforce it, following a rational versioning scheme like
-[Semantic Versioning](http://semver.org/) is highly recommended.
+[Semantic Versioning](http://semver.org/) or [Calendar Versioning](http://calver.org/)
+is highly recommended.
 
 ### authors
 


bcardiff:
If we stick to semver we could eventually try to validate if the change between one version and the next is a breaking change. I haven't try how this work in Elm or if if could be doable in crystal. It feels slightly better to know, at least from the author perspective, if a version change should have breaking changes in the api or not.

ysbaddaden:
I don't want to enforce a versioning scheme in particular. Even if we did, developers are free to not follow it (e.g. breaking change in a patch release that fixes a security issue), so it makes no sense to enforce it (we can't).

There are very different but still valid versioning schemes. CalVer is a nice and appealing alternative for example.

bcardiff:
@ysbaddaden Elm does some checks before bumping version numbers. https://github.com/elm-lang/elm-package#version-rules I like that, but I won't investigate it in the near future. If we lift semver constraint we won't be able to support this kind of contract. I just want to expose this benefits to avoid going back and forth.

ysbaddaden:
I don't know why SemVer actually ended up as a SHOULD. I guess I overlooked it, because there is the RECOMMENDATION to follow a good versioning scheme like SemVer, which I believe to be enough.

Here I propose to fix that, with explicit references to good versioning schemes. I personally prefer to read `16.10.1` that says first patch of the October 2016 release than `1.2.1` for example. I would be very saddened if suddenly Shards refused to let me upgrade the version to `17.4.0` because it's not an invalid increment, or it didn't break things so MAJOR can't be upgraded, or I can't upgrade directly to `18.10.x` for a 2 year release, etc.

spalladino:
Maybe I'm not fully understanding CalVer, but how would the `~>` version operator work with it? I understand how it works with SemVer, since it allows for any increment on the last version component (such as just allowing for minor updates), but how would it be with CalVer?

ysbaddaden:
Let's take Ubuntu's versioning scheme: 12.04.3 for example means april 2014 patch 3. The `~>` operator would only be safe (and meaningful) for the last number.

ysbaddaden:
I'd really like this to be merged. I'll try to explain my reasons again.

SemVer shook things for the greater good. It helped move the opensource software from weird numbers to meaningful numbers. The problem is, in my opinion, that it's too rigid. I assume this was in reaction to chaos, but it doesn't allow any flexibility. Also, arbitrary numbers still lack much meaning (unlike CalVer proposes).

Let's take an example: I got a security issue in v1.4.0, but fixing it requires a limited breaking change. This is only a **PATCH**, but semver enforces to release a new **MAJOR** at v2.0.0. That makes no sense; there is no rationale here: it's a patch, not a major new milestone achieved. It may even have a bad impact on upgrades, since upgrading to a MAJOR is usually hard. This can be dangerous for a security issue.

To sum up:
- a rationale versioning scheme is mandatory
- SemVer is too rigid and makes no sense, sometimes
- there are alternatives that are just as good (e.g., CalVer).

That being said, having a tool analyzing the public API (types, methods and signatures), capable to compare different versions, and knowledgeable about some versioning schemes, that would be great! I just don't think that *enforcing* anything is such a great idea.

spalladino:
I have to be honest here: I watched [Hickey's keynote at Clojure Conj](https://www.youtube.com/watch?v=oyLBGkS5ICk) during the weekend, and he does make a strong point against SemVer. I'd suggest watching it if you have the time. Among other things, he points out that having a versioning scheme that allows you to happily break anything by just changing a major number, and not providing any further info, is not really nice. He actually recommends CalVer as a better alternative, since it allows you to at least know whether to deps are close in time. 

That being said, he also advocates for a way to _describe_ what is being **grown** and **broken** (as opposed to just "changed") by using Spec; though since we have types information in Crystal could allow us to build that almost for free, as @ysbaddaden points out:
> That being said, having a tool analyzing the public API (types, methods and signatures), capable to compare different versions, and knowledgeable about some versioning schemes, that would be great! 

:+1: for this PR from me here

Damn, Hickey sure is a convincing guy...

Willamin:
I think it'd be nice to be able to specify either. For example:
```
dependencies:
    some_library:
        github: username/some_library
        semver: ~> 2.3.0
    some_other_library:
        github: username/some_other_library
        calver: ~> 18.5.0
```

t-richards:
Users of CalVer should be aware of what they’re getting themselves into. Having read through the recommendations, I can see the value of such a versioning scheme (although my initial reaction to CalVer was :nauseated_face: ). I believe the two questions at the end clarify exactly when it is appropriate to use such a scheme.

1. Is your project large or all-encompassing?

  Example given: Ubuntu.

  Ubuntu’s support model revolves around “release date plus N years”, so it makes sense for the versions of this software collection to be a date.

  However, the date on which a particular Ubuntu version was released does not determine whether the software contained in that release will be backward compatible with previously released versions. There can be, and has been, breaking changes of software between Ubuntu versions (e.g. desktop environment changes, language runtime upgrades), and the calendar date of the Ubuntu release does not communicate this fact.

2. Is your project time-sensitive in any way? Do other external changes drive new project releases?

  Example given: pytz.

  The date of a particular pytz release is significant and meaningful because of the external forces (i.e. legislation) which drive the need for a new version of the software.

  Again, however, the calendar date does not communicate to me whether the API of the package has changed significantly between two legislated timezone changes.

---

These two CalVer questions could certainly apply to shard maintainers, depending on the nature of their code. However, I'd like to also address SemVer

Back to SemVer:

>Let's take an example: I got a security issue in v1.4.0, but fixing it requires a limited breaking change. This is only a PATCH, but semver enforces to release a new MAJOR at v2.0.0. That makes no sense;

I disagree. It is exactly this difference, this change between the major version numbers which is significant and informative in SemVer. By looking at the two numbers, the project maintainer has effectively communicated a fact about the project/library. The fact is that a breaking change has occurred, and I will have to deal with the newly-introduced incompatibility accordingly. It does not matter that the change was “just a patch”, the important part is that I will have to perform some manual intervention on my part to avoid breaking my application.


If this project were instead versioned by CalVer, (for example, 2017-10 vs 2018-04), the version difference only informs me that (a) a new version exists, and, (b) it was released on this date. I do not receive any information about the API breaks within, unless I go out of my way to inspect the project’s README or CHANGELOG.

straight-shoota:
@t-richards What does this elaboration add to this issue? It's clear that SemVer, CalVer or any other version naming system has it's pros and cons. The main point is, there is none that fits for all and maintainers should choose what's best for their shard. And `shards` should enable that by not enforcing to use a specific scheme. Which it doesn't. This PR is just about removing the strong recommendation for SemVer in particular from the documentation.

@ysbaddaden Is anything holding this back?

ysbaddaden:
Nothing. Let's merge, at last.

Issue #124, opened 2016-09-23
Compiling Crystal code in postinstall


ysbaddaden:
The `postinstall` script was added to allow the compilation of external C libraries when a Shard is installed as a dependency. But, it may be used to build Crystal binaries that the library (or the project that installed the dependency) could leverage. This may happen more often when #95 is implemented.

The issue is for transitive dependencies, that is when a dependency requires other dependencies to build the executables: they must be 1. installed and 2. accessible.

Shards currently installs dependencies in no specific order —they're actually installed in the order they are found— and immediately runs the `postinstall` script. It's thus likely that the dependencies of a dependency won't be available when a Shard tries to build Crystal code. Furthermore, dependencies are installed inside a single `lib` folder, thus even though they are accessible when building the main project, they aren't made accessible when building inside the dependency (i.e. there is no `lib` folder inside the dependency).

Possible solutions to the **availability** of transitive dependencies:
- install all shards and postpone the `postinstall` script (and executables installation) until they are all installed (basic, can fail if a shard requires a dependency binary);
- sort dependencies, install them in a manner that when installing a library, it's dependencies have already been installed (complex, beware of recursive dependencies).

Possible solution to the **accessibility** of transitive dependencies:
- manipulate the `CRYSTAL_PATH` environment variable when running the `postinstall` script, so it includes the main project's `lib` folder.

Sija:
> install all shards and postpone the postinstall script (and executables installation) until they are all installed;

That sounds like sensible interim solution.

Issue #123, opened 2016-09-23
Integrate new search path algorithm


ysbaddaden:
We must follow the search patch changes of https://github.com/crystal-lang/crystal/pull/2788. This is a **breaking change**; projects will have to run `rm -rf libs; shards install` in their projects for the new dependencies to be installed correctly.
- [x] install dependencies in `lib` instead of `libs`;
- [x] install dependencies as a whole (not merely the `src` folder).

ysbaddaden:
This is blocking #95 (and possibly #110) from being implemented, since the `bin` folder of dependencies isn't available.

RX14:
Should shards detect the old `libs` format and delete it before install? Have a marker file? Is that too much work for little gain?

ysbaddaden:
Maybe print a message? Maybe have `shards prune` remove it? Or ask for permission? I'm never comfortable deleting files on someone's computer...

RX14:
Maybe print a message saying it's in the old format and must be removed, then exit.

PR #126, opened 2016-09-23
Install executables

ysbaddaden:
WIP to implement #95 as per https://github.com/crystal-lang/shards/issues/95#issuecomment-222552996

blocked by ~~#123~~ and maybe #124

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -30,6 +30,9 @@ development_dependencies:
 libraries:
   libgit2: ~> 0.24
 
+executables:
+  - shards
+
 scripts:
   postinstall: make ext
 
@@ -234,6 +237,24 @@ development_dependencies:
   minitest:
     github: ysbaddaden/minitest.cr
     version: ~> 0.1.3
+```
+
+### executables
+
+A list of executables to be installed (Array).
+
+The executables can be of any type or language (e.g., shell, binary, ruby), must
+exist in the `bin` folder of the Shard, and have the executable bit set (on
+POSIX platforms). When installed as a dependency for another project the
+executables will be copied to the `bin` folder of that project.
+
+Executables are always installed last, after the `postinstall` script is run, so
+libraries can build the executables when they are installed by Shards.
+
+```yaml
+executables:
+  - micrate
+  - icr
 ```
 
 ### libraries
--- a/src/commands/update.cr
+++ b/src/commands/update.cr
@@ -14,6 +14,7 @@ module Shards
             Shards.logger.info "Installing #{package.name} (#{package.report_version})"
             package.install
           end
+          package.install_executables
         end
 
         if generate_lockfile?
--- a/src/package.cr
+++ b/src/package.cr
@@ -1,3 +1,4 @@
+require "file_utils"
 require "./resolvers/*"
 require "./helpers/versions"
 
@@ -80,15 +81,20 @@ module Shards
           source = File.join(resolver.install_path, "bin", name)
           destination = File.join(Shards.bin_path, name)
 
-          {% if flag?(:windows) %}
-            FileUtils.cp(source, destination)
-          {% else %}
-            if File.exists?(destination)
-              next if File.stat(destination).ino == File.stat(source).ino
-              File.delete(destination)
-            end
+          if File.exists?(destination)
+            next if File.stat(destination).ino == File.stat(source).ino
+            File.delete(destination)
+          end
+
+          begin
             File.link(source, destination)
-          {% end %}
+          rescue ex : Errno
+            if {Errno::EPERM, Errno::EXDEV}.includes?(ex.errno)
+              FileUtils.cp(source, destination)
+            else
+              raise ex
+            end
+          end
         end
       end
     end
--- a/src/package.cr
+++ b/src/package.cr
@@ -73,27 +73,27 @@ module Shards
     end
 
     def install_executables
-      if installed? && spec.executables.any?
-        Dir.mkdir_p(Shards.bin_path)
+      return if !installed? || spec.executables.empty?
 
-        spec.executables.each do |name|
-          Shards.logger.debug "Install bin/#{name}"
-          source = File.join(resolver.install_path, "bin", name)
-          destination = File.join(Shards.bin_path, name)
+      Dir.mkdir_p(Shards.bin_path)
 
-          if File.exists?(destination)
-            next if File.stat(destination).ino == File.stat(source).ino
-            File.delete(destination)
-          end
+      spec.executables.each do |name|
+        Shards.logger.debug "Install bin/#{name}"
+        source = File.join(resolver.install_path, "bin", name)
+        destination = File.join(Shards.bin_path, name)
 
-          begin
-            File.link(source, destination)
-          rescue ex : Errno
-            if {Errno::EPERM, Errno::EXDEV}.includes?(ex.errno)
-              FileUtils.cp(source, destination)
-            else
-              raise ex
-            end
+        if File.exists?(destination)
+          next if File.stat(destination).ino == File.stat(source).ino
+          File.delete(destination)
+        end
+
+        begin
+          File.link(source, destination)
+        rescue ex : Errno
+          if {Errno::EPERM, Errno::EXDEV}.includes?(ex.errno)
+            FileUtils.cp(source, destination)
+          else
+            raise ex
           end
         end
       end
--- a/src/package.cr
+++ b/src/package.cr
@@ -71,6 +71,28 @@ module Shards
       raise ex
     end
 
+    def install_executables
+      if installed? && spec.executables.any?
+        Dir.mkdir_p(Shards.bin_path)
+
+        spec.executables.each do |name|
+          Shards.logger.debug "Install bin/#{name}"
+          source = File.join(resolver.install_path, "bin", name)
+          destination = File.join(Shards.bin_path, name)
+
+          {% if flag?(:windows) %}
+            FileUtils.cp(source, destination)
+          {% else %}
+            if File.exists?(destination)
+              next if File.stat(destination).ino == File.stat(source).ino
+              File.delete(destination)
+            end
+            File.link(source, destination)
+          {% end %}
+        end
+      end
+    end
+
     def to_lock(io : IO)
       key = resolver.class.key
       io << "    " << key << ": " << @dependency[key] << "\n"
--- a/src/resolvers/resolver.cr
+++ b/src/resolvers/resolver.cr
@@ -44,7 +44,7 @@ module Shards
       end
     end
 
-    protected def install_path
+    def install_path
       File.join(Shards.install_path, dependency.name)
     end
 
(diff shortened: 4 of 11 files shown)


src/validator.cr
```diff
@@ -0,0 +1,54 @@
+module Shards
+  class ValidationError < Error
+  end
+
+  module Validate
+    macro error(message)
```
RX14: Does this have to be a macro?

src/validator.cr
```diff
@@ -0,0 +1,54 @@
...
+      end
+    end
+
+    def self.version(version)
+      unless version =~ /[0-9]/
+        error "version must contain digits"
+      end
+
+      unless version =~ /\A[0-9.]+(-[-_a-z.])?\Z/
+        warning "version should follow semantic versioning"
```
RX14: Mention calver?

test/integration/install_test.cr
```diff
@@ -249,4 +249,22 @@ class InstallCommandTest < Minitest::Test
...
+
+  def test_creates_binstubs
+    metadata = {
+      dependencies: {
+        binary: { type: "path", path: rel_path(:binary) },
+      }
+    }
+
+    with_shard(metadata) do
+      puts run("shards install --no-color", capture: true)
```
RX14: Debug puts

ysbaddaden:
@RX14 bad commit, the `src/validator.cr` file shouldn't have been there!

SPEC.md
```diff
@@ -195,6 +198,25 @@ development_dependencies:
     version: ~> 0.1.3
 ```
 
+### executables
+
+A list of executables to be linked by the shared (Array).
+
+The executables can be of any type or language (e.g., shell, binary, ruby), they
+must have the executable bit set (POSIX platforms) and they must exit in the
```
Sija: `exit` -> `exist`

ysbaddaden:
I'm changing my point of view, again. We should have stubs instead of copying the executables, in order to have some environments variables about the project, namely `APP_ROOT`, `APP_NAME`, `SHARD_ROOT` as well as a tweaked `CRYSTAL_PATH`. This will allow to have some information about  the project, and allow to run/build crystal code on-the-fly using the project's `lib` folder for dependencies (for example).

This will require a `binstubs` command to regenerate the stubs, for example after moving a project to another folder on the computer.

ysbaddaden:
I'm now unconvinced by binstubs :confounded:  I really prefer simply copying executables into the `bin` folder. Bonus: it's already implemented and can be merged right now.

For @bcardiff scenario: the executable can assume (or require) to be run from the project's top directory (e.g. `bin/generator`). That would automatically use the `lib` folder if calling `crystal` for example. If the executable is actually a crystal source code with a shebang (`#! /usr/bin/env crystal`) then it may require dependencies or the shard itself as a dependency (`require "name/file"`) or directly (`../lib/name/src/file`). The app context could be merely `Dir.current` or `File.dirname(Dir.executable_path)`; then the shard contents can be found under `lib/name`, etc.

RX14:
@ysbaddaden What are the disadvantages of binstubs? It doesn't seem like all too much more code.

src/resolvers/resolver.cr
```diff
@@ -44,7 +44,7 @@ module Shards
       end
     end
 
-    protected def install_path
+    def install_path
```
dmitryrck: You have to test this method if you are turning it public #justsaying

ysbaddaden: No. Writing unit tests for details would merely duplicate the integration test.

dmitryrck: I don't agree. If you are creating something public you have to guarantee it is correct and the way to do that is automated tests.

But it is up to you.

ysbaddaden: This is an application, not a library. You may use it as a library, but at your own risk: I don't guarantee the internal API to be stable.

RX14: It's not "public" because shards is an application not a library. It doesn't have to guarantee anything except it's command-line behaviour.

dmitryrck: Good to know that you don't guarantee _the internal API to be stable_ for you or for those who one day would contribute.

Like I said: it is up to you.

src/package.cr
```diff
@@ -67,6 +67,20 @@ module Shards
...
+    def install_executables
+      if installed? && spec.executables.any?
+        bin_path = File.join(Dir.current, "bin")
+        Dir.mkdir_p(bin_path)
+
+        spec.executables.each do |name|
+          Shards.logger.debug "Install bin/#{name}"
+          source = File.join(resolver.install_path, "bin", name)
+          dest = File.join(bin_path, name)
+          FileUtils.cp(source, dest)
```
RX14: Why not symlink?

RX14:
Re: binstubs i'm actually not sure what they would enable.

RX14:
I agree with @ysbaddaden that executables should just assume they're run from the project root. Both `crystal` and `shards` assume this. Let's just merge this as-is.

src/package.cr
ysbaddaden: Because `bin` should assume to have binaries? Thus, to prepare a package for example, I can `cp bin/foo dist/bin/` rather than `cp $(realpath bin/foo) dist/bin/` or digging `cp lib/foo/bin/foo dist/bin` because `bin/foo`.

Maybe we could hardlink?

src/commands/update.cr
```diff
@@ -14,6 +14,7 @@ module Shards
             Shards.logger.info "Installing #{package.name} (#{package.version})"
             package.install
           end
+          package.install_executables
```
ysbaddaden: Hum, executables haven't been rebuilt if the dependency was already installed, why bother reinstalling it? Or maybe a loose version (reinstall if missing?)

ysbaddaden:
Reworked. It now creates a hardlink on POSIX systems, unless the same executable was already installed (i.e. same inode). On Windows it just copies the executable.

src/package.cr
```diff
@@ -71,6 +71,28 @@ module Shards
...
+          destination = File.join(Shards.bin_path, name)
+
+          {% if flag?(:windows) %}
+            FileUtils.cp(source, destination)
+          {% else %}
+            if File.exists?(destination)
+              next if File.stat(destination).ino == File.stat(source).ino
+              File.delete(destination)
+            end
+            File.link(source, destination)
```
straight-shoota: Just because not being executed on windows doesn't mean the file system necessarily supports hard links. There should be a fallback in case `File.link` does not work.

RX14: windows supports hard links though

src/package.cr
```diff
@@ -71,6 +71,28 @@ module Shards
...
+        spec.executables.each do |name|
+          Shards.logger.debug "Install bin/#{name}"
+          source = File.join(resolver.install_path, "bin", name)
+          destination = File.join(Shards.bin_path, name)
+
+          {% if flag?(:windows) %}
+            FileUtils.cp(source, destination)
+          {% else %}
+            if File.exists?(destination)
+              next if File.stat(destination).ino == File.stat(source).ino
```
RX14: I'd just take this out, and either:
- always replace
- always ignore if it exists.

ysbaddaden: We can't ignore: updating a shard may build a new executable, if shards made the hardlink, we're all set (no need to copy), but if the destination was changed for some reason, then we should delete/link again.

src/package.cr
ysbaddaden: Sorry, I'm blinded. To me 1. Windows is stuck on FAT32 file systems, and 2. anything else is POSIX and supports hard links (or we're talking embedded systems and they're out-of-scope).

I'll remove the Windows check and assume POSIX until someone with better win32 knowledge ports shards.

straight-shoota: POSIX generally supports hard links, but not every supported filesystem does. That can be an issue for example with remote file systems.

I typically use crystal in a linux VM which mounts the windows host's file system. Creating hard links is not possible on such a file system.

There are probably other issues as well but the main point is: it doesn't need to be a hard link, a symlink or copy work as well (allthough with minor drawbacsk). Hard links are not 100% supported everywhere, so there should be an alternative as fallback.

src/package.cr
RX14: Well inodes are going to be removed from `File::Info` because they're platform-specific so there's little point in having this check just to remove it later.

Let's just always hardlink.

src/package.cr
ysbaddaden: Sigh.

src/package.cr
ysbaddaden: Sigh.

ysbaddaden: Let's just make File::Info have mtime, file? and directory?. Let's drop everything else. Not cross platform, useless or whatever else. I'll use `stat(2)` directly to get anything done 😭

ysbaddaden: Well:
```crystal
File.link("foo", "bar") # => 0
File.link("foo", "bar") # => Error creating link from foo to bar: File exists (Errno)
```

The only solutions are:
- delete/copy all the time (stupid, waste);
- check inodes, delete + link if they're different (smart, efficient).

I'm sorry, but I won't remove the inode check. You'll have to add `File::Info#inode` for systems that support it, so we have a compile time error on windows, for example.

RX14: No, the correct solution is to hide the inode and supply a `File::Info#same?(File::Info)`. On unix it will check if device and inode are equal (but those fiends are still hidden). On windows it does whatever.

src/package.cr
```diff
@@ -80,15 +80,20 @@ module Shards
...
+          if File.exists?(destination)
+            next if File.stat(destination).ino == File.stat(source).ino
+            File.delete(destination)
+          end
+
+          begin
             File.link(source, destination)
-          {% end %}
+          rescue Errno : ex
+            if ex.errno == Errno::EPERM
```
RX14: Does it hurt to try regardless of the error? Stuff like `EXDEV`... (extremely unlikely)

RX14:
Apart from the above incredibly tiny nitpick i'm :+1: to merge.

plainas:
OK, the discussion has branched quite a bit since 'executables' was mentioned for the first time.

Personally, I think a simple way of installing binaries is of primary importance. Crystal programs, being, by and large, dependency free self contained libraries, are a perfect fit for small command line utilities.

Being able to run `crystal init`, write some code, `shards install` and have a command line tool on the system, is of tremendous convenience. How far is this feature from becoming a reality?

Is #124 still blocking this?

RX14:
Would be nice to get this in for next release. Could help avoid things like rails cli being global in crystal.

Sija:
What's the flow of installing executables specified in `targets`?
Would below code work?

```yaml
# shard.yml

# ...

targets:
  foo:
    main: src/cli.cr

scripts:
  postinstall: shards build --no-debug

executables:
  - foo
```

Great to see this merged btw! 🎉

ysbaddaden:
Yes, but transitive dependencies will prevent building your executable. See #124.

ysbaddaden:
BTW: don't use `--no-debug`. You're only shaving a few kilobytes —we only generate file:line info by default— and make it damn harder to debug exceptions. You should use `--release` thought.

PR #127, opened 2016-09-25
New Crystal search path algorithm

### LINKED ISSUE #123: Integrate new search path algorithm

ysbaddaden:
- Shards counterpart to support https://github.com/crystal-lang/crystal/pull/2788
- closes #123 
- unblocks #126 

Breaking changes:
- dependencies are installed into the `lib` folder;
- dependencies are fully installed (not merely the `src`);
- `postinstall` scripts are now run from the root of a dependency, not the `src` folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,8 @@
 /.crystal
 /.shards
-/libs
+/lib
 /bin/shards
 /test/.repositories
 /test/.shards
-/test/.libs
+/test/.lib
 /tmp
--- a/SPEC.md
+++ b/SPEC.md
@@ -235,7 +235,7 @@ installed or upgraded in a project that requires it. This may be used to compile
 a C library, to build tools to help working on the project, or anything else.
 
 The script will be run from the dependency's installation directory, for example
-`libs/foo` for a Shard named `foo`.
+`lib/foo` for a Shard named `foo`.
 
 ```yaml
 scripts:
--- a/src/config.cr
+++ b/src/config.cr
@@ -13,7 +13,7 @@ module Shards
   end
 
   def self.install_path
-    @@install_path ||= ENV.fetch("SHARDS_INSTALL_PATH") { File.join(Dir.current, "libs") }
+    @@install_path ||= ENV.fetch("SHARDS_INSTALL_PATH") { File.join(Dir.current, "lib") }
   end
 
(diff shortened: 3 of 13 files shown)


asterite:
Now that 0.19.4 is out *I think* this PR should work if re-run on travis.

The nice thing is that now Crystal's `require` will work both for the current way (`libs`, without `src` directory) and with the new way (`lib`, with `src` directory). Because of that, I think shards should, regardless of the command run, check if there's a `libs` directory. If so, delete it and run `shards install` again. That will build the necessary structure but this time in `lib`. Then run the command the user asked, which will target `lib`. That way if you don't run `shards` your project will work, and if you do run it, it will continue to work, and the user doesn't have to know about this migration :-)

Side note: in the last release I made sure to include both `lib` and `libs` in the default CRYSTAL_PATH, in the binary wrapper. You should do the same for the Alpine, FreeBSD and OpenBSD releases if you didn't do so.

ysbaddaden:
1. Travis is happy now :-)

2. I'm never comfortable deleting data, so maybe I'll print a warning (found deprecated `libs` folder), with an explanation of what changed, so developers can fix their Shards or report that they broke, and why to their authors.

3. I overlooked that, I fixed the OpenBSD/FreeBSD releases (also added version information, since ABI compatibility isn't guaranteed between releases) and the Alpine releases.

asterite:
1. 🎉 

2. Sounds good, I forgot that later a user will see the `libs` directory gone and maybe panic :-P

3. Nah, I forgot to announce that when I made the release 😊

luislavena:
Hello @ysbaddaden and @asterite

Perhaps will be good to get this one merged now that 0.19 has been out for quite some time? (and we don't miss the 0.20 release) :smile: 

Thank you
:heart: :heart: :heart:

asterite:
Yes, it would be really, really nice to have this merged before 0.20.0, and also a new shards version so I can use it in the next release.

PR #134, opened 2016-11-06
Do not generate lockfile when there are no deps.

barisbalic:
I noticed when running the `shards update` command that it generates an invalid lockfile for projects with no dependencies, this then upsets `crystal deps` of course.  It seems a bit odd so this is a proposal to avoid generating a lockfile if there are no dependencies to lock.
I would be happy to modify it if it's unsuitable for any reason.

### DIFF
--- a/src/commands/update.cr
+++ b/src/commands/update.cr
@@ -5,6 +5,11 @@ module Shards
     class Update < Command
       # TODO: only update specified dependencies (ie. load locked versions, but don't enforce them)
       def run
+        unless has_dependencies?
+          Shards.logger.info "Dependencies are satisifed"
+          return
+        end
+
         manager.resolve
 
         manager.packages.each do |package|
@@ -17,6 +22,10 @@ module Shards
         end
 
         manager.to_lock(lockfile_path)
+      end
+
+      private def has_dependencies?
+        spec.dependencies.any? || (!shards.production? && spec.development_dependencies.any?)
       end
     end
   end
--- a/src/commands/update.cr
+++ b/src/commands/update.cr
@@ -5,11 +5,6 @@ module Shards
     class Update < Command
       # TODO: only update specified dependencies (ie. load locked versions, but don't enforce them)
       def run
-        unless has_dependencies?
-          Shards.logger.info "Dependencies are satisifed"
-          return
-        end
-
         manager.resolve
 
         manager.packages.each do |package|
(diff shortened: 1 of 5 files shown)


src/commands/update.cr
```diff
@@ -18,6 +23,10 @@ module Shards
 
         manager.to_lock(lockfile_path)
       end
+
+      private def has_dependencies?
+        spec.dependencies.any? || !shards.production? || spec.development_dependencies.any?
```
RX14: Should be `spec.dependencies.any? || (!shards.production? && spec.development_dependencies.any?)` I think.

barisbalic: Ahh, I lifted it from: https://github.com/crystal-lang/shards/blob/master/src/commands/check.cr#L20

barisbalic: (Although what you say makes sense and I'll change it now.)

ysbaddaden:
Thanks for fixing this bug, and writing a test!

Maybe we could use the same logic than the install command: add a `generate_lockfile?` method and prevent calling `manager.to_lock`. This would give a better symmetry between the commands.

```crystal
def generate_lockfile?
  !Shards.production? && manager.packages.any?
end
```

BTW: running the `update` command in production mode should be prohibited, because the generated lockfile would drop versions for development dependencies.

test/integration/update_test.cr
```diff
@@ -100,4 +100,11 @@ class UpdateCommandTest < Minitest::Test
       refute_locked "orm"
     end
   end
+
+  def test_empty_dependencies
+    metadata = { dependencies: {} of Symbol => String }
+    with_shard(metadata) do
+      refute_lockfile
```
ysbaddaden: No need to create a refutation. I don't think we'll reuse it.

barisbalic:
Thanks for the feedback, everything you said sounded very sensible so I've applied it in the following commit.  Let me know if you have any more, and if not I will squash the commits.

ysbaddaden:
Reworked and merged as 32d1bdd24ace181186d44f68ed6be88b94759c1e. Thanks!

barisbalic:
No problem, but out of curiosity, after applying your feedback, what made you decide not to merge my PR?

ysbaddaden:
I did merge your PR, but squashed and simplified manually, and GitHub doesn't reflect that.

barisbalic:
Ah, in that case, my apologies.

Issue #110, opened 2016-06-12
Use as a Build System


keplersj:
Should `shards` be used as a build system manager in the future - much like Rust's `cargo`, node.js's `npm`, or Java's `maven`?

ysbaddaden:
I can't find anything about building with NPM... I suppose you want shards to become aware of targets to build, downloading dependencies, building the targets, and maybe installing them?

It may be a little out of scope, at least for the time being. More often than not a simple Makefile is usually enough.

What would be your use cases? And what kind of problems should it be solving?

I don't know much about Cargo's building features nor Maven's, and I only skimmed rebar or the Haskell build tool (Cabal?).

luislavena:
I think it becomes interesting to define a `build` part for `shards` when you can actually define executables that can be build.

In that scenario, having different options for different targets (like cross-compilation, compiler or linker options) become necessary.

However I feel that Crystal still lacks a bit of maturity in relation to cross compilation (ie, it always uses `CC` so cross compilation requires you prepend `CC=xxx` to the build command).

hattwj:
I could see adding the ability to define arbitrary groups as being useful. Something akin to a Gemfile in ruby. Then a makefile etc.. could build the executable the appropriate groups enabled. I only gave the shards yml SPEC a cursory glance, but it appears to only allow dependencies and development_dependencies. I could see the possible need for test, production etc... groups as well.

ysbaddaden:
The more I think about it, the more I like the idea. I'm not saying I will work on this anytime soon, but I believe this is a good development for Crystal.

An example use case (thinking out loud). Crystal's source could follow the Git scheme and build multiple binaries like `crystal-build` `crystal-docs`, and so on for each subcommand. A Shards SPEC could contain something like this:

```yaml
targets:
  crystal-build:
    main: src/crystal/commands/build.cr
  crystal-docs:
    main: src/crystal/commands/docs.cr
```

Then:

```
$ shards build [all]
$ shards build docs [--force]
```

- The `build` command would resolve/install dependencies then build the specified targets (or everything). Actually it could even parse the files for `require` statements, look for missing dependencies, and verify if we actually need to build the binary, or not (with a flag to force a rebuild).
- Each target could have different helpers; for example `cc: musl-gcc`, `release`, `debug`, ...
- Maybe a global `$HOME/.shard.yml` (for example) for global settings.

iDev0urer:
[...]

PR #136, opened 2016-11-15
Added a feature `shards build`

tbrand:
Hi all,

This PR is based on https://github.com/crystal-lang/crystal/pull/3538.
I added a `build` command for `shards`.

In shard.yml, we set
```
targets:
  default:
    main: src/main.cr
    options:
      - -o bin/main
      - --release
  sub:
    main: src/sub.cr
```

Then,
```
$ shards build # Build "default" target without specified target
Building: default

$ shards build sub # Build specified target, here it is "sub"
Building: sub

$ shards build all # Build all targets
Building: default
Building: sub

$ shards build fake # Undefined targets raise an error
Target 'fake' is not found
```

Thanks for your review!

### DIFF
--- a/src/commands/build.cr
+++ b/src/commands/build.cr
@@ -3,34 +3,69 @@ require "./command"
 module Shards
   module Commands
     class Build < Command
+
+      @@args = [] of String
+
+      # command line arguments
+      @targets = [] of String
+      @options = [] of String
+
       def run
-        # Return if 'crystal' command is not installed
-        return unless has_crystal_command?
-
-        @sub ||= "default"
-
-        if @sub == "all"
+        # Install dependencies before the build
+        Install.run(@path)
+        # Parse to find targets and options
+        parse_args
+        # mkdir bin
+        mkdir_bin
+        if @targets.size == 0
+          raise Error.new("Error: No target found in shard.yml") if manager.spec.targets.nil?
           manager.spec.targets.each do |target|
             build target
           end
         else
-          target = manager.spec.targets.find{ |t| t.name == @sub }
-          raise Error.new("Target \'#{@sub}\' is not found") if target.nil?
-          build target
+          @targets.each do |name|
+            target = manager.spec.targets.find{ |t| t.name == name }
+            raise Error.new("Error: target \'#{name}\' is not found") if target.nil?
+            build target
+          end
         end
+      end
+
+      def parse_args
+        is_option? = false
+        @@args.each do |arg|
+          is_option? = true if arg.starts_with?("-")
+          if is_option?
+            @options.push(arg)
+          else
+            @targets.push(arg)
+          end
+        end
+      end
+
+      def mkdir_bin
+        bin_path = File.join(@path, "bin")
+        Dir.mkdir bin_path unless Dir.exists?(bin_path)
       end
 
       def build(target)
         Shards.logger.info "Building: #{target.name}"
 
+        args = ["build", target.main] + @options
+        unless @options.includes?("-o")
+          args.push("-o")
+          args.push(File.join("bin", target.name))
+        end
+
         error = MemoryIO.new
-        status = Process.run("/bin/sh", input: MemoryIO.new(target.cmd), output: nil, error: error)
+        status = Process.run("crystal",
+                             args: args,
+                             output: nil, error: error)
         raise Error.new("#{error.to_s}") unless status.success?
       end
 
-      def has_crystal_command?
-        raise Error.new("\'crystal\' is not installed") unless system("which crystal > /dev/null 2>&1")
-        true
+      def self.set_args(args : Array(String))
+        @@args = args
       end
     end
   end
--- a/src/commands/build.cr
+++ b/src/commands/build.cr
@@ -1,0 +1,36 @@
+require "./command"
+
+module Shards
+  module Commands
+    class Build < Command
+      def run
+        # Return if 'crystal' command is not installed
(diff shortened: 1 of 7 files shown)


src/commands/command.cr
```diff
@@ -5,13 +5,14 @@ require "../spec"
...
     getter path : String
+    getter sub : String | Nil
     getter spec_path : String
     getter lockfile_path : String
 
     @spec : Spec?
     @locks : Array(Dependency)?
 
-    def initialize(path)
+    def initialize(path, @sub = nil)
```
ysbaddaden: What is `sub`? Other commands don't need this, so we should avoid it.

bcardiff:
Related to #110

src/spec.cr
```diff
@@ -94,6 +95,21 @@ module Shards
...
             development_dependencies << dependency
           end
+        when "targets"
+          read_mapping(pull) do
+            target = Target.new(pull.read_scalar)
+            read_mapping(pull) do
+              case pull.read_scalar
+              when "main"
+                target.main = pull.read_scalar
+              when "options"
```
ysbaddaden: Let's avoid `options` for now; they could be passed as options to the command for example:

```
$ shards build --release
```

src/commands/command.cr
ysbaddaden: What is `sub`? Can we avoid it?

src/target.cr
```diff
@@ -0,0 +1,17 @@
...
+    property main : String
+    property options : Array(String)
+
+    def initialize(@name)
+      super()
+      @main = ""
+      @options = [] of String
+    end
+
+    def cmd
```
ysbaddaden: What about `shell_command`? Thought the model shouldn't have to deal with the invocation. Command::Build should be the one generating the command.

src/target.cr
```diff
@@ -0,0 +1,17 @@
+module Shards
+  class Target
+    getter name : String
+    property main : String
+    property options : Array(String)
+
+    def initialize(@name)
+      super()
```
ysbaddaden: Not needed: there are no superclass.

test/integration/build_test.cr
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def test_fails_when_target_is_missing
+    metadata = { targets: { mock: { main: "mock.cr" } } }
+
+    with_shard(metadata) do
+      Dir.cd(application_path) do
+        File.open "mock.cr", "w"
+        ex = assert_raises(FailedCommand) { run "shards build mock_fake" }
+        assert_match "\e[31mTarget\e[0m 'mock_fake' is not found\n", ex.stdout
```
ysbaddaden: Please pass `--no-color` to the command.

tbrand:
Thanks for your review!
I believe I've implemented what you suggested :)
Please tell me if I haven't do it.

We set
```
targets:
  test0:
    main: src/main0.cr
  test1:
    main: src/main1.cr
```
Then
```
$ shards build # Build main0.cr and main1.cr and generate binaries onto bin/test0 and bin/test1
Building: test0
Building: test1

$ shards build test1 # Build main1.cr and generate a binary onto bin/test1
Building: test1

$ shards build test0 test1 # Same as the top case
Building: test0
Building: test1
```

One thing that I have to say is that it is quite difficult to know whether an argument is a "target" or "option"
For example,
```
shards build -aaa bbb
```
In this case, shards cannot decide bbb is a value of "-aaa" or a target.
If it is okay to decide the order to "shards -> build -> targets -> options", it is possible to regard the input after the arguement with "-" prefix as options.
```
shards build aaa bbb -ccc ddd
# => aaa and bbb are targets, ddd is a value of -ccc
```
I can hard code `crystal build` options, but it is not ideal from the view point of frexibility.

For now, I took first approach.(Regarding the input after the argument with "-" prefix as options)
Please give me comments especially for this implementation.

As a supplement, I don't understand why the test/cli.cr parses a hash manually.
Why don't it use yaml.cr?

ysbaddaden:
Thanks! I'll correctly review this when I have some free time.

src/commands/command.cr
```diff
@@ -5,6 +5,7 @@ require "../spec"
 module Shards
   abstract class Command
     getter path : String
+    getter sub : String | Nil
```
bcardiff: unused?

src/commands/command.cr
```diff
@@ -24,8 +25,8 @@ module Shards
 
     abstract def run
 
-    def self.run(path)
-      new(path).run
+    def self.run(path, sub = nil)
```
bcardiff: unused change? It should be rolledback probably.

src/commands/build.cr
```diff
@@ -3,34 +3,69 @@ require "./command"
...
+            raise Error.new("Error: target \'#{name}\' is not found") if target.nil?
+            build target
+          end
         end
       end
 
+      def parse_args
+        is_option? = false
+        @@args.each do |arg|
+          is_option? = true if arg.starts_with?("-")
```
Sija: You can use char here `'-'`

ysbaddaden:
Squashed and merged as 82fef4d544078dbdddb1ba956dd5f5da399ac9f5.

tbrand:
You meant you keep `sub`?
https://github.com/crystal-lang/shards/commit/82fef4d544078dbdddb1ba956dd5f5da399ac9f5#diff-3d1bd35545e168a4d913eb6fb9a9f87bR8

Issue #137, opened 2016-11-19
Uniqueness on which context?


tomas:
So according to the [shards.yml spec](https://github.com/crystal-lang/shards/blob/master/SPEC.md), all package names must be unique. However there's no mention of an official registry to see which names have been registered and which are available.

Or does it simply mean unique in the context of the dependencies installed locally? It that's the case I would suggest a better wording to make it clearer. Like Crystal clear. ;)

bcardiff:
They must be unique in the context of the shards.yml . There is no global registry for crystal packages. The intention is for them to be descentralized. 

This description is in the context of the shards.yml spec. I would think is clear enough but feel free to PR a better wording :-)

tomas:
OK, will do. 

Now, when you say "the intention" is it because this topic hasn't been settled yet? 

I think this should be as clear as possible, considering there already are [two](http://crystalshards.xyz/) [websites](https://crystal-shards-registry.herokuapp.com/) showcasing Crystal shards.

ysbaddaden:
Shards is decentralized. There is no central registry that we talk to, and there is no plan (in the foreseeable future) for Shards to ever resolve dependencies using such a registry. Shards' names are thus unique in the context of a `shard.yml`.

Still, the shard name _must_ be unique, otherwise adding a dependency may eventually generate a conflict that Shards will be incapable to handle, more so to install.

An informational registry will eventually happen to ease the pain of searching for unique names.

tomas:
@ysbaddaden I understand the reasoning and I fully embrace the idea of a non-centralized package system. However the idea of requiring package names to be unique without a proper way of enforcing it seems problematic.

I'm fine with any decision, by the way. Just pointing out that this is a crucial part of Crystal as an ecosystem, and thus shouldn't be treated lightly. I personally think that for names to be unique there needs to be a registry where packages are published (a-la Rubygems, npm, etc). If that's not the case then there's no way of really knowing whether a package *is* unique or not, and thus the requirement seems kind of pointless.

Go seems to have resolved this, by the way, although I don't know much about the Go ecosystem myself. :)

asterite:
Well, here's a good reason to have a centralized repository: https://www.reddit.com/r/programming/comments/7olf5e/npm_registry_many_packages_are_disappearing_all/

If I have a popular crystal repo in github and I decide to remove it, good luck with, I just broke your code. 

[...]

PR #138, opened 2016-11-21
Be more specific about uniqueness of crystal shard names.

tomas:


ysbaddaden:
As stated in https://github.com/crystal-lang/shards/issues/137#issuecomment-262414743 a shard name must still be unique, whatever the context.

vegai:
Should this be closed if it won't be merged?

tomas:
@RX14 @vegai Why not?

RX14:
Because the statement is incorrect. Your package name SHOULD be globally unique.

tomas:
Even though that's pretty much impossible without a central registry, it still wouldn't hurt to state *that* in the SPEC.md file, would it? 

Anyway, thanks for at least considering this. It would have been nice if you had explained the reason why the PR wasn't merged though... (without me having to ask for it).

RX14:
@tomas, I think @ysbaddaden was perfectly clear:

>a shard name must still be unique, whatever the context

tomas:
But you're missing the part where he says that 

> shards' names are thus unique in the context of a shard.yml.

That's why I'm saying that it wouldn't hurt to be extra clear about this. That's all.

ysbaddaden:
Your phrasing would bring ambiguity when there is none: a ahead name is unique, period.

SPEC details rules for the `shard.yml` file, and it doesn't care about implementation choices (no central registry). We could eventually decide to have one, or to have an alternative implementation with a central registry, still compatible with SPEC.

PR #146, opened 2017-01-11
Documented `~>` version specifier

mverzilli:
Reused the explanation from http://bundler.io/v1.5/gemfile.html

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -146,10 +146,10 @@ A version requirement (String).
 
 Examples: `1.2.3`, `>= 1.0.0` or `~> 2.0`.
 
-Most of the version operators, like `>= 1.0.0`, are self-explanatory. 
-The specifier `~>` has a special meaning, best shown by example. 
-`~> 2.0.3` is identical to `>= 2.0.3 and < 2.1`. 
-`~> 2.1` is identical to `>= 2.1 and < 3.0`. 
+Most of the version operators, like `>= 1.0.0`, are self-explanatory, but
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -146,6 +146,12 @@ A version requirement (String).
 
 Examples: `1.2.3`, `>= 1.0.0` or `~> 2.0`.
 
+Most of the version specifiers, like `>= 1.0.0`, are self-explanatory. 
+The specifier `~>` has a special meaning, best shown by example. 
+`~> 2.0.3` is identical to `>= 2.0.3 and < 2.1`. 
+`~> 2.1` is identical to `>= 2.1 and < 3.0`. 
+`~> 2.2.beta` will match prerelease versions like `2.2.beta.12`.
```
ysbaddaden: I'm not sure about this last one. Better remove it unless we can prove that the non-digit parts in version numbers are working correctly.

SPEC.md
```diff
@@ -146,6 +146,12 @@ A version requirement (String).
 
 Examples: `1.2.3`, `>= 1.0.0` or `~> 2.0`.
 
+Most of the version specifiers, like `>= 1.0.0`, are self-explanatory. 
+The specifier `~>` has a special meaning, best shown by example.
```
ysbaddaden: We use the word `operator` above instead of `specifier`.

mverzilli: Modified.

SPEC.md
mverzilli: Agreed. I removed it.

PR #154, opened 2017-02-18
Remove custom Logger class

Sija:
Since crystal-lang/crystal#1781 is fixed, custom `Logger` class can be removed (as per comment in `src/logger.cr`).

ATM trying to `require` logger with shards fails with:

```
in /usr/local/Cellar/crystal-lang/0.20.5_2/src/logger.cr:42: can't reopen enum and add more constants to it
```

### DIFF
--- a/src/logger.cr
+++ b/src/logger.cr
@@ -1,69 +1,5 @@
-#require "logger"
+require "logger"
 require "colorize"
-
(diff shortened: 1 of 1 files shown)


ysbaddaden:
Thanks for the cleanup.

PR #159, opened 2017-06-27
Fix logger formatter for Crystal 0.23

zatherz:
`severity` is now a Logger::Severity, not a string.

### DIFF
--- a/src/logger.cr
+++ b/src/logger.cr
@@ -34,7 +34,7 @@ module Shards
                   message
                 end
         else
-          io << severity[0] << ": " << message
+          io << severity.to_s[0] << ": " << message
         end
       end
     end


kostya:
i think it slow, transforming to string on every message.

zatherz:
Good point, but I don't think it's a huge problem in this case

RX14:
No I think this is pretty much as fast as it gets. Regardless, logging is very very unlikely to be a performance bottleneck for shards.

ysbaddaden:
`Enum#to_s` generates string literals at compile time, but strings exist in the HEAP. It could be better, but that really isn't a bottleneck as pointed by @RX14.

ysbaddaden:
Thank you!

RX14:
@ysbaddaden actually `Enum#to_s` generates the strings as literals for non-flags enums. This means they're not allocated on the heap at all.

Issue #162, opened 2017-07-19
Invalid Int32 when specifying a commit


vladfaust:
This
```
dotenv:
  github: gdotdesign/cr-dotenv
  commit: 341678090110cdb5436f75cc796c785ed392c4be
```
fails with
```
Updating https://github.com/gdotdesign/cr-dotenv.git
Invalid Int32: 341678090110 (ArgumentError)
```
I suppose this is a bug.

Even wrapping in double quotes (`"341678090110cdb5436f75cc796c785ed392c4be"`) raises the same error.

ysbaddaden:
I don't understand why specifying a `commit` hash ends up calling `NaturalSort` on this hash. That shouldn't happen, I guess.

ysbaddaden:
The problem is `GitResolver#available_versions` that injects the specified Git refs as an "available version" (it even defaults to `HEAD`). I guess we should just resolve the `version_at(refs)` but make sure later on to override the resolved version with the specified ref (when installing & locking).

PR #167, opened 2017-08-20
Fixes issue #162

mverzilli:
When comparing commit hashes, we attempt to convert portions of them `to_i`. These portions might end up being bigger than what an `Int32` can hold, thus this bug.

This commit fixes the issue by converting those parts `to_i64` instead.


ysbaddaden:
It assumes intermediary values will fit into an Int64, which may not be the case if there aren't any a-z character in 19 characters (possible).

mverzilli:
Closing in favor of #169

ysbaddaden:
A `2147483648.0.1` version number will still crash... but the odds it happening are low.

PR #168, opened 2017-08-20
Corrects off by 1 error when applying natural sort between version descriptors.

mverzilli:
I'm not sure about this, but I think this was an error. Take for example the commit hash in issue #162:

"341678090110cdb5436f75cc796c785ed392c4be"

Before this change, this natural sort implementation would first consider "341678090110" then skip the subsequent "c" then continue with "db5436f75cc796c785ed392c4be". This is because the `ia` and `ib` indexes get moved `size + 1` instead of just `size`. I made the correction and tests still pass. 

Apologies if I'm missing something about how the algo is supposed to work :).

### DIFF
--- a/src/helpers/natural_sort.cr
+++ b/src/helpers/natural_sort.cr
@@ -16,12 +16,12 @@ module Shards
           return 0 if ia > la
           return 0 unless a[ia .. -1] =~ EXTRACT_ALPHA_OR_NUMBER_GROUP
           aaa = $1
-          ia += $1.size + 1
+          ia += $1.size
 
           return 0 if ib > lb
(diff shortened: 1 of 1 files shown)


ysbaddaden:
It's been 2 years since I last touched this, but while experimenting I believe the `+1` is an arbitrary "skip version dot"... which is plain wrong and not required, since the different regular expressions will skip non alphanumerical chars, which ain't exactly correct either, but okayish for shards:

```crystal
natural_sort("0.01", "0.1")  # => 0 (correct)
natural_sort("0.+?1", "0|1") # => 0 (okayish)
```

PR #186, opened 2017-12-18
Fix README typo, and clarify

lukeholder:


straight-shoota:
While you're at it, could you replace `crystal deps` with `shards`? `crystal deps` just calls `shards` and the shards exectuable should better be used independetly from the compiler.

/cc @ysbaddaden

RX14:
@straight-shoota please submit that change as a separate PR.

straight-shoota:
Sure, but it will need a rebase after this is merged (or vice versa). That's why I suggested to include it here for simplicity.
I wouldn't consider it anything else than a clean-up to keep up with the other mentions of which-command-to-be-used in this file so I think this would also fit in here nicely.

RX14:
This wasn't fixed. The typo is still in the readme.

Issue #144, opened 2016-12-21
shards add/get


asterite:
I'd like to revive this but under a different issue than #81 (which I don't know if it's exactly about this)

Today I wanted to install [this shard](https://github.com/bcardiff/ghshard) in a project (well, actually, I just wanted to try it out). It was straight-forward, I:

* opened `shard.yml`
* added a `dependencies:` entry (because I didn't have dependencies yet)
* added `gshards:`
* added `github: bcardiff/ghshard`
* saved and ran `shards install`

That's OK. However, the more we can automate these trivial tasks I think it's better.

Now, we don't have a centralized repository so we can't do `shards get some_shard` because shards wouldn't know where to find it. But we could have:

```
$ shards get https://github.com/bcardiff/ghshard
```

This will find a Resolver (github, gitlab, bitbucket, path) that understands the argument. For github, gitlab and bitbucket we can use a regex, for "path" we can use `Dir.dir?(path)`.

Let's consider the case of GitHub. Once the command find out it's about a github repository, it opens a connection to it and reads its `shard.yml` to find out its name. Then it adds the name and all necessary info to the existing `shard.yml`.

So with just a single command we can install a shard. And since we usually see a project in GitHub, copying and pasting that URL into the command line, instead of having to add a few formatted lines into `shard.yml`, seems easier and less tedious.

The original reason why I discarded this idea is that editing a YAML file is hard: formatting and comments have to be preserved, etc. However, using a pull parser we can more or less know where the "dependencies" section ends and add right after that, assuming 2 spaces of indentation.

I made a program that does that for GitHub [here](https://gist.github.com/asterite/7fc71c4b11e5ad8cee57a836c99c61f4). I wanted to start adding this to shards, taking into account different resolvers, but I first wanted to check with you, @ysbaddaden , if you agree with this. I'm not super strongly in favor of this, so if you say no then it's not that terrible :-)

kostya:
also #75 about pain with manual edit

ysbaddaden:
Interesting hack. It assumes a 2 space indentation and that the `{}` notation isn't used (which is a strong assumption); maybe it should eventually validate the generated YAML file, to verify that it's still valid before saving it? 

I'm not sure I really want to have to maintain this feature, but since there is popular demand for it, if we have a good solution for updating `shard.yml`, why not?

kostya:
from_yaml -> edit -> to_yaml?

samueleaton:
[...]

PR #192, opened 2018-02-18
Add/remove command

nedpals:
**Keep in mind that this is still a work in progress. I'm still figuring out how will this plays well with the existing features/things found on Shards and integrate them seamlessly.**

The much awaited feature for Shards is here (from #144) and is based on the work I've done on [Sharn](https://github.com/nedpals/sharn). It allows you to add or remove dependencies without manually editing the `shards.yml` file and installs them automatically.

Same as Sharn, the pattern is still the same.
```bash
shards add author/repo#branch@1.0
```
```bash
shards add gitlab:author/repo#branch@1.0
```
## Roadmap
- [ ] Add command
  - [x] Path/repo extraction
  - [ ] Merge new deps into the spec
  - [ ] Write the merged dependencies set into the spec
  - [x] Add dev dependencies with the `--dev` flag
- [ ] Remove command
  - [ ] Extract spec deps and remove selected
  - [ ] Write it back into the spec file.

Also, I haven't tried using unit tests yet and still confuses me how it works.

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -8,12 +8,12 @@ module Shards
     puts "Commands:"
     puts "    build [targets] [options]"
     puts "    check"
-    #puts "    info <package>"
+    # puts "    info <package>"
     puts "    init"
     puts "    install"
     puts "    list"
     puts "    prune"
-    #puts "    search <query>"
+    # puts "    search <query>"
     puts "    update"
     # puts "    update [package package ...]"
     puts "    version [path]"
@@ -30,6 +30,7 @@ module Shards
       opts.on("--no-color", "") { self.colors = false }
       opts.on("--version", "") { puts self.version_string; exit }
       opts.on("--production", "") { self.production = true }
+      opts.on("--dev", "") { self.dev = true }
       opts.on("-v", "--verbose", "") { self.logger.level = Logger::Severity::DEBUG }
       opts.on("-q", "--quiet", "") { self.logger.level = Logger::Severity::WARN }
       opts.on("-h", "--help", "") { self.display_help_and_exit(opts) }
@@ -40,9 +41,9 @@ module Shards
           build(path, args[1..-1])
         when "check"
           Commands::Check.run(path)
-        #when "info"
-        #  display_help_and_exit(opts) unless args[1]?
-        #  Commands::Info.run(args[1])
+          # when "info"
+          #  display_help_and_exit(opts) unless args[1]?
+          #  Commands::Info.run(args[1])
         when "init"
           Commands::Init.run(path)
         when "install"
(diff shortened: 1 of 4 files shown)


src/cli.cr
```diff
@@ -30,6 +30,7 @@ module Shards
       opts.on("--no-color", "") { self.colors = false }
       opts.on("--version", "") { puts self.version_string; exit }
       opts.on("--production", "") { self.production = true }
+      opts.on("--dev", "") { self.dev = true }
```
ysbaddaden: No abbreviations please, this should be `--development`.

I'm uncomfortable having both `--production` (don't install development dependencies) and `--development` (add development dependency to `shard.yml` as global options, but for the time being individual commands didn't need special flags, yet. Maybe an evolution should allow this (later).

plainas:
Looks like this is design discussion as there is a specific trade off that is pretty much a matter of preference.

So it boils down to accepting machine generated changes to shard.yml or settling for manual edits as the sole intended workflow.

Not that it should be blindly followed, but in node.js world, ``npm install --save`` is encouraged as the standard way to add dependencies to your package.json.

Apart from taking ideas from other languages, I would like to point out a specific use case: scriptability.

I am automating the process of creating a new crystal project tailored for internal company usage. In short, I call ``crystal init`` then add a bunch of scripts and other files which I generate from templates.
Right now, the only way to add dependencies to shard.yml is by relying on external tools, being them hacky string replaces or yaml parse-serialize round trips. The command proposed on this issue would simply solve my problem.

ysbaddaden:
@plainas Wrong thread. Keep it focused on *reviewing this pull request* (and delete/paste your comment somewhere else, thanks).

nedpals:
I have stopped the development and will close this PR. Seems that I have lost interest in making this feature for an obvious reason. Don't want to make a discussion but it is not only you would be using this. It would be a key feature for people who are new to Crystal and used to npm-esque type of package managers and wouldn't want to make their `shards.yml` be unreadable by the program itself. Asterite and others have agreed with this approach (for now) because it focuses more on speed and time to type because it is a  CLI. It doesn't matter if its very expensive to maintain. That is why there is GitHub and people can contribute which in return would gradually be stable.

I apologize to anyone who were hyped in this feature. This is my first major PR for a large project and would like to be part of Crystal's ecosystem evolution. I don't want to disappoint anyone that I made this commit and the design decisions I made for it to work on shards I know it is highly discouraged to download a separate helper but in the meantime I would continue my developments in [Sharn](https://github.com/nedpals/sharn) in order for me to make it more stable and as much as possible reduce the problems it such as YAML formatting so that in the future implementation and user experience would be great for all when it arrives on shards.

PR #197, opened 2018-03-27
Makefile: Add install rules

jirutka:


RX14:
Looks good, although I'd like a `make uninstall`.

Makefile
```diff
@@ -58,3 +64,22 @@ test_unit:
...
+.PHONY: install-man
+install-man:
+	mkdir -p $(MANDIR)/man1 $(MANDIR)/man5
+	cp $(filter %.1,$(MANPAGES)) $(MANDIR)/man1/
+	cp $(filter %.5,$(MANPAGES)) $(MANDIR)/man5/
+
+.PHONY: uninstall
+uninstall:
+	rm -f "$(BINDIR)/shards"
+	rm -f "$(MANDIR)/man1/$(notdir $(filter %.1,$(MANPAGES)))"
```
RX14: Shouldn't this be `$(addprefix $(MANDIR)/man1/,$(notdir $(filter %.1,$(MANPAGES))))`?

jirutka: Uh, yes, mea culpa. Fixed now.

ysbaddaden:
Looks good. But I believe the Makefile needs a good rewrite. Something like https://gist.github.com/ysbaddaden/e7fac797ce5a8fabcff722b7dc116ee6 which assumes a POSIX compliant Makefile instead of expecting GNU make, drops the `release` and `releases` targets, and adds the `install` and `uninstall` targets.

ysbaddaden:
Done in a0054b2050e88936ba6571635b9241d6a73562f4 which fully rewrote the Makefile (no more GNU make dependency) and introduced the `install` and `uninstall` targets.

Issue #157, opened 2017-04-30
Run command


pta2002:
I have a suggestion for a new command: `shards run [target] [args]`. This would be the same as executing `shards build [target]` and then running the compiled file with the arguments specified, just shortened to one command. This can be useful for quickly testing out new changes, without needing to run two commands. In fact, crystal already has a `run` command, so maybe use that instead of outputting the file?

GloverDonovan:
`shards run` would be great. Rust has [something similar](https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_run.rs) and it helps a lot when you only have to use one command instead of multiple for quick changes.

ysbaddaden:
There may be many targets configured to build. Which one should a run command execute? Do binaries depend on each other? So, shall we build all targets, then run a single one?

Build is a minimalistic tool, and meant to stay this way; I don't want to add any more configuration. I'm not sure we can achieve anything proper for most scenarios without adding configuration.

RX14:
If there's more than one target, you must specify `shards run target`. No dependencies between targets. That's a simple zero-config way to do this.

ysbaddaden:
So `shards run [target]` where `target` can be omitted if only one is configured (no ambiguity) and prints a helpful error message otherwise. That sounds acceptable. Not really my workflow, but I understand it can be.

If someone wants to take a stab at this, with a integration tests, updated man pages, please do.

eliasjpr:
@ysbaddaden this is really helpful and benefits some web projects that happen to need to run scripts on remote servers with limited access and permission.

ysbaddaden:
Well, I'd deploy pre-built binaries, not the source code, avoiding to install crystal/shards on production servers.

arnavb:
Has there been any progress on this? I'd also like to see this command implemented (it makes it much more sane to run Rust projects, from my experience).

straight-shoota:
I don't think so. But it should be rather easy to implement. If you're willing to take a stab at this, go ahead :+1:

denolfe:
Is this being worked on? If not, I'd like to take a stab at it.

denolfe:
Initial implementation here: https://github.com/denolfe/shards/commits/run-command. Any feedback is welcome.

I wrapped the build command because there needed to be some additional logic for checking for a number of targets then handling appropriately. It didn't make sense to me to shove that into `cli.cr`. Tests yet to be added.

ysbaddaden:
[...]

Issue #98, opened 2016-04-02
New command: release


miketheman:
When developing a library, it is convenient to have a method of releasing a version with well-known semantics, such as version tagging and github push.

In Ruby's Bundler, it is common to use `rake release` which performs a few actions, including a push to the remote rubygems.org site.

While Crystal Shards don't yet have a "release" location, a command like `shards release` from a working directory of a shard could validate that the current state is safe, tag the version, and push to github.

duraki:
Any update on this?

wontruefree:
`rake release` mostly just tags and pushes to rubygems.com right?
would we want shards release to do anything more?

miketheman:
@wontruefree I'd take that as a starting point. I think it does a "git push --tags" as well - to ensure that the tag actually makes it to the remote.

straight-shoota:
I don't think this makes any sense as a shards subcommand. `shards` shouldn't care about how you release your shard. There is no one-size-fits-all solution.

It is much easier to use a customized Makefile or CI script for a release process.
*Maybe* #202 could provide a way to integrate a custom release script into `shards` CLI.

caspiano:
Having a `shards release` command makes sense to me.
As `shards` uses semver and the reference is pulled from version control tags, it's a pretty limited surface to manage... but the operations happen enough that it seems routine for me, and worthy of a tool enhancement.

Something like `shards release <major|minor|patch>` that updates the `shard.yml` and attempts to tag the current commit with the updated version would be handy IMO.

straight-shoota:
The other day I read this introduction blog post for `cargo dist`: https://blog.axo.dev/2023/02/cargo-dist

A basic building block for that is `cargo release` which seems to work quite well. Cargo has a centralized registry where the release can be published to with `cargo publish`. That Crystal doesn't have, but publishing would just mean to push the tag to the public repository. So maybe this could be available as a convenience feature in a single step as well.

PR #202, opened 2018-04-25
Add a fallback to alternate shards commands

Willamin:
This allows additional functionality to be added to the shards command in a similar way extra commands are added to `git`. Basically, if the shards command isn't found in shards itself, it will run `shards-#{command}`, passing STDIO through. If there is no executable that matches the attempted command, the regular shards usage message is shown.

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -78,8 +78,8 @@ module Shards
           Commands::Version.run(args[1]? || path)
         else
           program_name = "shards-#{args[0]}"
-          if Process.find_executable(program_name)
-            run_shards_subcommand(program_name, args)
+          if program_path = Process.find_executable(program_name)
+            run_shards_subcommand(program_path, args)
           else
             display_help_and_exit(opts)
           end
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -103,7 +103,7 @@ module Shards
   def self.run_shards_subcommand(process_name, args)
     Process.exec(
       command: process_name,
-      args: args[1..-1],
+      args: args[1..],
     )
   end
 
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -77,7 +77,12 @@ module Shards
         when "version"
           Commands::Version.run(args[1]? || path)
         else
-          display_help_and_exit(opts)
+          process_name = "shards-#{args[0]}"
+          if Process.find_executable(process_name).nil?
+            display_help_and_exit(opts)
+          else
+            run_shards_subcommand(process_name, args)
+          end
         end
 
         exit
@@ -93,6 +98,20 @@ module Shards
       shards_opts = ENV.fetch("SHARDS_OPTS", "").split
     {% end %}
     ARGV.concat(shards_opts)
+  end
+
+  def self.run_shards_subcommand(process_name, args)
+    Process.exec(
+      command: process_name,
+      args: args[1..-1],
+      env: nil,
+      clear_env: false,
+      shell: false,
+      input: STDIN,
+      output: STDOUT,
+      error: STDERR,
+      chdir: nil
+    )
   end
 
   def self.build(path, args)
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -104,13 +104,6 @@ module Shards
     Process.exec(
       command: process_name,
       args: args[1..-1],
-      env: nil,
-      clear_env: false,
-      shell: false,
-      input: STDIN,
-      output: STDOUT,
-      error: STDERR,
-      chdir: nil
     )
   end
 
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -77,11 +77,11 @@ module Shards
         when "version"
           Commands::Version.run(args[1]? || path)
         else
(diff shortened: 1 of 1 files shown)


ysbaddaden:
Any use cases?

Willamin:
Extensibility. 
Sometimes subcommands don't belong in the core shards project, but would be nice to have and run under the same shards umbrella. This allows the community to create and maintain subcommands without needing approval from the maintainers of shards itself.

For example, rust has implemented a similar scheme in cargo (https://github.com/rust-lang/cargo/wiki/Third-party-cargo-subcommands). Some good subcommand examples from this list include:
  - audit (Audit lockfile for packages with security vulnerabilities)
  - graph (Build GraphViz DOT files of dependency graphs)

Additionally, I would argue that any of these are valid use cases:
  - Add/remove command (https://github.com/crystal-lang/shards/pull/192)
  - Run command (https://github.com/crystal-lang/shards/issues/157)
  - Release command (https://github.com/crystal-lang/shards/issues/98)

RX14:
I'm not all too sure of this. If your tool is separate from shards why not give it a separate command instead of stuffing it under the shards command? I imagine it's just confusing to users what's part of shards and what is not.

ysbaddaden:
Well, `git` provides this and I don't think there are much confusion.

That being said `shards-x` and `shards x` are just one letter, and strictly equivalent —except that the latter may eat up some arguments...

Willamin:
This PR's implementation should allow seamless passthrough of arguments.

straight-shoota:
The benefit of having `shards x` vs `shards-x` is that it can be used to develop features individually and later adopt them into `shards`.

ysbaddaden:
I'll consider this for post v0.8.0 release.

Sija:
We're already post v0.8.0, just a reminder ;)

Sija:
@Willamin Are you willing to continue working on this PR?

Willamin:
@Sija Yes, sorry for the delay, I'll implement the suggestions from @straight-shoota. 😄

src/cli.cr
```diff
@@ -60,14 +60,33 @@ module Shards
         when "version"
           Commands::Version.run(args[1]? || path)
         else
-          display_help_and_exit(opts)
+          process_name = "shards-#{args[0]}"
```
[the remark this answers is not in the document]
Willamin: Ah, good call. I like that name better too!

src/cli.cr
```diff
@@ -60,14 +60,33 @@ module Shards
...
+          if Process.find_executable(process_name).nil?
```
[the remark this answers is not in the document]
Willamin: Excellent suggestion 👍

src/cli.cr
```diff
@@ -67,14 +67,33 @@ module Shards
...
+    Process.exec(
+      command: process_name,
+      args: args[1..-1],
+      env: nil,
+      clear_env: false,
+      shell: false,
+      input: STDIN,
+      output: STDOUT,
+      error: STDERR,
+      chdir: nil
```
Sija, line 93: These are redundant

```suggestion
```

Willamin: Given that shards is an important tool, I opted for explicitly using these arguments rather that falling back to the default. This would avoid a problem if the defaults were ever changed. 

That being said, I'm happy to make this change if you think it's a positive change! 😄

Sija: Given that wound be a breaking change I don't see a need for being explicit that much - in the end using the defaults has some merits too (like using `:redirect` pipe instead of explicitly passing `STD*`)

RX14:
I've never actually seen a `git-foo` executable not included in the standard git. Is this feature actually used there?

jhass:
yeah. Two good examples are https://github.com/tummychow/git-absorb and http://hg-git.github.io/. If you want you can also count https://github.com/github/hub, which is supposed to be aliased to `git` and adds new subcommands if you will.

straight-shoota:
And a big collection of extra commands: https://github.com/tj/git-extras

src/cli.cr
```diff
@@ -67,14 +67,26 @@ module Shards
...
 
         exit
       end
     end
   end
 
+  def self.run_shards_subcommand(process_name, args)
+    Process.exec(
+      command: process_name,
+      args: args[1..-1],
```
Sija, line 86:
```suggestion
      args: args[1..],
```

straight-shoota (approved):

jhass (approved):

src/cli.cr
```diff
@@ -77,7 +77,12 @@ module Shards
         when "version"
           Commands::Version.run(args[1]? || path)
         else
-          display_help_and_exit(opts)
+          program_name = "shards-#{args[0]}"
+          if Process.find_executable(program_name)
+            run_shards_subcommand(program_name, args)
```
Sija, line 82:
```suggestion
          if program_path = Process.find_executable(program_name)
            run_shards_subcommand(program_path, args)
```

straight-shoota (approved):

straight-shoota:
Thank you @Willamin

PR #206, opened 2018-06-04
Crystal/0.25.0

bcardiff:
Fix compatibility for 0.25 without loosing 0.24.
Updating minitest was needed for 0.25 compatibility also.

Unless anything else is needed 0.8.0 can be tagged so the build scripts can move to that version.

### DIFF
--- a/shard.lock
+++ b/shard.lock
@@ -2,5 +2,5 @@ version: 1.0
 shards:
   minitest:
     github: ysbaddaden/minitest.cr
-    commit: 7fcb2dbba54d14ed693d76dfd6da2a4b220d1259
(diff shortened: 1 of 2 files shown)


ysbaddaden:
Thanks! Cherry-picked as 3d3a24c15d301e1df0f204b3269949c53d5a3196!

RX14:
I really don't see the benefit of macroing support in for two versions...

bcardiff:
It’s probably just for a couple of versions. And the ci depends on current release.

RX14:
Would be nice to be able to CI on nightlies...

PR #209, opened 2018-06-12
Fix path resolver recognize incorrect install path

### LINKED ISSUE #208: Path resolver doesn't fix incorrect path
`PathResolver#installed?` only checks if the install path is a symlink, not if it actually targets the specified path. 

In my case, I was using the same folder on different file systems - and thus different paths - but `shards update` didn't recognize the path didn't point to the correct location.

`shards install` did notice an outdated dependency:
```
 Outdated shard.lock (foo requirements changed). Please run shards update instead.
```

Then running `shards update` updated the path in `shard.lock` but didn't actually change the link in `lib`. However, it still reports the correct version being used: `Using foo (0.1.0)`.

I expect adding a check in `PathResolver#installed?` to verify the link target should solve all these issues.

straight-shoota:
Fixes #208

@ysbaddaden I was almost going crazy because `cleanup_install_directory` didn't remove symlinks to non-existing targets. It took my while to finally figure that `FileUtils.rm_rf` doesn't actually call the stdlib method but is a custom implementation which completely ignores broken symlinks.
I'd strongly suggest to rename that method (or the entire module) to a different name in order to avoid such mistakes.

### DIFF
--- a/test/path_resolver_test.cr
+++ b/test/path_resolver_test.cr
@@ -53,6 +53,56 @@ module Shards
       assert_raises(Error) { resolver("unknown").install }
     end
 
+    def test_installed_reports_library_is_installed
+      resolver("library").tap do |resolver|
+        refute resolver.installed?
+
+        resolver.install
+        assert resolver.installed?
+      end
+    end
+
+    def test_installed_when_target_is_incorrect_link
+      resolver("library").tap do |resolver|
+        File.symlink(git_path("legacy"), resolver.install_path)
+        refute resolver.installed?
+
+        resolver.install
+        assert resolver.installed?
+      end
+    end
+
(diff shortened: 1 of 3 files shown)


test/path_resolver_test.cr
```diff
@@ -53,6 +53,44 @@ module Shards
       assert_raises(Error) { resolver("unknown").install }
     end
 
+    def test_installed_reports_if_library_is_linked
```
ysbaddaden: Can you write a few targeted tests, instead of a big one that tests many things?

straight-shoota: sure

ysbaddaden:
Sorry, maybe `FileUtils.rm_rf` didn't exist in 2015 and it sticked.

PR #210, opened 2018-06-12
 Rename FileUtils.rm_rf to Shards::FileUtils.rm_rf_logging

straight-shoota:
See comment in #209

### DIFF
--- a/src/commands/prune.cr
+++ b/src/commands/prune.cr
@@ -13,7 +13,7 @@ module Shards
           name = File.basename(path)
 
           if locks.none? { |d| d.name == name }
-            FileUtils.rm_rf(path)
+            FileUtils.rm_rf_logging(path)
             Shards.logger.info "Pruned #{File.join(File.basename(Shards.install_path), name)}"
           end
         end
--- a/src/file_utils.cr
+++ b/src/file_utils.cr
@@ -1,5 +1,5 @@
-module FileUtils
-  def self.rm_rf(path)
+module Shards::FileUtils
+  def self.rm_rf_logging(path)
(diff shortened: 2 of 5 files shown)


test/support/factories.cr
```diff
@@ -44,7 +44,7 @@ module Shards
     def create_git_commit(project, message = "new commit")
       Dir.cd(git_path(project)) do
         run "git add ."
-        run "git commit --allow-empty -m '#{message}'"
+        run "git commit --no-gpg-sign --allow-empty -m '#{message}'"
```
ysbaddaden: Maybe useful, but unrelated to this pull request.

straight-shoota: Well, it was required to run the test suite without being asked about signing any random commits.

RX14:
Why not just specify `Shards::FileUtils` every place we use shard's fileutils. That or rename the module.

straight-shoota:
@RX14 Because you don't need to fully qualify the name inside the `Shards` namespace. I don't think it makes sense to rely on voluntarily overspecifying the name to make it obvious what this is targeting.

Using an entirely different name would be fine, though. Any suggestions? `Shards::Tools`? `Shards::Util`? Or maybe just put these methods directly into `Shards` namespace?

RX14:
`Shards.rm_rf` is fine to me.

And personally, I don't mind voluntary stuff like this for small projects.

ysbaddaden:
Or just drop it and move `#escape(path)` to helpers.

Done in125304d9ddfb2bb84e9c6ad2433de5fcdfb613ca

PR #212, opened 2018-06-17
Fix: Git resolver doesnt pull tags anymore

### LINKED ISSUE #211: Global cache issues
I just updated to 0.25 of Crystal which uses shards 0.8 with the new global cache. 

The shard cache failed to update under these circumstances:

* Defined a shard (`lucky_cli`) in `shard.yml` using `version`
* Then changed the definition to use `branch: master`
* Ran `shards update`
* Then made some changes to the `lucky_cli`
* Ran `shards update`
* Cache still used the old version of `branch: master` instead of updating

## Temporary fix

`rm -rf ~/.cache/shards` and re-run `shards update.

## Possible fix

* When running `shards update` always update cache if using `branch` or `tag`?

ysbaddaden:
Creating bare Git repositories doesn't configure the local repository to fetch remote refs, such as branches and tags, which totally broke the update command.

This patch reverts back to mirroring Git repositories, and verifies whether a cloned repository is valid, or not before fetching new refs.

fixes #211

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -181,16 +181,23 @@ module Shards
       Shards.logger.info "Fetching #{git_url}"
 
       if cloned_repository?
-        fetch_repository
+        # repositories cloned with shards v0.8.0 won't fetch any new remote
+        # refs; we must delete them and clone again!
+        if valid_repository?
+          fetch_repository
+        else
+          delete_repository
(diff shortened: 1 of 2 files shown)


paulcsmith:
Thanks for that super quick PR. Can’t wait to give this a try!

PR #223, opened 2018-08-30
Add CircleCI 2.0 with nightly builds

### LINKED ISSUE #177: CI: test master against Crystal nightly
From the sentence "Shards master tries to follow Crystal master" stems 2 issues:

1. Shards development isn't very active (it usually works well) and compatibility with Crystal master is only manually detected at some point, later after the issue, and sometimes just before a Crystal release.
2. Since the master branch follows crystal master, CI ends up being broken until the next Crystal release when a breaking change occurs.

To improve this, we should:

1. test master against a crystal nightly build —but do they exist?
2. enable a Travis cron job, to test master daily —unless a commit was issued during the day.

Follow up from https://github.com/crystal-lang/crystal/issues/5083.

bcardiff:
Fixes #177.
Adds CircleCI as an alternative CI using docker images. 

* It will use crystal:latest image on every push, and
* It will build master using crystal:nightly image every midnight
* There is not much dependencies in shards, but I've already ~~shamelessly copy~~ setup the appropriate commands for caching the global shards cache between builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ dry:
     - run: *prepare
     - checkout
     - run: make
-    - run: bin/shards install
+    - run: shards install
     - run: make test
 
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "0 0 * * *"
+          cron: "0 2 * * *"
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,31 +3,28 @@ version: 2
 dry:
   prepare: &prepare
     command: |
-      make
+      crystal --version
       git config --global user.email "you@example.com"
       git config --global user.name "Your Name"
       git config --global column.ui always
+  
+  steps: &steps
+    - run: *prepare
+    - checkout
+    - run: make
+    - run: bin/shards install
+    - run: make test
 
 jobs:
   test:
     docker:
       - image: crystallang/crystal:latest
-    steps:
-      - run: crystal --version
-      - checkout
-      - run: *prepare
-      - run: shards
-      - run: make test
+    steps: &steps
 
   test-on-nightly:
     docker:
       - image: crystallang/crystal:nightly
-    steps:
-      - run: crystal --version
-      - checkout
-      - run: *prepare
-      - run: shards
-      - run: make test
+    steps: &steps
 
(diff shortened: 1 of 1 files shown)


ysbaddaden:
I'd be fine to move completely from Travis to Circle, but does the configuration has to be so complex?

.circleci/config.yml
```diff
@@ -0,0 +1,64 @@
...
+
+workflows:
+  version: 2
+  ci:
+    jobs:
+      - test
+  nightly:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
```
RX14: I'm sure circle would appreciate it if you randomized these values instead of everyone submitting jobs at exactly midnight.

bcardiff:
@ysbaddaden the shards cache can be removed, a small bash script could be added to remove the git config, but at the end of the day circleci are more verbose than travis by it's nature.

Currently the shipped shards bin is used to grab development dependencies. Maybe generated binary from this repo should be used in the CI. Since shards itself has no dependencies that is possible.

Let me know if you prefer some of these changes to reduce the boilerplate.

Sija:
> Since shards itself has no dependencies that is possible.

@bcardiff They do, at least `development_dependencies`:

https://github.com/crystal-lang/shards/blob/0c9ac5750c4e69de8f2c3970539c022091d89287/shard.yml#L7-L9

ysbaddaden:
Sigh, Circle used to have a simpler configuration some years ago :(

ysbaddaden:
@sija yes and no. Shards doesn't have dependencies to build itself, but it needs one external library to build the test suite after that.

bcardiff:
@ysbaddaden done

ysbaddaden:
I DRYed the test steps.

I consider building `bin/shards` is a test itself (it must compile). The just built binary should also be capable to install its own test dependency (minitest), but maybe we want to delay this to the integration tests? I'm not sure.

bcardiff:
Given that there are no actual dependencies for the binary itself, using the shipped shards vs the built shards only impact in the case there is a breaking change in the format. I'm fine either way for now. But in order to change one thing at a time the state of this PR is enough IMO.

ysbaddaden:
You're right, let's use `shards` not `bin/shards` which is what Travis does.

ysbaddaden:
Merged, all that's left is to enable the crystal-lang/shards project on CircleCI :)

bcardiff:
And a small fixup in the DRY test steps commit `steps: &steps` -> `steps: *steps`. I didn't notice that issue.

PR #224, opened 2018-08-31
Remove support for obsolete Projectfile

### LINKED ISSUE #142: Drop support for empty and legacy dependencies (Projectfile)
We should print a warning in a patch release when we generate a `shard.yml` file on-the-fly when a project doesn't only has a `Projectfile` (legacy) or nothing at all that support for these cases will be removed in the next MINOR release, and one should create / bug the author that they must define a `shard.yml` file, now.

Shards is mature enough!

Sija:
Fixes #142

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -39,13 +39,8 @@ module Shards
       if file_exists?(refs, SPEC_FILENAME)
         capture("git show #{refs}:#{SPEC_FILENAME}")
       else
-        if file_exists?(refs, "Projectfile")
-          contents = capture("git show #{refs}:Projectfile")
-          dependencies = parse_legacy_projectfile_to_yaml(contents)
-        end
-
         version = version_at(refs) || DEFAULT_VERSION
-        "name: #{dependency.name}\nversion: #{version}\n#{dependencies}"
+        "name: #{dependency.name}\nversion: #{version}\n"
       end
     end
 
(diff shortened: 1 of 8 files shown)


Sija:
@ysbaddaden Is it GTG? My knowledge about the internals is limited thus would be good if you'd review it for any possible side-effects.

src/resolvers/git.cr
```diff
@@ -39,13 +39,8 @@ module Shards
       if file_exists?(refs, SPEC_FILENAME)
         capture("git show #{refs}:#{SPEC_FILENAME}")
       else
```
ysbaddaden: I suppose we shouldn't generate an empty `shard.yml` when it's missing anymore, but fail with an error instead.

src/resolvers/path.cr
```diff
@@ -12,14 +12,7 @@ module Shards
...
       else
-        projectfile_path = File.join(local_path, "Projectfile")
-
-        if File.exists?(projectfile_path)
-          contents = File.read(projectfile_path)
-          dependencies = parse_legacy_projectfile_to_yaml(contents)
-        end
-
-        "name: #{dependency.name}\nversion: #{DEFAULT_VERSION}\n#{dependencies}"
+        "name: #{dependency.name}\nversion: #{DEFAULT_VERSION}\n"
```
ysbaddaden: ditto

Sija:
@ysbaddaden Ready for re-review.

Issue #189, opened 2018-01-14
Update man pages


straight-shoota:
The man pages for `shards` and `shards.yml` have outdated information (e.g. repository URL). The commands `build`, `init`, `list`, `prune` and `version` are missing, as well as the dependency attribute `gitlab` and attributes `libraries`, `scripts` and `targets`. I've probably missed somthing.

lebogan:
Are man pages supposed to be installed with Crystal? Or, should I clone the source and build locally? I'm on Fedora 27 running Crystal 0.24.2.

Thanks.

lebogan:
I used the instructions for Fedora using dnf.

RX14:
@lebogan which ones? You mean you added a new repo and installed the package? The man pages should be in `/usr/share/man/`, so they should work...

straight-shoota:
The man pages are missing in the deb package, as well.

Judging from https://github.com/crystal-lang/distribution-scripts/blob/master/linux/Dockerfile they're probably not included in any package.

RX14:
Oh, I'm blind, this is about the shards manpage not the crystal one. Yeah that's a bug in crystal-lang/distribution-scripts.

straight-shoota:
This should be fixed by now.

PR #227, opened 2018-09-06
Extend documentation in manpage and `--help` command

straight-shoota:
Fixes part of #189

### DIFF
--- a/man/shards.1
+++ b/man/shards.1
@@ -1,50 +1,48 @@
-.TH SHARDS "1" "September 2015" "shards 0.8.1" "User Commands"
+.TH SHARDS "1" "September 2018" "shards 0.8.1" "User Commands"
 .SH NAME
 shards \- dependency manager for the Crystal Language
 .SH SYNOPSYS
 .B
 shards
-[\fIOPTION\fR]... [\fICOMMAND\fR]...
+[\fI<options>\fR...] [<\fIcommand\fR>]
+.PP
+If no \fIcommand\fR is given, \fBinstall\fR will be run by default.
 .SH DESCRIPTION
 .PP
 Manages dependencies for Crystal projects and libraries with reproducible
 installs across computers and systems.
 .PP
-Requires the presence of a \fIshards.yml\fR file. See \fBshard.yml\fR(5) for
+Requires the presence of a \fIshard.yml\fR file. See \fBshard.yml\fR(5) for
 more information on its format.
-.SH OPTIONS
-\-h, \-\-help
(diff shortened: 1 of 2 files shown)


src/cli.cr
```diff
@@ -3,36 +3,38 @@ require "./commands/*"
...
-    puts
-    puts "Options:"
+    puts <<-HELP
+      shards [<options>...] [<command>]
+
+      Commands:
+          build [<targets>] [<options>]  - Builds the specified <targets> in `bin` path.
+          check                          - Verifies all dependencies are installed.
+          init                           - Initializes a shard folder.
+          install                        - Installs dependencies from `shards.lock` file.
```
ysbaddaden: typo: `shard.lock`

man/shards.1
```diff
@@ -58,33 +56,66 @@ a requirement, but may succeed if a new dependency was added, as long as it
...
-.TP
-0
-dependencies are satisfied,
-.TP
-1
-dependencies aren't satisfied.
+Prints the current version of the shard.
+.RE
+.PP
+\fB\-\-version\fR
```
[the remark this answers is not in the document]
straight-shoota: It's intentional. `--version` and `--help` look like options, but they actually behave like commands and not as options for another command. That's why I prefer to list them under `COMMANDS`.

man/shards.1
```diff
@@ -1,50 +1,48 @@
-.TH SHARDS "1" "September 2015" "shards 0.8.1" "User Commands"
+.TH SHARDS "1" "September 2018" "shards 0.8.1" "User Commands"
 .SH NAME
 shards \- dependency manager for the Crystal Language
 .SH SYNOPSYS
 .B
 shards
-[\fIOPTION\fR]... [\fICOMMAND\fR]...
+[\fI<options>\fR...] <\fIcommand\fR>
```
ysbaddaden: Unlike `git` we don't have to specify a command since `install` is the default command. It should be `[command]` not `<command>`.

Thought, I'd be fine to *require* a command to be specified, now that the `crystal deps` shortcut has been removed.

man/shards.1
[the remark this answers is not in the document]
ysbaddaden: They trump whatever command or options are passed, but they're still options, not commands. See `man git` for example.

Can you revert that. thanks :)

man/shards.1
straight-shoota: Right, that was really just a typo and is correct in the embedded help.

PR #229, opened 2018-09-07
Format codebase with `crystal tool format`

Sija:
SSIA. These are mostly white-space changes.

There are couple of instances where I've left code unformatted, since formatter is mostly wrong in those (usually about indenting commented code).

This is the first of cleanup PRs I'd like to send.

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -92,15 +92,12 @@ end
 
 begin
   Shards.run
-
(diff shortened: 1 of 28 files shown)


ysbaddaden:
Most changes are fine. I really don't like hashes without a space separator after `{` and before `}`, thought, they look like tuples :(

PR #230, opened 2018-09-07
Use Hash composition over inheritance

Sija:
Inheriting from stdlib classes is considered a bad practice (it's my personal opinion, but there's at least https://github.com/crystal-lang/crystal/issues/3238#issuecomment-244548787 suggesting I'm not alone here).

### DIFF
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -1,8 +1,11 @@
 require "./ext/yaml"
 
 module Shards
-  class Dependency < Hash(String, String)
+  class Dependency
     property name : String
+
+    @store = {} of String => String
+    forward_missing_to @store
 
     def self.new(pull : YAML::PullParser) : self
       Dependency.new(pull.read_scalar).tap do |dependency|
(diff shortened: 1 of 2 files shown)


straight-shoota:
Yes, I think this would be a proper improvement @ysbaddaden  :+1:

Sija:
I don't agree it's _nitpicking_, and even if, that's still better than the status quo. Also, I dunno what does it have to do with other issues you mentioned. Of course it would be great to have them fixed, but why to block minor improvements over sth that has been waiting for to be done for 3 years now?

Sija:
@ysbaddaden I'd appreciate if you'd engage in any sort of merit-based conversion before closing PRs. Also, playing down even small, but still incremental improvements along with naming things like _nitpicking_ and _pedantic_ creates unwelcoming atmoshpere for all people involved.

This change for instance was seen as positive by several people already (2 PRs with exact same change were accepted, third was closed due to refactoring which took care of the root problem) so you it would be nice if you could tone-down your powerplay a bit...

ysbaddaden:
I apologize if it sounds harsh. I just don't want someone to spend much time on such changes, and prefer to close early —hinting at issues that are worth spending time on for developers and reviewers.

These classes are Hashes on purpose (they may have any kind of key/values); they do work without any issue. Legacy code accumulates, but there are much deeper changes required to make Shards a real package manager that will (in)directly involve these classes —maybe they'll be structs or just be removed.

I could shrug and merge —which I did for single vs double quotes— but this change is just opinion, doesn't change anything in my point-of-view, and let's spend time and energy on fixing issues that will require deep changes in the code base instead.

Sija:
Hi @ysbaddaden!

Because of the current design decisions, _shards_ used a library trggers Crystal ICE in certain conditions, therefore makes it impossible to use it alongside other shards (or at least one I know of - _raven.cr_), see for instance recent comment by @straight-shoota - https://github.com/crystal-lang/crystal/issues/6253#issuecomment-529153273.

Would you mind reconsidering this PR as a valid fix for the above problem - and not a _nitpicking_, like you've previously stated?

Best regards

Sija:
@ysbaddaden 🏓

ysbaddaden:
I have zero time to spend on crystal related stuff :(

I don't like the forward missing methods, it's proof the design is bad.

A real change would make Dependency and Target use explicit attributes, with custom yaml pull parsers, instead of behaving as a Hash.

Issue #218, opened 2018-07-12
New command: tree


rdp:
Sometimes I wonder why or where certain dependencies are coming from...why are they being included?
Maven has a "dependency:tree" option that might be nice here.

ysbaddaden:
Maybe as an option to the `list` command?

PR #237, opened 2018-10-08
Add --tree option to shards list

oz:
This should fix issue #218.

-----

On the other hand, it would be nice to have a more flexible options parser. It could make this code a lot cleaner too, and I'm sure some commands could be extended this way. What do you think?

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -11,7 +11,7 @@ module Shards
           check                          - Verifies all dependencies are installed.
           init                           - Initializes a shard folder.
           install                        - Installs dependencies from `shard.lock` file.
-          list                           - Lists installed dependencies.
+          list [<options>]               - Lists installed dependencies.
           prune                          - Removes unused dependencies from `lib` folder.
           update                         - Updates dependencies and `shards.lock`.
           version [<path>]               - Prints the current version of the shard.
@@ -50,7 +50,7 @@ module Shards
         when "install"
           Commands::Install.run(path)
         when "list"
-          Commands::List.run(path)
+          Commands::List.run(path, args)
         when "prune"
           Commands::Prune.run(path)
           # when "search"
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -50,7 +50,7 @@ module Shards
         when "install"
           Commands::Install.run(path)
         when "list"
-          Commands::List.run(path, args[1.. -1])
+          Commands::List.run(path, tree: args.includes?("--tree"))
         when "prune"
           Commands::Prune.run(path)
           # when "search"
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -50,7 +50,7 @@ module Shards
         when "install"
           Commands::Install.run(path)
         when "list"
-          Commands::List.run(path, args)
+          Commands::List.run(path, args[1.. -1])
         when "prune"
           Commands::Prune.run(path)
           # when "search"
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -11,7 +11,7 @@ module Shards
           check                          - Verifies all dependencies are installed.
           init                           - Initializes a shard folder.
           install                        - Installs dependencies from `shard.lock` file.
-          list [<options>]               - Lists installed dependencies.
+          list [--tree]                  - Lists installed dependencies.
           prune                          - Removes unused dependencies from `lib` folder.
           update                         - Updates dependencies and `shards.lock`.
           version [<path>]               - Prints the current version of the shard.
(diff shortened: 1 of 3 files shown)


src/cli.cr
```diff
@@ -50,7 +50,7 @@ module Shards
         when "install"
           Commands::Install.run(path)
         when "list"
-          Commands::List.run(path)
+          Commands::List.run(path, args)
```
straight-shoota: This should pass on `args[1, -1]` like in the build command.

src/commands/list.cr
```diff
@@ -22,9 +25,11 @@ module Shards
             raise Error.new("Dependencies aren't satisfied. Install them with 'shards install'")
           end
 
-          puts "  * #{_spec.name} (#{_spec.version})"
+          indent = "  " * level
+          puts "#{indent}* #{_spec.name} (#{_spec.version})"
 
-          list(_spec.dependencies)
+          indent_level = self.tree_mode ? level + 1 : level
```
straight-shoota: Don't need `self` here.

straight-shoota:
A spec would be nice. Look at `test/integration/list_test.cr`

src/cli.cr
ysbaddaden: Other commands don't take specific options, but arguments, I'd prefer the following to avoid pushing this parsing to command objects:

```crystal
class Command::List
  def self.run(@@tree = false)
  end
end

Command::List.new(tree: args.includes?("--tree"))
```

oz: I'm not sure how **you** would refactor the whole `Command::List` to do what you want though.

If we just call `Command::List.new(tree: args.includes?("--tree"))`, we lose the `path` that the abstract class expect to initialize, and well nothing works... I was not expecting to refactor all the CLI to just add `--tree`. I think it's a different PR. :sweat_smile:

straight-shoota: `path` should be there as first argument. @ysbaddaden probably just missed it.

oz: Alright, I think [that commit](https://github.com/crystal-lang/shards/pull/237/commits/95bd96a6f845ae25e4db84d153336323d284429b) is pretty much what @ysbaddaden was asking. :thinking:

src/commands/list.cr
```diff
@@ -4,15 +4,18 @@ require "./command"
 module Shards
   module Commands
     class List < Command
+      def self.run(path, @@tree = false)
```
straight-shoota: It doesn't matter much how in the current design, but setting a class variable is actually not very clean. You could just make it a local argument here and it will be forwarded to `super` and `#run`. You can just replace the splat argument `*args` with `tree`.

oz: @straight-shoota I don't know. I'm just following @ysbaddaden's request... Do you both agree on this? I'm just getting conflicting inputs from both of you. :confused:

ysbaddaden: @straight-shoota lets avoid asking for stylistic changes, especially when they conflict with previous comments, thanks.

The initial design issue is fixed: keep cli arguments parsing out of command objects.

src/cli.cr
```diff
@@ -11,7 +11,7 @@ module Shards
           check                          - Verifies all dependencies are installed.
           init                           - Initializes a shard folder.
           install                        - Installs dependencies from `shard.lock` file.
-          list                           - Lists installed dependencies.
+          list [<options>]               - Lists installed dependencies.
```
ysbaddaden: Maybe `list [--tree]` since there is only one option?

src/commands/list.cr
straight-shoota: @ysbaddaden I would do that but I don't consider this to be a stylistic issue. It's about putting data in the right place. Conceptually it doesn't make any sense as a class variable.
Also looking at consistency, no other command stores its arguments as class variables.

ysbaddaden: Ah yes, sorry, I shouldn't review on a phone, I was under the misconception that methods were class methods, when we do initialize an instance, so yeah, it should be an instance variable:
```crystal
def self.run(path, tree)
  super
end

def run(path, @tree = false)
end
```

ysbaddaden: Or better: change `Commands.run` method to pass `*args` and `**args`, so named arguments will be forwarded:
```crystal
class Shards::Command
  def self.run(path, *args, **args)
    new(path).run(*args, **args)
  end
end

class Shards::Commands::List
  def run(@tree = false)
  end
end
```

straight-shoota: `tree` still doesn't even need to be stored as an instance variable either. It's only used inside the method (or rather in `list`).

src/cli.cr
oz: Sure, fixed that.

Sija:
Seems GTG, is it a merge 🕦 ?

PR #238, opened 2018-10-11
Added possibility to install submodules of a git repository or local repository

Thellior:
<h1>Intentions</h1>
Some organisations are using a mono repository structure. They want to install shards based on a submodule of a git repository. Rails / Angular are examples of frameworks that are using the mono repo style. Because this was not supported in shards yet i adjusted the code to be able to install submodules.

<h1>What is done</h1>

- Changed the git resolver to check for the "dep" keyword in the dependencies. If there is a a dep defined it will add it to the refs of the git commands so the correct submodule is installed. If there is no dep defined it will use the normal flow that was already there.
- Changed the factory support test file to have seperate functions so we can reuse them in the new mono repo factory methods.
- Add the mono repo factory method. The factory method will initialise a mono repo with submodules. 
- Added integrations tests to verify that the submodules are installed with git and path.

<h1>Instructions</h1>

In the shards.yml you can now specify a extra attribute called "dep". The dep will describe the path to the submodule inside the git repository.

```yml
dependencies:
  logger:
    git: git@github.com:ultimategogetter/test.git
    dep: libs/logger
```

<h1>TODO</h1>

- When the change is accepted write documentation how to use this new feature

### DIFF
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -30,6 +30,10 @@ module Shards
       self["branch"]? || self["tag"]? || self["commit"]?
     end
 
+    def dep
+      fetch("dep", nil)
+    end
+
     def inspect(io)
       io << "#<" << self.class.name << " {" << name << " => "
       super
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -36,10 +36,10 @@ module Shards
       update_local_cache
       refs = git_refs(version)
 
-      if file_exists?(refs, SPEC_FILENAME)
-        capture("git show #{refs}:#{SPEC_FILENAME}")
+      if file_exists?(refs, shards_location)
+        capture("git show #{refs}:#{shards_location}")
       else
-        raise Error.new("Missing \"#{refs}:#{SPEC_FILENAME}\" for #{dependency.name.inspect}")
+        raise Error.new("Missing \"#{refs}:#{shards_location}\" for #{dependency.name.inspect}")
       end
     end
 
@@ -84,17 +84,27 @@ module Shards
       cleanup_install_directory
       Dir.mkdir_p(install_path)
 
-      unless file_exists?(refs, SPEC_FILENAME)
+      unless file_exists?(refs, shards_location)
         File.write(File.join(install_path, "shard.yml"), read_spec(version))
       end
 
-      run "git archive --format=tar --prefix= #{refs} | tar -x -f - -C #{Helpers::Path.escape(install_path)}"
+      run "git archive --format=tar --prefix= #{shard_ref(version)} | tar -x -f - -C #{Helpers::Path.escape(install_path)}"
 
       if version =~ RELEASE_VERSION
         File.delete(sha1_path) if File.exists?(sha1_path)
       else
         commit = capture("git log -n 1 --pretty=%H #{version}").strip
         File.write(sha1_path, commit)
+      end
+    end
+
+    def shard_ref(version)
+      refs = version && git_refs(version) || dependency.refs || "HEAD"
+
+      if dependency.dep
+        "#{refs}:#{dependency.dep}"
+      else
+        refs
       end
     end
 
--- a/src/resolvers/resolver.cr
+++ b/src/resolvers/resolver.cr
@@ -8,8 +8,7 @@ module Shards
   abstract class Resolver
     getter dependency : Dependency
 
-    def initialize(@dependency, @update_cache = true)
-    end
+    def initialize(@dependency, @update_cache = true); end
 
     def spec(version = nil)
       Spec.from_yaml(read_spec(version))
@@ -21,7 +20,7 @@ module Shards
       path = File.join(install_path, SPEC_FILENAME)
       return Spec.from_file(path) if File.exists?(path)
 
-      raise Error.new("Missing #{SPEC_FILENAME.inspect} for #{dependency.name.inspect}")
+      raise Error.new("Missing #{shards_location.inspect} for #{dependency.name.inspect}")
     end
 
     def installed?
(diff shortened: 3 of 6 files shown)


ysbaddaden:
Sorry, but I think enforcing 1 repository = 1 shard isn't much to ask, and is a better idea overall. If many repositories aren't possible for some reason, individual modules can be under `src` and be included with an explicit `require "foo/bar"`.

Thellior:
@ysbaddaden I understand that in some use cases the 1 repository = 1 shard is good option. For some companies or frameworks the mono repo with multiple shards is a better option for them. I think providing both options as a shards manager is a nice addition to the stack. What is holding you back from allowing both solutions?

ysbaddaden:
1. Increased complexity: dealing with branches and refs is hellish enough and the source of many bugs; I don't want to deal with shards in subfolders in addition to that.
2. No real benefit of having sub-shards: we still clone the whole repository (where rails/angular generate smaller packages).
3. I really don't like this design.

I prefer monolithic applications, but I believe libraries should be either fat (allowing explicit requires of sub-modules) or developed in their own repository if they're supposed to be autonomous.

eliasjpr:
This is a nice feature contribution for `shards`, for an organization or open source project be able to use a Monorepo style to host their shards. There some real benefits of Monorepos, this might shed some light https://www.google.com/search?source=hp&ei=f36_W5imFaaMggfTmoCwBQ&q=monorepo+advantages&btnK=Google+Search&oq=monorepo+advantages&gs_l=psy-ab.3..0.1179.7531..7693...1.0..0.95.1488.27......0....1..gws-wiz.....0..0i131j0i10j0i22i30.tiUBGOKeJoE 

I think there are some valid use cases that outweighs any personal preferences, if thats fair to say. I do not see the added complexity to this project `Shards` in 6 file change. 

If an org wants to have the complexities that comes with Monorepos I think that should be their choice, the good part is that there can be support within the official crystal tool `shards`, a tool that aims to simplify development workflow within crystal ecosystem.

straight-shoota:
I'm not sure  follow the intentions of this change completely.

Shards is based on single repository distribution. And I think that's a good thing. Changes like this introduce a lot of complexity to the application. It's "changes in 6 files" (whatever that metric is supposed to mean) and these changes need to be supported in future versions and integrated with future feature enhancements. It adds a burden on the continuing development of this software.

And frankly, I see hardly any benefit from it.

People are welcome to develop in a monorepo. Or whatever development strategy they favour. You can develop it anyway you want. You don't even need to use any SCM system for development, if you don't feel like.

But Shards is not about *development*: It's about *distribution*.

And the distribution format (currently) supported by Shards uses a Git repository for each shard. That's a relatively simple rule (but it already comes with great complexity to handle correctly).
All you have to do to distribute a shard for using it with Shards is to put it in a git repository. If you already develop it in a single git repository, you can just use that, and it's ready to go. Otherwise you'll just have to setup a Git repository as distribution channel.  This step is similar to publishing a library with any centralized dependency register like rubygems, npm etc. With shards you can be the registry yourself (or you trust any Git provider with that), you simply have to follow the distribution format: a single git repository with a `shard.yml` file.

straight-shoota:
That being said, I suppose more or less the same as this PR could already be accomplished with a configuration like this:

```yaml
dependencies:
  monorepo:
    git: git@github.com:ultimategogetter/test.git
  logger:
    path: lib/monorepo/libs/logger
```

I haven't tested it and it might not work right out of the box, but it shouldn't be very difficult to make this work. I'm not sure it's worth it considering my previous comment, but it might be a useful workaround at least during development. But you could also just put the monorepo in any other location manually and just link to its libraries with `path:`.

ysbaddaden:
If someone wants monorepo I let *them* deal with the complexity. *I* won't deal with the complexity. Sorry.

I don't want to deal with bugs such as "shards is using wrong shard.yml from monorepo subfolder" or conflict issues because they all point to the same git url, but are actually different dependencies, or worse : "I want to install different versions of shards under monorepo"...

Testing the happy path isnt enough, I'm afraid most integration tests should be duplicated to test that they do work with monorepos. Implementing new features would mean to take care that they do work with monorepos, and so on :-(

I'm closing. I don't want to give the impression that we'll support monorepos.

Maybe `path` could be fixed to accept relative paths, but one can just create symlinks manually under `lib` and commit them, so I'm not even sure.

RX14:
Also, since we base versions off git tags, and git tags are global to a whole repo, every shard sharing a repo would need to have the same version which is impractical. I don't see how this could be reasonably implemented.

eliasjpr:
Versioning is something that the Monorepo try to address. 

I understand if this is something that Shards and Community does not want to support.

eliasjpr:
@ysbaddaden you could have left this PR open or move it to a thread, maybe it would've lead to a solution that is simply, efficient and make everyone happy. Closing it just shuts down any possibility and gives an impression that people thoughts and arguments are not enough to persuade the maintainers of this open source project. We do not have to agree but a least allow for the conversation flourish so we can arrive to a conclusion.

jwoertink:
I know this is an old issue, but wanted to bring this up again as I just ran in to this. I have a repo that contains more than just crystal. I put my shard in to a subfolder of the repo, but I can't link to it without getting this error since the `shard.yml` isn't at the top level of the repo.

> Failed git ls-tree -r --full-tree --name-only refs/tags/v0.1.0 -- shard.yml (). Maybe a commit, branch or file doesn't exist?

When if the key just specified where the location of the shard.yml was? Maybe something like:

```yml
dependencies:
  thing:
    github: jwoertink/thing
    branch: master
    require: subdirectory/
```

Where the shard.yml is located in a folder called "subdirectory". Then in this case, we could have a single repo that contains multiple shards by doing something like

```yml
dependencies:
  orm:
    github: jwoertink/monorepo
    branch: master
    require: orm/
  logger:
    github: jwoertink/monorepo
    branch: master
    require: logger/
```

Where the structure is like:

```
orm/
  src/
    orm.cr
  shard.yml
logger/
  src/
    logger.cr
  shard.yml
README.md
```

Obviously the key name `require` could be whatever, but I think this wouldn't be a major change, and would support the ideas of this PR.

PR #244, opened 2018-12-26
Remove CLI from main file shards.cr

straight-shoota:
When using `shards` as a library, this lets `require "shards"` not automatically launch the CLI anymore.


ysbaddaden:
Something broke.

ysbaddaden:
All crystal distribution scripts shall be checked, too, as they may assume that `src/shards.cr` is the main executable source file (actually, it is).

straight-shoota:
Oh yes, I was missing the Makefile.

RX14:
I suggest making compiling the old file fail somehow so that all the distros which package shards don't produce buggy packages.

straight-shoota:
> Making `src/shards.cr` fail to compile defeats the very purpose of the pull request.

That's only temporarily to alert package maintainers to fix this. In the next release (or the one after) we can add `shards` again.

Yes, shards is an application, but its backend is also useful as a library for other applications for dealing with `shards.yml`, extracting versions, resolving dependencies. In fact, it is really important to use the implementation of `shards` for this in order to be consistent.

That's specifically the reason why I'm providing this patch, because I currently can't simply `require "shards"`. Instead I need to require each `shards/config`, `shards/logger`, `shards/errors` and `shards/version` just to be able to use any code from shards.

This code organization jumble is not user friendly.

ysbaddaden:
Only the `shards` binary has a stable interface. Anything internal will break at anytime in minor and patch releases. Until some internal API becomes stable (if it ever), having `require "shards"` *not working* as expected is a good reminder that you're not dealing with a proper library.

straight-shoota:
> having require "shards" not working as expected is a good reminder that you're not dealing with a proper library.

Then it should nevertheless fail to require, not require and start a CLI when run.

IMHO it's bad design to name a file that can't be used as a library the same as the shard, thus making it resolve to `require "shardname"`.

straight-shoota:
And I'd very much like to get the internal API stable to make it reusable for related features. I've been hacking at a shards database and that should be able to depend on `shards` as a library in order to ensure the same behaviour when discovering releases for example.

ysbaddaden:
A stable library API may come, but it won't in the short term. I'd like to keep `src/shards.cr` as the entrypoint for the `shards` binary. it's an exact naming match, and using something else like `src/cli.cr` is just weird.

We can, however, make dependencies explicit in different files, to ease their usage "as a library". For example `require "shards/spec"` or `require "shards/resolve/git_resolver"` could just work.

PR #245, opened 2018-12-27
Force git commits created by the spec suite not to be signed

RX14:
In the case of the user running the tests setting the git configuration commit.gpgSign=true, running the specs can prompt for a smartcard to be inserted or for the keychain to be unlocked. This is undesirable.


straight-shoota:
This has bugged me as well. :+1:

Issue #108, opened 2016-05-27
git command failed: git ls-tree


kostya:
``` yaml
name: bla
version: 0.1.0

dependencies:
  http_parser:
    github: kostya/http_parser.cr

license: MIT
```

```
crystal deps
git command failed: git ls-tree -r --full-tree --name-only v0.8 -- shard.yml (). Maybe you didn't install git?
```

```
git --version
git version 2.8.3
```

if i add line `branch: master` it works.

ysbaddaden:
The problem are the tags. Shards is looking for a `v0.8` tag but it can't be found because it's `0.8` (missing a leading `v`).

I tried to support both namings (with or without a leading `v`), but it complicates the resolver.

jhass:
Might still be worth to attempt to improve the error output here.

ysbaddaden:
Definitely.

I don't think we should parse the Git message. Different versions may have different messages. Maybe buffer and display the command output in addition to the command?

jhass:
Yeah, also I think a general "does git work" test before the first of any git operations with an appropriate error message would be good, and then changing the error message from "does your git work?" here to something else.

kostya:
why not add, branch: master by default?

ysbaddaden:
Because shards target the last release by default.

It couldn't find version tags, but it found the `shard.yml` in master that specifies a version, and thus it tried to find this version. Maybe this is an odd behavior of an edge case (the version tags should exist).

straight-shoota:
The error message was improved a little bit, but the actual issue still exists: shards breaks when a git tag doesn't start with `v`.

As the issues linking here show, this is an ongoing problem. And it's not just the shard's maintainer's fault (they could just use `v` tags), because `shards` *accepts* plain tags but can't treat them properly.

When reading the versions from git tags, in `Shards::RELEASE_VERSION` the leading `v` is optional and stripped off. However, `#git_refs(version)` prepends a `v` unless the version already starts with it (which should rarely be the case because it's stripped off).

This could be solved by enforcing git tags to start with `v`. This would at least fail clean and could provide a meaningful error message. It's not a good solution, though, because people *are* using plain git tags without `v` prefix. They're also already in the history of many shards.

I think this could be solved by storing the version as a git tag in it's original form, without stripping, and only convert to a non-prefixed string when it's used externally. Maybe a simple `GitTag` wrapper struct could be introduced to make this relatively straightforward.

RX14:
[...]

PR #247, opened 2018-12-27
Enforce prefix 'v' in git version tags

straight-shoota:
Shards recognizes tags without 'v' prefix as valid version identifiers but then can't use them as valid ref for accessing it in the git repo. That leads to errors.

Maintainers of shards need to prefix release version tags with a `v`. This can be fixed retroactively.
A useful shell command for adding v-prefixed tags:
```bash
for tag in $(git tag --list | grep -P '^[\d.]+$'); do git tag v$tag $tag; done
```

This issue has been discussed in #108

### DIFF
--- a/test/git_resolver_test.cr
+++ b/test/git_resolver_test.cr
@@ -8,6 +8,9 @@ module Shards
 
       create_git_repository "library", "0.0.1", "0.1.0", "0.1.1", "0.1.2"
       create_git_release "library", "0.2.0", shard: "name: library\nversion: 0.2.0\nauthors:\n  - julien <julien@portalier.com>"
+
+      # Create a version tag not prefixed by 'v' which should be ignored
+      create_git_tag "library", "99.9.9"
     end
 
     def test_available_versions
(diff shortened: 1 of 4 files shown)


straight-shoota:
Ah yes, that's even better.

test/git_resolver_test.cr
```diff
@@ -8,6 +8,11 @@ module Shards
 
       create_git_repository "library", "0.0.1", "0.1.0", "0.1.1", "0.1.2"
       create_git_release "library", "0.2.0", shard: "name: library\nversion: 0.2.0\nauthors:\n  - julien <julien@portalier.com>"
+
+      Dir.cd(git_path("library")) do
+        # Create a version tag not prefixed by 'v' which should be ignored
+        run "git tag 99.9.9"
+      end
```
ysbaddaden: Just one tiny thing: could you add a `create_git_tag(project, tag)` helper in `test/support/factories.cr`? then could it `create_git_tag "library", "99.9.9"`?

That would be perfect :)

Issue #184, opened 2017-12-07
Error resolving pre-release version


vladfaust:
I'm trying to use meta versions as described in SemVer. My lib has `v0.2.0-pre1` tag.

My app's shard looks like this:

```yaml
mylib:
  github: username/mylib
  version: 0.2.0-pre1
```

Getting `Error resolving mylib (0.2.0-pre1)`. What's wrong?

vladfaust:
What's the right way to define pre-release versions in shards?

From https://semver.org:

> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

vladfaust:
I've also tried `v0.2.0-alpha` - no luck

vladfaust:
ping @ysbaddaden (sorry, but this *is* important) Just tell me what am I doing wrong...

ysbaddaden:
Sorry, that's probably not supported.

ysbaddaden:
Fixing the detection of release tags, is easy (a mere regular expression), but then the pre-releases, which are unstable, would become the latest release for shards, which must be changed to distinguish stable releases from unstable pre-releases, default to releases, but still allow installing specified pre-releases, and probably add some command line argument to enable pre-releases by default... I.e. a lot of work.

You may use `tag: v0.2.0-pre1` for the time being.

straight-shoota:
I'm willing to work on this.

I think a good solution would be to introduce a `Shards::Version` type to combine all the logic related to parsing and interpreting version strings.
Rubygem's [`Gem::Version`](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/version.rb) could serve as a guideline, including `VERSION_PATTERN = '[0-9]+(?>\.[0-9a-zA-Z]+)*(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?` and `#prerelease?` predicate. The logic is quite simple: Any version string with a letter in it is considered a pre-release. This works with semver as well as other versioning schemes (for example `1.1.alpha`).

RX14:
Why not just parse the versions using the stdlib's semver parser and handle them properly instead of using regexes?

straight-shoota:
Because `SemanticVersion` strictly expects a string complying to the semver spec. But shards explicitly doesn't enforce using semver.

[...]

PR #248, opened 2019-01-02
Add version_restriction_test

straight-shoota:
Also transform helper methods to class methods for better isolation and testability.

This is a preliminary step for implementing #184

### DIFF
--- a/src/commands/check.cr
+++ b/src/commands/check.cr
@@ -4,8 +4,6 @@ require "./command"
 module Shards
   module Commands
     class Check < Command
-      include Helpers::Versions
-
       def run(*args)
         if has_dependencies?
           locks # ensures that lockfile exists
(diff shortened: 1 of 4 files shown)


ysbaddaden:
Adding tests is great. Breaking the includable modules for class methods is an opinionated stylistic change. Without further commits to support this change, it's hard to understand and validate it.

If there is a bug in NaturalSort, a fix would be very welcome.

Last but not least, the unit test file should symmetric to the file being tested. For example `test/helpers/versions_test.cr`

straight-shoota:
It may be stylistic, but it's foremost practical as it makes these methods easier to test. I don't see why breaking the includes is a bad think, if anything it clarifies the API.

There is a bug in resolving `~>` restrictions. I won't fix `NaturalSort` but propose a different algorithm instead (which doesn't have this bug).

From my perspective this PR is a simple improvement, self-contained and easy to review. It stands by itself and shouldn't depend on any planned changes.

Moved the file.

ysbaddaden:
Thanks for the tests. They were applied in #249.

PR #249, opened 2019-01-08
Implements pre-releases

### LINKED ISSUE #183: Dependencies are satisfied rarely works
Hi @ysbaddaden @RX14 

Always I execute `shards build` to build my crystal projects and it works very well, However shards is fetching my depencencies everytime I execute it, first time is slower because is cloning repositories, next times are a bit faster but still take some time.

I'm wondering why it doesn't show something like:

```
➜ shards build project
Dependencies are satisfied
Building: project
```

Almost always I see:

```
➜ project shards build project
Fetching https://github.com/amberframework/amber.git
Fetching https://github.com/luislavena/radix.git
Fetching https://github.com/jeromegn/kilt.git
Fetching https://github.com/jeromegn/slang.git
Fetching https://github.com/stefanwille/crystal-redis.git
Fetching https://github.com/mosop/cli.git
Fetching https://github.com/mosop/optarg.git
Fetching https://github.com/mosop/callback.git
Fetching https://github.com/mosop/string_inflection.git
Fetching https://github.com/mosop/teeplate.git
Fetching https://github.com/samueleaton/sentry.git
Fetching https://github.com/juanedi/micrate.git
Fetching https://github.com/jwaldrip/shell-table.cr.git
Fetching https://github.com/askn/spinner.git
Fetching https://github.com/crystal-lang/crystal-mysql.git
Fetching https://github.com/crystal-lang/crystal-sqlite3.git
Fetching https://github.com/amberframework/granite-orm.git
Fetching https://github.com/amberframework/quartz-mailer.git
Fetching https://github.com/raydf/smtp.cr.git
Fetching https://github.com/amberframework/jasper-helpers.git
Fetching https://github.com/amberframework/garnet-spec.git
Fetching https://github.com/ysbaddaden/selenium-webdriver-crystal.git
Using amber (0.3.6)
Using radix (0.3.8)
Using kilt (0.4.0)
Using slang (1.7.0)
Using redis (1.9.0)
Using cli (0.6.10)
Using optarg (0.5.8)
Using callback (0.6.3)
Using string_inflection (0.2.1)
Using teeplate (0.4.5)
Using sentry (0.1.1)
Using micrate (0.3.0)
Using db (0.4.3)
Using shell-table (0.9.2)
Using spinner (0.1.1)
Using pg (0.13.4)
Using mysql (0.3.3)
Using sqlite3 (0.8.3)
Using granite_orm (0.7.8)
Installing quartz_mailer (0.2.0 at 0.2.1)
Installing smtp (0.1.0 at 0.1)
Using jasper_helpers (0.1.6)
Using garnet_spec (0.1.1)
Using selenium (0.3.0)
Building: project
```

Even if my shards are already satisfied

Versions:

```
➜  project shards --version
Shards 0.7.2 (2017-11-23)
➜  project crystal --version
Crystal 0.23.1 (2017-09-10) LLVM 5.0.0
```

### LINKED ISSUE #184: Error resolving pre-release version

ysbaddaden:
I wanted to fix the issue found by tests in #248 but actually refactored `Helper::NaturalSort` and `Helper::Versions` as a single `Versions` module, responsible for comparing and sorting version numbers correctly:

- support for unbalanced version numbers (`1.0.0 == 1.0` and `1.0.1 > 1.0)`;
- support for pre-release version numbers;
- fixed approximate resolver where `~> 0.1.0` matched `0.10.0`.

Supporting pre-releases in install/update commands was then just extracting pre-release versions correctly, and not filtering them out unless a requirement specifically requires one.

closes #184 
fixes #183


ysbaddaden:
Also happens to fix #183 where unbalanced version numbers since 1.0 and 1.0.0 are now considered the same.

test/versions_test.cr
```diff
@@ -0,0 +1,149 @@
...
+      assert_equal -1, Versions.compare("1.0.0.beta", "1.0.0.alpha")
+      assert_equal -1, Versions.compare("1.0.0", "1.0.beta")
+      assert_equal -1, Versions.compare("1.0.0-beta", "1.0.alpha")
+
+      assert_equal -1, Versions.compare("1.0-pre2", "1.0-pre1")
+      assert_equal -1, Versions.compare("1.0-pre10", "1.0-pre2")
+    end
+
+    def test_sort
+      100.times do
```
straight-shoota: Why so many randomized iterations? The comparisons between items always stay the same, so this doesn't provide any additional coverage.
You don't get any more complete result than iterating through the sorted list and asserting that every item compares lower than its successors.

straight-shoota:
Hm, I was already working on a patch for this 🤔 That's why I assigned #184 to myself.

Both result and key implementation are pretty similar in your approach, so :+1:

test/versions_test.cr
ysbaddaden: The original NaturalSort had a bug where the initial order influenced the output ordering. Adding randomization catched it. 100 is a magic number, chances of not catching the issue is veeery low, and this has no impact on speed test (fast).

straight-shoota: It just has some smell to use randomized brute force specs 😄

src/config.cr
```diff
@@ -5,6 +5,9 @@ module Shards
   DEFAULT_COMMAND = "install"
   DEFAULT_VERSION = "0"
 
+  VERSION_REFERENCE = /^v?\d+[-.][-.a-zA-Z\d]+$/
+  VERSION_TAG = /^v(\d+[-.][-.a-zA-Z\d]+)$/
```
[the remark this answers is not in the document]
ysbaddaden: Hum, maybe, but then they'd be named `Versions::REFERENCE` and `Versions::TAG` with a plural which sounds weird. I think I prefer global constants.

ysbaddaden:
Sorry for duplicating work. I went overboard on some free time I had :bow:

src/versions.cr
```diff
@@ -146,10 +146,19 @@ module Shards
       end
     end
 
+    def self.prerelease?(str)
+      str.each_char.any?(&.ascii_letter?)
```
straight-shoota: Hyphens are often also used to determine pre-releases. Rubygems considers any version with a letter or hyphen as pre-release.
This obviously only makes a difference when the pre-release tag doesn't contain any letters (like `0.1-1`), which is probably pretty rare.

ysbaddaden: Does it really? Sigh. `0.1.1-1` looks like a build number that supersedes a `0.1.1` release, it doesn't look like a pre-release to me.

Shall we really support that?

ysbaddaden: Rubygems doesn't look for a hyphen, only for a letter:
https://github.com/rubygems/rubygems/blob/master/lib/rubygems/version.rb#L290-L297

straight-shoota: I'm not sure if this needs to be supported. But rubygems made extra effort to achieve this, so maybe they have a good reason?

Anyway, it's really trivial to implement here, so why not?

ysbaddaden: Sigh. I really start to hate semver altogether. The first draft was okay, but it's really become ugly.

ysbaddaden: > it's really trivial to implement here

Actually, it's not, because dashes and dots are equivalent, so `1.0.0-1` == `1.0.0.1`. I must use the same trick than rubygems and replace `-` with a `.pre.` so `1.0.0-1` becomes `1.0.0.pre.1` to be considered a prerelease, and thus older to `1.0.0`.

ysbaddaden: The replace leads to oddities, like `0.0.1-alpha > 0.0.1.beta` because we parse `0.0.1-alpha` as `0.0.1.pre.alpha` but don't transform `0.0.1.beta`.

We should probably check the partition result (dot or dash) in addition to the segment, but... I honestly, I'm tired, here.

test/versions_test.cr
```diff
@@ -0,0 +1,149 @@
+require "./test_helper"
+
+module Shards
+  class VersionsTest < Minitest::Test
+    def test_compare
+      # a is older than b:
```
straight-shoota: I would't use older and newer here. Version comparison is not temporal.

src/versions.cr
RX14: If we don't currently support semver perfectly then we don't have to support it perfectly in this PR.

ysbaddaden:
I added a few tests, and also ignore semver metadata. I failed to add support for prereleases made of only numbers. Still, it's better than the current one, and there shouldn't any *regression* since the overall algorithm is the same.

src/versions.cr
Sija: btw, `n.n.n-n` notation is used by homebrew, where last part denotes _revision_, far from being pre-release.

straight-shoota: @Sija This is pretty common with any package distribution. Crystal's apt and rpm packages also use this release + iteration format.
But in semver, everything following a hyphen makes a pre-release. We can't fit both. But I agree with @RX14.

So every version containing a letter is considered a pre-release.

CHANGELOG.md
```diff
@@ -2,7 +2,15 @@
 
 ## UNRELEASED
 
+Features:
+- Experimental support for prereleases. Add a letter to a version number to
```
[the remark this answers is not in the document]
ysbaddaden: I mean experimental as in "works on my computer, not tested in the wild".

ysbaddaden:
You're welcome —even if it took 2 years.

PR #256, opened 2019-01-14
Fix: transitive dependencies in postinstall script

### LINKED ISSUE #124: Compiling Crystal code in postinstall

ysbaddaden:
Postpones the execution of the postinstall script, so it's run after all dependencies have been installed, to solve the availability of transitive dependencies. Also postpones the installation of executables, since they're likely to be built by the postinstall script.

Links the project's `lib` folder as the shard's own `lib` folder, to solve the accessibility of transitive dependencies (for executable for the postinstall script). This solution was preferred over manipulating the CRYSTAL_PATH environment variable, since a wrapper script of the crystal executable could replace it (e.g. the `bin/script` of crystal's repository does just that).

closes #124

### DIFF
--- a/src/commands/install.cr
+++ b/src/commands/install.cr
@@ -44,7 +44,13 @@ module Shards
       end
 
       private def install(packages : Set)
-        packages.each { |package| install(package) }
+        packages
+          .compact_map { |package| install(package) }
+          .each(&.postinstall)
+
+        # always install executables because the path resolver never installs
+        # dependencies, but uses them as-is:
+        packages.each(&.install_executables)
       end
 
       private def install(package : Package, version = nil)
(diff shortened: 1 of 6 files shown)


RX14:
An integration test would be nice

ysbaddaden:
Now with tests and both install & update commands are correct.

PR #257, opened 2019-01-15
Add outdated command to know which dependencies are outdated

### LINKED ISSUE #56: New command: outdated
The command shall list the outdated shards (and only them) along with the installed version, the available version (matching the requirements) and the latest version.

```
$ shards outdated
pg (0.3.0, available: 0.3.2, latest: 0.4.0)
minitest (0.1.3, available: 0.1.5)
```

ysbaddaden:
Basically: it resolves dependencies just like the `install` command does, then reports any dependency that has a new release matching all constraints (`available: x.y.z`), and the latest release if it's different.

Optionally supports checking for pre-releases with `--pre`.

The command assumes that dependencies are installed. It doesn't fail if `shard.yml` is missing; maybe it should?

The command is meant for developers to know about new releases of dependencies, and thus always succeeds (i.e. `exit 0`) unless there was an error.

closes #56

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -12,6 +12,7 @@ module Shards
           init                           - Initializes a shard folder.
           install                        - Installs dependencies from `shard.lock` file.
           list [--tree]                  - Lists installed dependencies.
+          outdated [--pre]               - Lists dependencies that are outdated.
           prune                          - Removes unused dependencies from `lib` folder.
           update                         - Updates dependencies and `shards.lock`.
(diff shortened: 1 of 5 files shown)


AlexWayfer:
I think, this could be in CHANGELOG.

ysbaddaden:
It is, see v0.9.0.rc1

AlexWayfer:
> It is, see v0.9.0.rc1

OK, there are separated Changelogs for beta and stable versions, excuse me.

PR #259, opened 2019-01-23
SAT solver

### LINKED ISSUE #1: Smarter dependency resolver
The current dependency resolver is dumb.

It takes the current project `shard.yml` then clones dependencies, extracts the available versions and takes the `shard.yml` for the latest version matching the requirement (or HEAD).

It then does the same for the dependencies of each dependency, pushing new dependencies or requirements to a flat list of packages, and so on, recursively.

If any package in the flattened list packages and requirements can't eventually be resolved to at least one version, the resolver will fail by raising a conflict.

This resolver should work for quite some time, until Crystal starts having lots of libraries with lots of dependencies, in which case it will quickly fail.

A smarter resolver will then be appreciable.

### LINKED ISSUE #58: Only update the specified shards
It would be nice to be able to only update a specified set of shards, instead of everything. The update command would also do its best to only update nested dependencies to satisfy new requirements, keeping as much as possible the locked versions.

```
$ shards update mock
Using library
Using extra
Updating mock
```

ysbaddaden:
Crystal port of the [simple-sat](https://github.com/sahands/simple-sat/) solver (python). We could probably achieve better performance (using glucose or minisat) but it's fast enough for the time being. We could also switch to a [CDCL]() algorithm but the current [DPLL]() one has acceptable performance, so this can come later.

Adds a `lock` command ~~that may eventually~~ will persist ~~or not~~ as a smart and interactive solution to create a lock file.

The command digs all specs for all dependencies (advice: use `--local` once they're in cache) and recursively applies ~~the main spec~~ dependency requirements to limit the dependency graph ~~a bit~~ to reachable dependencies. ~~Some additional work could be done to skip more impossible solutions (limits generated SAT clauses) but that doesn't seem to be performance/limiting factor for the time being.~~ This significantly speeds up the dependency generation, greatly reduces the amount of generated clauses, and computing a solution is much faster (e.g. from 6s down to 1/2s for a lucky app).

Once the dependency graph is built, it generates SAT clauses for them, calculates an ideal (latest versions, as close to the locked versions as possible, a mix of both) then goes to solve the problem, chooses the best solution (as close to ideal as possible) and eventually ~~reports how many solutions were found~~ generates a lock file or prints conflicts if it failed. ~~Small projects with strict version constraints should have a handful of solutions, larger projects like a Lucky app will have nearly a million of solutions. The total is exponential to how many versions of a dependency can be installed —if I can select a dozen versions of A and B and C dependencies, I immediately get 42840 solutions.~~

~~The solver probably misses:~~
- ~~more clauses to limit the number of solutions or to quickly reach the best one~~ done;
- ~~multiple runs, first with a stricter list of clauses, try to solve it, try a looser list, solve again, etc.~~ not required;
- ~~hints to use given versions of a dependency first, to try and get a best solution sooner than later~~ not required;
- ~~stop solving when we reach a best-enough solution (or be interactive)~~ not required.

~~Anyway, each solution should be given a mark depending on what we want to achieve, like being up-to-date (everything at the latest version) or a conservative update (least changes compared to lockfile).~~ Done: Shards defines an ideal version (latest versions and/or locked versions), computes the distance from the ideal for each solution, and eventually uses the closest, or the one with less dependencies in case of a tie.

~~There is no analysis/report of why the graph couldn't be solved. It means there is a *conflict* somewhere, but we don't know where and what caused it (yet).~~ Done: conflicting clauses are now printed in plain English (can't install version X and Y at the same time, A depends on B, ...).

close #1
close #58


src/sat.cr
```diff
@@ -0,0 +1,207 @@
...
+          if a == 0
+            next if brief
+            str << '~'
+          end
+          str << v << ' '
+        end
+      end
+    end
+
+    protected def to_assigment(assignment : Array(Literal?), brief = false)
```
ysbaddaden: This method is super slow and must be optimized. The solve method below can find 1 million solutions out of 300 variables and 5000 clauses in 0.8s (release mode) which is acceptable performance; calling this method to transform the assignment array to an array of variables easily takes more than 10s :fearful:

RX14:
I'm worried about error messages. It seems most package managers end up using simple backtracking solvers - there must be a good reason for this.

RX14:
https://research.swtch.com/version-sat is a good read on this problem, if we did enforce semver we'd have a simple solution: generate a list of all dependencies and their constraints, check if any of the constraints have differing major versions, if so fail, else choose the latest version of that major for all dependencies. i.e. the only possible dependency constraint you could specify would be "greater than this minor release but still the same major". I don't think that's too bad. Or we could choose the earliest version like vgo does.

RX14:
Another good article with a paper linked, this one solves the full problem shards is trying to solve, and generates good error messages: https://medium.com/@nex3/pubgrub-2fb6470504f

ysbaddaden:
This is an experiment to try a SAT solver, to see whether it's viable or not, if it can lead to useful results or that it won't. Maybe it won't make it, maybe it will be kept as a smart interactive alternative to a simpler backtracking solver for edge cases.

The `dep-selector` gem uses a SAT solver and detects and reports conflicts. Same for other package managers. I'm using a simple SAT solver that I don't quite understand yet, and still have to dig that.

Enforcing semver will solve nothing. This is basically what Shards does currently: select the latest release and hope that there won't be any conflict in nested dependencies, otherwise it will fail stupidly, and tell developers to fix it, whatever if there was a solution (that a SAT solver will find).

RX14:
the pubgrub solution seems the best since it's well-tested in exactly this problem domain and has a very nicely written design doc on the algorithm. And it doesn't require breaking the current dependency model in shards.

ysbaddaden:
How am I breaking the current dependency model?

PubGrub looks really nice. I'm pleased to see that I'm on track so far, but using the simpler [DPLL](https://en.wikipedia.org/wiki/DPLL_algorithm) algorithm, instead of the [CDCL](https://en.wikipedia.org/wiki/Conflict-driven_clause_learning]) algorithm that PubGrub uses.

I'll eventually work on failures to solve a dependency graph, and try to report the exact incompatibilities correctly, while allowing to speed things up with a CDCL solver instead of DPLL. For now, I'm focusing on the can-be-solved-despite-incompatibilities paths and how a SAT solver will be helpful in a number of scenarios: initial resolve, adding/removing a dependency, updating everything to their latest versions, update of a few selected dependencies (least changes to the graph) and so on.

RX14:
> How am I breaking the current dependency model?

Sorry for the confusion, this was in reference to my own suggestion of enforcing semver.

Having found PubGrub, I agree that working on SAT is the way forward. My previous posts were made while I was still doing research on the problem.

ysbaddaden:
Ah, I understand better :)

ysbaddaden:
Now with proper dependency graph builder that only collects reachable dependencies & versions (and should support git refs) and failure reporting:

screenshot2

ysbaddaden:
A bunch of new commits that extract a `Solver` class out of the `Commands::Lock`, makes the former smarter (by using locked dependencies) and makes the latter actually generate a lockfile (without installing).

This `Solver` class will eventually replace a lot of now useless methods and even classes (`Package`, `Manager`, ...) when it will be used in the `install` and `update` commands.

src/commands/outdated.cr
```diff
@@ -3,63 +3,67 @@ require "./command"
...
+        solver = Solver.new(spec, @prereleases)
+        solver.prepare(development: !Shards.production?)
 
-        if @up_to_date
-          Shards.logger.info "Dependencies are up to date!"
+        if packages = solver.solve
+          packages.each { |package| analyze(package) }
+
+          if @up_to_date
+            Shards.logger.info "Dependencies are up to date!"
```
Sija: Why no block version as in other places? ditto below.

src/solver.cr
```diff
@@ -0,0 +1,220 @@
...
+      @graph = Graph.new
+      @sat = SAT.new
+      @solution = nil
+      @solution_distance = Int32::MAX
+    end
+
+    def prepare(development = true) : Nil
+      @graph.add(@spec, development)
+
+      if locks = @locks
```
Sija: One scope less with (ignore me if you feel that's too pedantic ;)):

```cr
@locks.try &.each do |lock|
  # ...
end
```

src/commands/outdated.cr
ysbaddaden: It's a mere static string, it doesn't allocate anything.

src/solver.cr
ysbaddaden: Tried it but it's slightly harder to read because of eventual block.

ysbaddaden:
I added the missing tests. Should be :green_apple: and ready to :boat:

PR #260, opened 2019-01-29
Fix: some versions issues

ysbaddaden:
The first patch is a simple bug fix:  don't consider a version number such as `1.0+meta` as a prerelease (i.e. skip metadata);

The other one is a convenience, it maps a version tag definition to be the actual version requirement in dependencies. In the following example, the `tag: v0.12.0-rc1` is identical to `version: 0.12.0-rc1`:
```yaml
dependencies:
  lucky:
    github: luckyframework/lucky
    tag: v0.12.0-rc1
```

This is a nice to have because with Shards < 0.9 this was the only way to declare a pre-release dependency.

### DIFF
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -12,18 +12,23 @@ module Shards
       end
     end
 
-    def initialize(@name)
+    protected def initialize(@name)
       super()
     end
 
-    # DEPRECATED: with no replacement
-    def initialize(@name, config)
+    protected def initialize(@name, config)
       super()
       config.each { |k, v| self[k.to_s] = v.to_s }
     end
 
     def version
-      fetch("version", "*")
+      if version = self["version"]?
+        version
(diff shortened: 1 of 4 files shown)


src/dependency.cr
```diff
@@ -12,18 +12,23 @@ module Shards
       end
     end
 
-    def initialize(@name)
```
asterite: Note that `protected` is the default visibility for `initialize`, so this is a bit redundant.

ysbaddaden: And the visibility isn't reflected on the `new` constructor?

asterite: No... but it's a good point, now that you mention it. What works is using `private`, but then you can't access it from within the same namespace. Hmm...

ysbaddaden:
Merged into #259

Issue #253, opened 2019-01-11
Rewrite manpages in asciidoc


ysbaddaden:
It would be nice to rewrite the manpages in asciidoc and to build them using the [asciidoctor](https://asciidoctor.org/) gem. It's nicer to write in, and that would allow to have different output formats (man, html, md, ...). Git does that and I really like it.

straight-shoota:
I've just manually transformed the man page to Markdown 😄 (crystal-lang/crystal-book#316).
Having a single source format would definitely help to keep these instances in sync.

ZaWertun:
Calling `asciidoctor` with empty `SOURCE_DATE_EPOCH` env variable gives this error:
```bash
SOURCE_DATE_EPOCH= asciidoctor -a shards_version=0.14.0 docs/shards.adoc -b manpage -o man/shards.1
asciidoctor: FAILED: /home/zawertun/tmp/shards/docs/shards.adoc: Failed to load AsciiDoc document - invalid value for Integer(): ""
Use --trace for backtrace
```
(I'm build from shards from the tar.gz archive, not from the git repo)

straight-shoota:
That makefile call is completely broken. Sorry, this was a really crappy copy and paste 😢 

Fix is in #479.

szabgab:
There are two man pages committed in the man/ that are removed when you run `make clean`. Should those be removed from the repository or are they still needed? If they are still needed then probably make clean should not remove them.

straight-shoota:
For the time being, they're checked in as a convenience so you don't have to install asciidoctor to read the manpages.

So yeah, probably they shouldn't be cleaned. Feel free to make a PR. It's low priority, though.

PR #262, opened 2019-02-11
Rewrite manpages in Asciidoc

straight-shoota:
This translates the manpages to Asciidoc as suggested in #253

It's a simple translation and tries to keep close to the original. Further enhancements are probably coming.


Makefile
```diff
@@ -20,12 +20,15 @@ bin/shards: $(SOURCES) $(TEMPLATES)
...
 	$(CRYSTAL) build src/shards.cr -o bin/shards $(CRFLAGS)
 
-install: bin/shards phony
+install: bin/shards manpages phony
 	$(INSTALL) -m 0755 -d "$(BINDIR)" "$(MANDIR)/man1" "$(MANDIR)/man5"
 	$(INSTALL) -m 0755 bin/shards "$(BINDIR)"
 	$(INSTALL) -m 0644 man/shards.1 "$(MANDIR)/man1"
 	$(INSTALL) -m 0644 man/shard.yml.5 "$(MANDIR)/man5"
 
+manpages:
```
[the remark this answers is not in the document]
straight-shoota: Sure, I can merge it into this one. I just figured when we're going to add more output formats, it might be better to keep this all separated.

wontruefree:
I agree with @bcardiff concern about adding the dependency.  This seems like a big dependency to add for preference in man page file format.

straight-shoota:
I've done a huge update on the documentation.

* Documentation is now written in asciidoc and resides in `docs/`. All documentation formats are generated from a single source.
* Man pages are automatically generated from asciidoc.
* `SPEC.md` has been merged with `man/shard.yml.5` into `docs/shard.yml.adoc`.
* relevant parts have been synced with [shards on crystal-book](https://github.com/crystal-lang/crystal-book/blob/master/the_shards_command/README.md)
* Added documentation for new commands not yet documented
* Added documentation for env vars

Integration with CI to build manpages automatically for shards releases is still missing. I'd like to ask for feedback first before fiddeling with that.

Next steps:
* doc files should also be made available as HTML on http://crystal-lang.org/reference

straight-shoota:
Bump can we please merge this before documentation gets out of sync again?

straight-shoota:
@bcardiff Could you take a look at this, please?

docs/shard.yml.adoc
```diff
@@ -0,0 +1,392 @@
+= shard.yml(5)
+:date: 2020-11-13
+:shards_version: 0.11.0
```
bcardiff, line 3: is this information right or should it get from the `SHARDS_VERSION` set in the Makefile?

bcardiff (commented):
Is really needed to move the SPEC.md to asciidocs? Maybe is because I haven't used asciidocs in the past that I am not eager to see that change.

I would suggest also to commit the generated man files. Otherwise distribution-scripts and other package mantainers will need to deal with asciidocs+python dependency. I usually prefer to avoid committing generated files, but in this case I think it make sense. 

Eventually the generated man files could go away, but to avoid blocking shards releases I would go that way.

docs/shard.yml.adoc
straight-shoota: Oh yeah, this can be automated. IIRC I didn't implement that right away because I wasn't sure it's a good idea because it could propagate incorrect information if the documentation is not accurate for the newest version. But that's probably hard to coordinate anyways, so it might be better to assume we keep the docs always up to date and it always displays the most recent version.

straight-shoota:
> Is really needed to move the SPEC.md to asciidocs?

`SPEC.md` is entirely merged into the manpage for `shard.yml`. I don't think it makes sense to keep two separate documents with duplicate content.

> I would suggest also to commit the generated man files. Otherwise distribution-scripts and other package mantainers will need to deal with asciidocs+python dependency. I usually prefer to avoid committing generated files, but in this case I think it make sense.
> 
> Eventually the generated man files could go away, but to avoid blocking shards releases I would go that way.

Sounds good :+1: I was also hesitant about this too. But yeah maybe it's better to commit them (at least for now).

straight-shoota:
Building man pages in CI is not an issue. We could use that for validation to ensure the rendered pages are up to date. But I'd leave that for a follow-up. Priority is to get this merged to avoid falling out of sync again.

bcardiff (approved):
I guess we should add a shards release checklist procedure to the "Crystal Release Checklist" doc in order to avoid forgetting to update the manpage upon release.

PR #265, opened 2019-04-02
Add SPEC for repository, homepage, documentation properties

### LINKED ISSUE #251: [SPEC] Add specification for metadata fields (website URL etc.)
When looking up information about a shard, it's important to know some key locations like where you can get more information and where it is hosted.
This is currently not provided in `shard.yml`.

I propose to introduce fields for informational URLs to relevant locations such as the website, repository, API docs, issue tracker etc.
Users could simply ad-hoc custom fields for this, but it would be beneficial to have a standardized format for extracting such information automatically. The `shards` command itself would not need to process them (except for display at some point).

Similar examples: Rubygems has `homepage` and [`metadata`](https://guides.rubygems.org/specification-reference/#metadata) fields. The latter accepts a mapping of custom descriptions to URLs. [Cargo](https://doc.rust-lang.org/cargo/reference/manifest.html#package-metadata) has `documentation`, `homepage` and `repository` fields. [NPM](https://docs.npmjs.com/files/package.json) has `homepage` and `repository` (holds a mapping similar to shards' dependencies).

I think the key references are to a website about the shard (usually called `homepage`) and the repository (`repository`).
These should probably simply be added as top level fields in the shards.yml specification.
I'd also consider links to a web presentation of the documentation and an issue tracker as important.  But perhaps these might added in a `metadata` field similar to Rubygems' where authors could also add custom mappings.

straight-shoota:
Fixes #251 

This adds properties `repository`, `homepage` and `documentation` to `SPEC.yml`.

I don't think there is any need to implement support in the shards executable for now.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -124,6 +124,25 @@ work with the project (e.g., an automated documentation generator).
 crystal: 0.19.2
 ```
 
+### repository
+
+The URL of the shard's canonical repository (String, recommended).
+
+The URL should be compatible with typical VCS tools without modifications.
+`http`/`https` is preferred over VCS schemes like `git`.
+It is recommended that this URL is publicly available.
+
+Copies of a shard (such as mirros, development forks etc.) should point to the same:
+canonical repository address, even if hosted at different locations.
+
+### homepage
+
+The URL of the shard's homepage (String, recommended).
+
+### documentation
+
+The URL to a website providing the shard's documentation for online browsing (String).
+
 ### dependencies
 
 A list of required dependencies (Hash).
--- a/SPEC.md
+++ b/SPEC.md
@@ -137,7 +137,7 @@ canonical repository address, even if hosted at different locations.
 
 ### homepage
 
-The URL of the shard's homepage (String, recommended).
+The URL of the shard's homepage (String).
 
 ### documentation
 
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -124,6 +124,25 @@ work with the project (e.g., an automated documentation generator).
...
 
+### repository
+
+The URL of the shard's canonical repository (String, recommended).
+
+The URL should be compatible with typical VCS tools without modifications.
+`http`/`https` is preferred over VCS schemes like `git`.
+It is recommended that this URL is publicly available.
+
+Copies of a shard (such as mirros, development forks etc.) should point to the same:
```
ysbaddaden: Typo: `mirros` and I feel like the colon after `same:` is artificial.

SPEC.md
```diff
@@ -124,6 +124,25 @@ work with the project (e.g., an automated documentation generator).
...
+canonical repository address, even if hosted at different locations.
+
+### homepage
+
+The URL of the shard's homepage (String, recommended).
```
ysbaddaden: I'm not sure the homepage has to be recommended.

straight-shoota: Sorry, lost track of this PR.

Yeah I guess it doesn't need to be, but I'd still recommend it 😄  And IMO it makes sense: when your shard has a homepage, you would want to make it publicly known anyway.

bcardiff: In many cases, the homepage will be obvious from the repository. But I don't think it bothers. Also, some might use a nicer homepage that the repository one. So 👍 on my side.

RX14:
I'd like to this PR to be merged. In the interest of not bikeshedding, I suggest removing "recommended".

PR #271, opened 2019-05-07
Addes Purge Command

wontruefree:
This command clears local shards cache.

I have had to clear the shards cache to fix some issues with chards and thought it would be a useful command.

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -16,6 +16,7 @@ module Shards
           outdated [--pre]               - List dependencies that are outdated.
           prune                          - Remove unused dependencies from `lib` folder.
           update [<shards>]              - Update dependencies and `shard.lock`.
+          purge                          - Clears the local cache
           version [<path>]               - Print the current version of the shard.
 
       Options:
@@ -64,6 +65,8 @@ module Shards
             path,
             args[1..-1].reject(&.starts_with?("--"))
           )
+        when "purge"
+          Commands::Purge.run(path)
         when "version"
           Commands::Version.run(args[1]? || path)
         else
--- a/src/commands/purge.cr
+++ b/src/commands/purge.cr
@@ -1,0 +1,18 @@
+require "./command"
+require "../config"
+
+module Shards
+  module Commands
+    class Purge < Command
+      def run
+        FileUtils.rm_r(Dir.children(Shards.cache_path))
+
+        if Dir.empty?(Shards.cache_path)
+          Shards.logger.info "Cache has been cleaned"
+        else
+          Shards.logger.info "Unable to verify cache has been cleaned"
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


straight-shoota:
I guess this might be a useful addition, but it should have a more specific name. `purge` could mean anything and I'd assume it clears the `lib` directory or something similar (like `prune`).
So maybe `purge-cache`.

OTOH `rm -r ~/.shards/cache` is obviously a pretty simple command as well.

But you'd need to know exactly which cache path `shards` actually uses and this depends on several environment variables including `SHARDS_CACHE_PATH`. So letting `shards` handle that directly is probably better.

It would be nice to have a way to expose the cache path, like `crystal env`.

wontruefree:
I had trouble with the name and wanted it to follow a single name similar to other commands.  But I would be open to suggestions.

As for just running `rm -r ~/.shards/cache`.  I have done that before, and it works fine but I had to find the path through a GitHub issue, and I think this could be way more discoverable.  I consider the location of the cache directory an implementation detail and something that should be able to be handled by a high-level command.

Here all the locations the cache could be.
https://github.com/crystal-lang/shards/blob/master/src/config.cr#L17

straight-shoota:
> I consider the location of the cache directory an implementation detail

It is configurable using `SHARDS_CACHE_PATH`, so it's not just internal. It's not documented though. But it probably should be.

ysbaddaden:
AFAIK the environment variable is only meant for testing purposes.

Could you detail the issues you met? The cache is a mere `git clone --mirror` and there shouldn't be issues/conflicts, so I'm curious.

wontruefree:
I believe this fixed a problem I had when adding a shard conflicted with the dependencies of another shard.  To fix that I forked the repo but my local `shard install` was failing.  I believe clearing the cache fixed it.

It has been a little while so I don't remember all the details.

jwoertink:
Not sure if this command would take care of an issue I have or not, but I actually have an issue with one of my lucky apps. When I run `lucky dev`, I can guarantee at least 98% of the time I get this error:

```text
web    | while requiring "./start_server" (Exception)
web    |   from Crystal::ASTNode+@Crystal::ASTNode#accept<Crystal::TopLevelVisitor>:Nil
...
web    | Caused by: Unable to create directory '/Users/jeremywoertink/.cache/crystal/Users-jeremywoertink-Sites-app-lib-teeplate-src-lib-file_tree-macros-directory.cr': File exists (Errno)
web    |   from Dir::mkdir_p<String>:Int32
...
web    | Error: you've found a bug in the Crystal compiler. Please open an issue, including source code that will allow us to reproduce the bug: https://github.com/crystal-lang/crystal/issues
web    |
web    | Exited
```

Clearing that crystal cache makes it work again. Being able to just append a `shards cache --purge` or something would be nice.

ysbaddaden:
@jwoertink that's crystal cache (compile object files and macro executables), not shards' cache of git repositories;

@wontruefree repositories are cloned I such a way that they can't clash (cache/server/repo), so I'm a little bewildered.

wontruefree:
@ysbaddaden my problems might have been fixed from clearing crystals cache, not shards cache.  I am going to close this.

PR #275, opened 2019-05-14
Fix: user/repo is case insensitive in github repositories

ysbaddaden:
We linearize the dependency definition to always be lowercase to avoid issues later (e.g. cloning a repository multiple times in the cache).

### DIFF
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -7,7 +7,15 @@ module Shards
     def self.new(pull : YAML::PullParser) : self
       Dependency.new(pull.read_scalar).tap do |dependency|
         pull.each_in_mapping do
-          dependency[pull.read_scalar] = pull.read_scalar
+          key = pull.read_scalar
+          value = pull.read_scalar
+
+          # HACK: user/repo is case insensitive in github repositories, we thus
+          # linearize the dependency definition to always be lowercase to avoid
+          # issues later (e.g. cloning a repository multiple times in the cache)
+          value = value.downcase if key == "github"
+
+          dependency[key] = value
         end
       end
     end


src/dependency.cr
```diff
@@ -7,7 +7,15 @@ module Shards
...
       Dependency.new(pull.read_scalar).tap do |dependency|
         pull.each_in_mapping do
-          dependency[pull.read_scalar] = pull.read_scalar
+          key = pull.read_scalar
+          value = pull.read_scalar
+
+          # HACK: user/repo is case insensitive in github repositories, we thus
+          # linearize the dependency definition to always be lowercase to avoid
+          # issues later (e.g. cloning a repository multiple times in the cache)
+          value = value.downcase if key == "github"
```
Sija: Isn't GitLab doing the same?

ysbaddaden: No idea.

Sija: From my testing it seems it does.

straight-shoota: Yes, and bitbucket as well. I don't think any of them has case-sensitive names.
In fact, the case-sensitive resolvers are `path` (depending on the file system/os) and generic `git` URLs (partial, for example host is case-insensitive). The latter can be normalized with the improved `URL#normalize` in Crystal 0.29.0.

ysbaddaden: Git can use whatever URL scheme, for example `file:` and I don't think the path of `file:///path/to/repo` is case insensitive.

ysbaddaden:
This is just a quick hack. If someone can come up with something correct, that would be appreciated.

Are all Git providers case insensitive? Are all Git server software case insensitive? ...

src/dependency.cr
straight-shoota: Yes, but in any URI scheme, the generic host component and the scheme are case-insensitive. They can and should be normalized.

straight-shoota:
Hm, Github's own help page states repository names are case-sensitive: https://help.github.com/en/articles/error-repository-not-found#check-your-spelling

This is somehow contradicted by http://github.com/CRYSTAL-LANG/SHARDS however

I guess it might just be a difference between theory and practice and we're probably fine optimizing for the practical application. A safer alternative would be querying the Github, Gitlab, Bitbucket API for the canonical repository name.

bcardiff:
If the issue is about the collision of the cache, then maybe the solution is to make the cache use other convention. One that will work on case insensitive file system. The [Crystal::Compiler::CompilationUnit solves that](https://github.com/bcardiff/crystal/blob/affe8ec59c07514ea26c99bb26b7366009b7d5d6/src/compiler/crystal/compiler.cr#L599-L612).

straight-shoota:
@bcardiff I guess we haven't even considered case-insensitivity of the local filesystem yet. The issue has been about identifying case-insensitive references and thus using the same cache for different case variants of the same repository.

bcardiff:
Right, is not that two copies of the same dependency would make sense. And the user is not in control on which case a nested dependency is declared.

I'm ok with the current state of the PR for now. The canonical resolvers sound like a good idea.

straight-shoota:
If we normalize github, we should also normalize gitlab and bitbucket. They both seem to resolve repositories case-insensitive.

beta-ziliani:
IIUC this is already addressed by #471, closing it.

PR #278, opened 2019-05-15
Fix: prune command must only remove directories

ysbaddaden:
The issue was found in the nightly builds.


straight-shoota:
Should it also remove hidden directories?

bcardiff:
@straight-shoota I don't think so. it there is a hidden directory it should not be a shard name, so it's safe to never remove it.

ysbaddaden:
The command is still broken: it should prune `foo.sha1` files too, (created to know which commit of a shard _is_ installed) but probably doesn't —it's missing some additional tests.

PR #296, opened 2019-09-16
Use read_file macro instead using OS-dependent `cat`

Sija:
SSIA

### DIFF
--- a/src/version.cr
+++ b/src/version.cr
@@ -1,5 +1,5 @@
 module Shards
-  VERSION    = {{ `cat #{__DIR__}/../VERSION`.stringify.chomp }}
+  VERSION    = {{ read_file("#{__DIR__}/../VERSION").chomp }}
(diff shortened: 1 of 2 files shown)


ysbaddaden:
Looks good. Yet, CI is red. Maybe master got broken, but I'd like CI to be green before merging PR.

Sija:
Sure, that's understandable, yet AFAICS it's an unrelated fail - it fails because due to changes related to displaying error messages.

ysbaddaden:
Would you be so kind to provie a PR to fix it? That would highly appreciated!

Sija:
@ysbaddaden Sure, np. Btw, seems like `--no-color` switch ain't working.

ysbaddaden:
No colors only affects shards (logs have no colors) not the crystal build command (errors now always have colors, despite not being a tty).

ysbaddaden:
The cat call breaks at compile time if the VERSION file is missing. What happens with read_file macro?

straight-shoota:
`read_file` fails as well.

ysbaddaden:
Thank you!

PR #298, opened 2019-10-08
Add run command

### LINKED ISSUE #157: Run command

denolfe:
Adds `shards run` command. Resolves #157.


src/cli.cr
```diff
@@ -76,6 +79,31 @@ module Shards
...
+
+    begin
+      Commands::Check.run(path)
+    rescue
+      Commands::Install.run(path)
+    end
+
+    Commands::Run.run(path, targets, options)
+  end
+end
```
ysbaddaden: Maybe have a `check_and_install_dependencies` too, remove the `build` and `run_command` methods and inline the 3 calls in  the case above:

```crystal
targets, options = parse_args(args)
chech_and_install_dependencies
Command::Run.run(path, targets, options)
```

src/cli.cr
```diff
@@ -76,6 +79,31 @@ module Shards
...
+
+  def self.parse_args(args)
```
ysbaddaden: It should stop parsing args when `--` is found, and return them as `run_options`, in order to pass args to the run executable, such as:

```console
$ shards run test --static -- --seed=abc name
# crystal build src/test.cr -o bin/test --static
# bin/test --seed=abc name
```

denolfe: This is now handled by passing an additional parameter into the run command for `run_options`.

ysbaddaden:
Thanks. Can you write a few integration tests, to check that the targets are built and run, and the exit status properly reported?

denolfe:
For sure. Is there already a way to evaluate the exit_code somewhere in the tests? This is my first go with minitest, so it's a little difficult to debug my tests. I see that `run` from `factories.cr` is used in the `build_test.cr` tests.

denolfe:
Tests added, but 2 of them are behaving unexpectedly. I've verified the functionality in a dummy repo and using the `bin/shards` that is compiled. I'm not too sure why these are behaving differently within the tests.

The 2 tests in question are expected to throw a `FailedCommand` but are not.

- [test_fails_when_multiple_targets_no_arg](https://github.com/crystal-lang/shards/pull/298/commits/540c94109eeb49778ef9f0aea6520d5bdb811de2#diff-2b6f4b1521e592ae9a67436c2970d72eR53)
- [test_fails_when_no_targets_defined](https://github.com/crystal-lang/shards/pull/298/commits/540c94109eeb49778ef9f0aea6520d5bdb811de2#diff-2b6f4b1521e592ae9a67436c2970d72eR70)

Any help is welcome.

ysbaddaden:
The YAML blocks create a YAML with 2 leading spaces, it should be the following (notice that the blocking YAML is indented):

```crystal
str = <<-YAML
  name: xxx
  version: 0.0.0
  YAML
```

denolfe:
Hmm, there appears to be an issue beyond just that. Modifying as suggested did not make a difference.

ysbaddaden:
Yes, youre not changing the current directory in some tests.

src/commands/run.cr
```diff
@@ -0,0 +1,24 @@
+require "./command"
+
+module Shards
+  module Commands
+    class Run < Command
+      def run(targets, options, run_options)
+        targets = spec.targets.map(&.name) if targets.empty?
+        raise Error.new("Error No targets defined.") if targets.empty?
+        raise Error.new("Error More than 1 target defined. Must pass target name as parameter.") if targets.size > 1
```
ysbaddaden: We can improve the error mesages, here. For example we can know whether the user asked for a target or not. In the former case we could say something like "please specify one target only; if you meant to pass arguments you may use: shards run target -- args". In the latter case we could say "please specify which target you want to run" with maybe a list.

denolfe: Implemented.

src/commands/run.cr
```diff
@@ -4,19 +4,30 @@ module Shards
...
           Commands::Build.run(path, targets, options)
           Shards.logger.info { "Executing: #{target.name} #{run_options.join(' ')}" }
 
           error = IO::Memory.new
           status = Process.run(File.join(Shards.bin_path, target.name), args: run_options, output: Process::Redirect::Inherit, error: error)
-          raise Error.new("Error target #{target.name} failed to run:\n#{error}") unless status.success?
-        elsif
-          raise Error.new("Error target #{targets.first} not found.")
+          raise Error.new("Error Target #{target.name} failed to run:\n#{error}") unless status.success?
+        elsif raise Error.new("Error Target #{targets.first} not found.")
```
ysbaddaden: Typo: `elsif` ?

src/commands/run.cr
```diff
@@ -0,0 +1,35 @@
...
+          targets = spec.targets.map(&.name)
+        end
+
+        raise Error.new("Error No targets defined.") if targets.empty?
+
+        if targets.size > 1
+          if specific_target_requested
+            raise Error.new("Error Please specify only one target. If you meant to pass arguments you may use: shards run target -- args.")
+          elsif
+            raise Error.new("Error More than 1 target defined. Please specify what target you want to run.")
```
ysbaddaden: > Error please specify the target with 'shards run target'

src/commands/run.cr
```diff
@@ -0,0 +1,35 @@
...
+        if targets.empty?
+          specific_target_requested = false
+          targets = spec.targets.map(&.name)
+        end
+
+        raise Error.new("Error No targets defined.") if targets.empty?
+
+        if targets.size > 1
+          if specific_target_requested
+            raise Error.new("Error Please specify only one target. If you meant to pass arguments you may use: shards run target -- args.")
```
ysbaddaden: > Error Please specify only one target. If you meant to pass arguments you may use 'shards run target -- args'

src/commands/run.cr
```diff
@@ -0,0 +1,35 @@
...
+  module Commands
+    class Run < Command
+      def run(targets, options, run_options)
+        specific_target_requested = true
+        if targets.empty?
+          specific_target_requested = false
+          targets = spec.targets.map(&.name)
+        end
+
+        raise Error.new("Error No targets defined.") if targets.empty?
```
ysbaddaden: > Error no targets defined

src/commands/run.cr
ysbaddaden: Let's reuse the Build command message when the target doesn't exist:

> Error target #{target.name} was not found in #{SPEC_FILENAME}

src/cli.cr
```diff
@@ -86,14 +93,15 @@ module Shards
         targets << arg
       end
     end
+    { targets, options }
+  end
 
+  def self.check_and_install_dependencies(path)
     begin
```
Sija: This `begin` / `end` became redundant now.

src/commands/run.cr
Sija: `elsif` -> `else`, also please use indentation after `else`.

src/cli.cr
denolfe: Could you elaborate on this? It appears to me that the check command can still raise an error that should be resolved by doing the install command.

Sija: yes, I mean that you can use `rescue` without `begin ... end`:

```cr
def self.check_and_install_dependencies(path)
  # ...
rescue # ...
  # ...
end
```

denolfe: Great, thanks. I didn't know that was possible to do 👍

denolfe:
> Also, I think the `run` command should just pass STDIN, STDOUT and STDERR as-is. Maybe even replacing the current process with `exec` — but I'm not sure it's possible with Crystal.

@ysbaddaden Was this something that still needs to be addressed or just a nice to have?

src/commands/run.cr
```diff
@@ -0,0 +1,33 @@
...
+            raise Error.new("Error please specify only one target. If you meant to pass arguments you may use 'shards run target -- args'")
+          else
+            raise Error.new("Error please specify the target with 'shards run target'")
+          end
+        end
+
+        if target = spec.targets.find { |t| t.name == targets.first }
+          Commands::Build.run(path, targets, options)
+          Shards.logger.info { "Executing: #{target.name} #{run_options.join(' ')}" }
+          status = Process.exec(command: File.join(Shards.bin_path, target.name), args: run_options, output: Process::Redirect::Inherit)
```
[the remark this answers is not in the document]
denolfe: Oh wow, I thought I'd removed that. Will fix.

src/commands/run.cr
```diff
@@ -0,0 +1,33 @@
...
+            raise Error.new("Error please specify only one target. If you meant to pass arguments you may use 'shards run target -- args'")
+          else
+            raise Error.new("Error please specify the target with 'shards run target'")
+          end
+        end
+
+        if target = spec.targets.find { |t| t.name == targets.first }
+          Commands::Build.run(path, targets, options)
+          Shards.logger.info { "Executing: #{target.name} #{run_options.join(' ')}" }
+          Process.exec(command: File.join(Shards.bin_path, target.name), args: run_options, output: Process::Redirect::Inherit)
```
RX14: `output: :inherit` is actually the default already.

denolfe:
@ysbaddaden Anything else needed on this one before getting approved?

src/commands/run.cr
denolfe: Output parameter removed.

src/cli.cr
```diff
@@ -15,6 +15,7 @@ module Shards
           lock [--update] [<shards>]     - Lock dependencies in `shard.lock` but doesn't install them.
           outdated [--pre]               - List dependencies that are outdated.
           prune                          - Remove unused dependencies from `lib` folder.
+          run [<target>]                 - Build and run specific target
```
Sija:
```suggestion
          run [<target>] [<options>]     - Build and run specific target
```

Sija:
@denolfe Specs needs to be rewritten (see #334).

bcardiff:
Can we have a spec checking how to pass ARGV values from the shards run command to the actual command? I guess it should be `shards run foo -- bar`, but I am not sure from the code if that's the idea.

denolfe:
I've confirmed manually that passing ARGV values work, but simulating it using the run function from factories.cr has proven to be difficult. It spits errors out to the console when the test is run saying it cannot read ARGV[0] which is being passed. Here is the WIP test code:

```crystal
  it "runs specific target with args" do
    File.write(File.join(application_path, "src", "cli_with_args.cr"), "puts __FILE__;args = ARGV[0]")

    File.write File.join(application_path, "shard.yml"), <<-YAML
      name: build
      version: 0.1.0
      targets:
        app:
          main: src/cli_with_args.cr
      YAML

    Dir.cd(application_path) do
      run "shards run app -- foo"
      File.exists?(bin_path("app")).should be_true
      `#{bin_path("app")}`.chomp.should eq(File.join(application_path, "src", "cli_with_args.cr"))
    end
  end
```

Could this be something specific with how ARGV is handled in a shell environment vs. through Process.exec? Any help appreciated.

silasb:
Anything I can do to help move this feature over the finish line?

denolfe:
> Can we have a spec checking how to pass ARGV values from the shards run command to the actual command? I guess it should be `shards run foo -- bar`, but I am not sure from the code if that's the idea.

@silasb The only piece missing was a test for the above. I'm able to verify functionality manually, but testing it properly proved to be difficult because of how the commands are fed to the shell.

denolfe:
I no longer use Crystal personally and don't have time to get these tests running, so I'll be closing this PR out. My branch will still be accessible if someone wanted to use it to finish #157.

PR #306, opened 2019-11-19
YAML parser for Dependency and Target

straight-shoota:
Implements proper YAML parses for Dependency and Target to avoid relying on a hash for storing the mappings. Instead, these types use specific properties now.

Extra attributes are still collected in a hash. This has no real function currently, but it allows for future enhancements with new keys not causing the current implementation to fail reading the data.

/cc #230


test/git_resolver_test.cr
```diff
@@ -44,9 +44,9 @@ module Shards
       skip "TODO: install tag (whatever the version)"
     end
 
-    private def resolver(name, config = {} of String => String)
-      config = config.merge({"git" => git_url(name)})
-      dependency = Dependency.new(name, config)
+    private def resolver(name, **config)
```
ysbaddaden: No need to bother: the tests never pass `config`:

```crystal
private def resolver(name)
  dependency = Dependency.new(name)
  dependency.git = git_url(name)
  GitResolver.new(dependency)
end
```

test/path_resolver_test.cr
```diff
@@ -74,9 +74,9 @@ module Shards
...
     end
 
-    private def resolver(name, config = {} of String => String)
-      config["path"] = git_path(name)
-      dependency = Dependency.new(name, config)
+    private def resolver(name, **config)
+      config = config.merge(resolver: "path", url: git_path(name))
+      dependency = Dependency.new(name, **config)
       PathResolver.new(dependency)
     end
```
ysbaddaden: Same here no need for `config`. Passing anything but `path` doesn't even make sense for `PathResolver` anyway.

test/resolver_test.cr
```diff
@@ -3,10 +3,9 @@ require "./test_helper"
 module Shards
   class ResolverTest < Minitest::Test
     def test_find_resolver_with_unordered_dependency_keys
-      dependency = Dependency.new("test", {
-        "branch" => "master",
-        "git"    => "file:///tmp/test",
-      })
+      dependency = Dependency.new("test", branch: "master",
+        resolver: "git", url: "file:///tmp/test",
+      )
```
ysbaddaden: The test doesn't make sense anymore, unless it's changed to parse a YAML file with unordered keys, maybe.

src/target.cr
```diff
@@ -1,21 +1,40 @@
...
+    end
+
+    def initialize(@name : String, @main : String, @attributes = {} of String => YAML::Any)
     end
 
-    def initialize(@name)
-      super()
+    def source_path : String
+      main
     end
```
ysbaddaden: Nice but that's a mere alias. Let's keep `target.main` which is the name from SPEC.

src/target.cr
```diff
@@ -1,21 +1,40 @@
...
 
-    def main
-      self["main"]
+    def output_path : String
+      File.join(Shards.bin_path, name)
     end
```
ysbaddaden: Again, this is nice but this is only ever used once, and introduces a dependency on a configuration variable (`Shards.bin_path`) that's external to SPEC.

test/dependency_test.cr
```diff
@@ -5,25 +5,25 @@ class Shards::DependencyTest < Minitest::Test
     dependency = Dependency.new("app")
     assert_equal "*", dependency.version
 
-    dependency = Dependency.new("app", {version: "*"})
+    dependency = Dependency.new("app", version: "*")
```
ysbaddaden: These unit tests are the only use case for an initializer with additional values. They don't make much sense anymore... unless we rewrite them a little:

```crystal
def test_version
  dependency = Dependency.new("app")
  assert_nil dependency.version?
  assert_equal "*", dependency.version

  dependency.version = "<= 2.0.0"
  assert_equal "<= 2.0.0", dependency.version?
  assert_equal "<= 2.0.0", dependency.version
end

def test_version_tags
  dependency = Dependency.new("app")

  dependency.tag = "fix/something"
  assert_equal "*", dependency.version

  dependency.tag = "v1.2.3-pre1"
  assert_equal "v1.2.3-pre1", dependency.version
end
```

src/dependency.cr
```diff
@@ -1,60 +1,83 @@
...
+  class Dependency
     property name : String
+    setter version : String?
+    property! resolver : String?
+    property! url : String?
+    property tag : String?
+    property branch : String?
+    property commit : String?
+
+    getter unmapped = Hash(String, YAML::Any).new
```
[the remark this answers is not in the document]
straight-shoota: Yeah, skipping might be better. I thought about future extensibility, for example for a `validate` command that could identify misspelled key names, or other extensions that support non-specified mappings. But that might just be premature.

src/dependency.cr
```diff
@@ -1,60 +1,83 @@
...
+      @resolver : String? = nil, @url : String? = nil,
+      @version : String? = nil,
+      @branch : String? = nil, @tag : String? = nil, @commit : String? = nil,
+      @unmapped : Hash(String, YAML::Any) = Hash(String, YAML::Any).new
+    )
     end
 
     def version
-      version { "*" }
+      version? || "*"
```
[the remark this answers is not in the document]
straight-shoota: What's the benefit of a yielding method over the nilable? Usage is `value { in_case_of_nil }` vs `value? || in_case_of_nil`. The latter is more descriptive.

src/dependency.cr
```diff
@@ -1,60 +1,83 @@
...
+          dependency.version = pull.read_scalar
+        when "url"
+          dependency.url = pull.read_scalar
+        when "tag"
+          dependency.tag = pull.read_scalar
+        when "branch"
+          dependency.branch = pull.read_scalar
+        when "commit"
+          dependency.commit = pull.read_scalar
+        when "path", "git", "github", "gitlab", "bitbucket"
```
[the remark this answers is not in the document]
straight-shoota: Just mapping the names would be problematic when multiple mappings are specified in a single entry. Per current spec that's valid. When you only have the names you later don't know about the order of definition for picking the last one as the final one (which is IMO a sane option to treat this, and the current implementation in this PR).

Besides that, the data model is much easier and more efficient when you can simply retrieve the name of the resolver and its url (or "resolution value"), than having to iterate through the possible resolvers in order to fine which one's specified. I'd be happy to change `resolver` to `resolver_name`. Both `owner/repo` and local paths are URLs, so technically that's not incorrect. I'm feeling quite strongly that this is a superior solution over multiple properties.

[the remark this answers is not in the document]
ysbaddaden: - Its forbidden to have multiple resolvers for a single dependency declaration. SPEC can evolve to make that explicit. It makes no sense to have multiple resolvers.
- There is no order in dependency keys, either.
- `owner/repo` isn't an URL (barely a path) and path isn't an URL either but an absolute path (it would be `file://`).

src/dependency.cr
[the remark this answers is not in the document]
ysbaddaden: It avoids the nilable union type. This is common in stdlib.

src/dependency.cr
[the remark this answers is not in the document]
ysbaddaden: We'll see then. Maybe a validator command would use its own parser, for example a blunt YAML.parse, then walk and validate the tree, collecting issues as it goes.

src/dependency.cr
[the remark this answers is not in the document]
straight-shoota: I'll change it to error if multiple resolvers are mapped.

Technically, relative paths are a URI reference. But naming aside, do you agree on normalized properties for resolver name and reference?

src/dependency.cr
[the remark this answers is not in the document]
straight-shoota: A nilable union would still exist unless [`src/command/check.cr:45`](https://github.com/crystal-lang/shards/pull/306/files#diff-336e9a9c9a128eb8d43b15550425b36cR45) is refactored.

test/resolver_test.cr
straight-shoota: It still tests the behaviour of `Shards.find_resolver`.

src/commands/check.cr
```diff
@@ -42,7 +42,7 @@ module Shards
           return false
         end
 
-        if version = lock["version"]?
+        if version = lock.version { nil }
```
Sija:
```suggestion
        if version = lock.version?
```

src/dependency.cr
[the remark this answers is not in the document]
straight-shoota: @ysbaddaden ?

src/dependency.cr
```diff
@@ -29,32 +58,28 @@ module Shards
...
     def refs
-      self["branch"]? || self["tag"]? || self["commit"]?
-    end
-
-    def path
-      self["path"]?
+      branch || tag || commit
     end
 
     def to_human_requirement
```
ysbaddaden: This is very noisy. I don't think we need the local variables anymore, so just:

```crystal
if version?
  version
elsif branch
  "branch #{branch}"
elsif tag
  "tag #{tag}"
elsif commit
  "commit #{commit}"
else
  "*"
end
```

straight-shoota:
I suppose having only `git` and `path` properties and immediately expand specialized resolvers is fine.

The specialized classes could be removed regardless of this change, though.

src/resolvers/git.cr
```diff
@@ -15,6 +15,17 @@ module Shards
       "git"
     end
 
+    def self.expand_resolver_url(url, resolver)
+      case resolver
+      when "git"
+        url
+      when "github", "bitbucket", "gitlab"
+        "https://github.com/#{url}.git"
```
Sija: You sure about that? ;)

straight-shoota: Totally

straight-shoota: Seems there are no specs covering bitbucket and gitlab mappings...

straight-shoota:
Rebased on master

straight-shoota:
Rebased again.

Issue #308, opened 2019-12-30
Doesn't support ~ in path 


masukomi:
setup: 
Created a new shard. Wanted to test it locally before pushing

used this in my dependencies:

```
  bayes_classifier:
    path: ~/workspace/bayes_classifier
```

shards kept complaining about 

```
Missing "shard.yml" for "bayes_classifier"
```

but I was confident it was there (and readable). I tweaked `path.cr` to indicate _where_ shards was looking when it failed:

```
$ ~/workspace/shards/shards install
Resolving dependencies
Missing "shard.yml" for "bayes_classifier"
  Expected it at ~/workspace/bayes_classifier/shard.yml
$ ls ~/workspace/bayes_classifier/shard.yml
/Users/masukomi/workspace/bayes_classifier/shard.yml
```

Modifying the value of the `path:` to be `/Users/masukomi/workspace/bayes_classifier` solved the problem. 

using `~` to represent the current user's home directory is an incredibly common and  decades old practice. This should be supported or _at the very least_ the error message should indicate that `~` isn't supported in paths.

j8r:
The [Path](https://crystal-lang.org/api/master/Path.html) type and [File.expand_path](https://crystal-lang.org/api/master/File.html#expand_path(path,dir=nil,*,home=false):String-class-method) supports this special character.
This means shards doesn't use path expansion.

ysbaddaden:
Only absolute paths are supported for `path` value. SPEC should probably be updated to reflect that.

Considering `~` as absolute is debatable.

straight-shoota:
The error message should also be improved. It should not complain about missing `shard.yml` when the entire parent path does not exist.

straight-shoota:
@ysbaddaden Why only absolute paths? I've used relative paths quite successfully. And they are very useful when working on a project with multiple related repositories in a project folder. References between them are easy and you don't need to care about where the project folder lives.

ysbaddaden:
No opinion from me.

It may work with relative paths, but there are no guarantees that it will work properly.

I think we should properly design how the path resolver should work. I never did, mostly because of lack of personal use cases.

straight-shoota:
It actually seems to work well. I don't see a reason to change anything about how path resolver works. I haven't used it too much, but when I did it went smoothly. Since there are no complains about it, either others have the same positive experience or it's not used much at all. Both ways there's nothing to be done (at the moment).

Except for adding simple features such as home path expansion. That seems very reasonable to me.

[...]

PR #309, opened 2019-12-30
added support for tilde-initial paths

masukomi:
I.e. specifying the home directory by starting the
path with a tilde.

Additionally added extra info to the error message
when a shard.yml file isn't found.

This fixes Issue #308 

All tests pass. No new tests added to cover this because there don't appear to be ANY tests for this class.

### DIFF
--- a/src/resolvers/path.cr
+++ b/src/resolvers/path.cr
@@ -7,12 +7,13 @@ module Shards
     end
 
     def read_spec(version = nil)
-      spec_path = File.join(local_path, SPEC_FILENAME)
+      spec_path = File.join(Path[local_path].expand(home: true).to_s, SPEC_FILENAME)
+
 
       if File.exists?(spec_path)
         File.read(spec_path)
       else
-        raise Error.new("Missing #{SPEC_FILENAME.inspect} for #{dependency.name.inspect}")
+        raise Error.new("Missing #{SPEC_FILENAME.inspect} for #{dependency.name.inspect}\n  Expected it at #{spec_path}")
       end
     end
 
@@ -21,7 +22,7 @@ module Shards
     end
 
     def spec?(version)
-      spec_path = File.join(local_path, SPEC_FILENAME)
+      spec_path = File.join(Path[local_path].expand(home: true).to_s, SPEC_FILENAME)
 
       if File.exists?(spec_path)
         Spec.from_yaml(File.read(spec_path))
@@ -58,7 +59,7 @@ module Shards
     end
 
     def local_path
-      dependency["path"].to_s
+      path = dependency["path"].to_s
     end
 
     private def expanded_local_path


straight-shoota:
Please use `Path#expand`/`File#expand_path` to implement this.

ysbaddaden:
Please not that #308 may not be considered a bug.

masukomi:
> Please use Path#expand/File#expand_path to implement this. - @straight-shoota 

That was my first attempted solution. Technically i just replaced the calls to `local_path` with `expand_local_path` (better to use existing functionality) because it just does `File.expand_path`)

BUT doing that converted `~/workspace/bayes_classifier` to `/Users/masukomi/workspace/scuttlebutt/weekly_ssb_summarization/~/workspace/bayes_classifier`

there are 2 problems with this:

`/Users/masukomi/workspace/scuttlebutt/weekly_ssb_summarization/` is the directory i was in when i Ran it and has NOTHING to do with where it should be finding the path. 

The `~` is still there.


using `Path[local_path].expand(home: true).to_s` seems to work.... (committed, rebased, pushed) BUT I'm seeing some other issues that i wasn't seeing before. I think it's just other stuff on my system that's unrelated. I would feel more confident if there were tests on this class.

masukomi:
additional issues resolved. corrected. rebased pushed. 

it just needed to be expanded in `spec?(version)` too

src/resolvers/path.cr
```diff
@@ -58,7 +59,7 @@ module Shards
     end
 
     def local_path
-      dependency["path"].to_s
+      path = dependency["path"].to_s
```
Sija: That shouldn't be changed.

src/resolvers/path.cr
```diff
@@ -7,12 +7,13 @@ module Shards
...
     def read_spec(version = nil)
-      spec_path = File.join(local_path, SPEC_FILENAME)
+      spec_path = File.join(Path[local_path].expand(home: true).to_s, SPEC_FILENAME)
+
 
       if File.exists?(spec_path)
         File.read(spec_path)
       else
-        raise Error.new("Missing #{SPEC_FILENAME.inspect} for #{dependency.name.inspect}")
+        raise Error.new("Missing #{SPEC_FILENAME.inspect} for #{dependency.name.inspect}\n  Expected it at #{spec_path}")
```
Sija: No need for newlines.

ysbaddaden: In fact it's not needed when using `expanded_local_path` that will report that the directory doesn't exist.

ysbaddaden:
Note that I have nothing against `Path`, I just want the bugfix to be as minimal as possible :)

straight-shoota:
This PR has been superseeded by #538 and #541

PR #314, opened 2020-01-15
Make shards reproducible

jelly:
By default the shards binary embeds a build date which makes it
impossible to rebuild due to time differences. Use SOURCE_DATE_EPOCH to
use a predictable timestamp as input for the date function which in turn
makes shards reproducible.

Motiviation: https://reproducible-builds.org

### DIFF
--- a/src/version.cr
+++ b/src/version.cr
@@ -1,7 +1,7 @@
 module Shards
   VERSION    = {{ read_file("#{__DIR__}/../VERSION").chomp }}
   BUILD_SHA1 = {{ `git log --format=%h -n 1 2>/dev/null || echo ""`.stringify.chomp }}
-  BUILD_DATE = {{ `date -u +'%Y-%m-%d'`.stringify.chomp }}
+  BUILD_DATE = Time.unix({{ (env("SOURCE_DATE_EPOCH") || `date +%s`).to_i }}).to_s("%Y-%m-%d")
 
   def self.version_string
     if BUILD_SHA1.empty?


src/version.cr
```diff
@@ -1,7 +1,7 @@
 module Shards
   VERSION    = {{ read_file("#{__DIR__}/../VERSION").chomp }}
   BUILD_SHA1 = {{ `git log --format=%h -n 1 2>/dev/null || echo ""`.stringify.chomp }}
-  BUILD_DATE = {{ `date -u +'%Y-%m-%d'`.stringify.chomp }}
+  BUILD_DATE = {{ `date -u --date="@${SOURCE_DATE_EPOCH:-$(date +%s)}" +'%Y-%m-%d'`.stringify.chomp }}
```
straight-shoota: We must not rely on `--date` which is not available on all platforms. This is a better implementation:

https://github.com/crystal-lang/crystal/blob/42a64909fae022863e68651a3d35f3c45884efc8/src/compiler/crystal/config.cr#L35-L36

straight-shoota: The updated code doesn't work because the result is not assigned to the constant. 

```cr
  BUILD_DATE = Time.unix({{ (env("SOURCE_DATE_EPOCH") || `date +%s`).to_i }})).to_s("%Y-%m-%d")
```

src/version.cr
```diff
@@ -1,7 +1,9 @@
 module Shards
   VERSION    = {{ read_file("#{__DIR__}/../VERSION").chomp }}
   BUILD_SHA1 = {{ `git log --format=%h -n 1 2>/dev/null || echo ""`.stringify.chomp }}
-  BUILD_DATE = {{ `date -u +'%Y-%m-%d'`.stringify.chomp }}
+  BUILD_DATE = {{ (env("SOURCE_DATE_EPOCH") || `date +%s`).to_i }}
+  Time.unix(BUILD_DATE).to_s("%Y-%m-%d")
```
Sija: Something went wrong it seems...

src/version.cr
jelly: Thanks updated the PR

PR #315, opened 2020-01-20
Update origin_changed? to ignore protocol differences

### LINKED ISSUE #288: shards keeps downloading dependencies
Running ```shards install``` takes a very long time on a project with a number of dependencies, as ```shards``` keep refetching the same sources.

Here's an example from a ```lucky``` project (just an example, but could be anything really):

```
$ shards install
  Resolving dependencies                                                                                                                                          [19/9143]
  Fetching https://github.com/luckyframework/lucky.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/mosop/teeplate.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/avram.git
  Fetching https://github.com/kostya/blank.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/mosop/teeplate.git
  Fetching https://github.com/mosop/teeplate.git
  Fetching https://github.com/mosop/teeplate.git
  Fetching https://github.com/mosop/teeplate.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/luckyframework/lucky_router.git
  Fetching https://github.com/luckyframework/shell-table.cr.git
  Fetching https://github.com/paulcsmith/cry.git
  Fetching https://github.com/mosop/cli.git
  Fetching https://github.com/mosop/optarg.git
  Fetching https://github.com/mosop/callback.git
  Fetching https://github.com/mosop/string_inflection.git
  Fetching https://github.com/mosop/string_inflection.git
  Fetching https://github.com/crystal-loot/exception_page.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/avram.git
  Fetching https://github.com/luckyframework/lucky_router.git
  Fetching https://github.com/luckyframework/shell-table.cr.git
  Fetching https://github.com/paulcsmith/cry.git
  Fetching https://github.com/crystal-loot/exception_page.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/luckyframework/authentic.git
  Fetching https://github.com/luckyframework/lucky.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/mosop/teeplate.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/avram.git
  Fetching https://github.com/kostya/blank.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/kostya/blank.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/kostya/blank.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/kostya/blank.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/kostya/blank.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/luckyframework/lucky_router.git
  Fetching https://github.com/luckyframework/shell-table.cr.git
  Fetching https://github.com/paulcsmith/cry.git
  Fetching https://github.com/crystal-loot/exception_page.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/luckyframework/lucky_cli.git
  Fetching https://github.com/luckyframework/habitat.git
  Fetching https://github.com/luckyframework/wordsmith.git
  Fetching https://github.com/luckyframework/avram.git
  Fetching https://github.com/luckyframework/lucky_router.git
  Fetching https://github.com/luckyframework/shell-table.cr.git
  Fetching https://github.com/paulcsmith/cry.git
  Fetching https://github.com/crystal-loot/exception_page.git
  Fetching https://github.com/luckyframework/dexter.git
  Fetching https://github.com/luckyframework/lucky_cli.git
```

And so on and so on until eventually it finishes. But it takes *ages*.

shards 0.9.0 and crystal 0.29.0 but also experienced on older versions

kalafut:
I need an `insteadOf` git configuration and this issue has been really slowing things down, so I decided to make an attempt at fixing it. This is literally my first pushed Crystal code so any feedback is welcome!

Fixes #288

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -261,7 +261,7 @@ module Shards
     private def parse_git_uri(raw_uri)
       # Try normal URI parsing first
       uri = URI.parse(raw_uri)
-      return uri if uri.absolute?
+      return uri if uri.absolute? && !uri.opaque?
 
       # Otherwise, assume and attempt to parse the scp-style ssh URIs
       host, _, path = raw_uri.partition(':')
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -244,7 +244,7 @@ module Shards
 
     private def origin_changed?
       @origin_url ||= capture("git ls-remote --get-url origin").strip
-      origins_equal(@origin_url, git_url)
+      !origins_equal(@origin_url, git_url)
     end
 
     # Returns whether origin URLs have matching hosts and paths.
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -35,7 +35,7 @@ module Shards
       refs = git_refs(version)
 
       if file_exists?(refs, SPEC_FILENAME)
-         capture("git show #{refs}:#{SPEC_FILENAME}")
+        capture("git show #{refs}:#{SPEC_FILENAME}")
       else
         raise Error.new("Missing \"#{refs}:#{SPEC_FILENAME}\" for #{dependency.name.inspect}")
       end
@@ -131,7 +131,7 @@ module Shards
 
     def local_path
       @local_path ||= begin
-        uri = URI.parse(git_url)
+        uri = parse_git_uri(git_url)
 
         path = uri.path.to_s[1..-1]
         path = path.gsub('/', File::SEPARATOR) unless File::SEPARATOR == '/'
@@ -242,35 +242,40 @@ module Shards
       false
     end
 
-    private def origin_changed?
+    def origin_url
       @origin_url ||= capture("git ls-remote --get-url origin").strip
-      !origins_equal(@origin_url, git_url)
     end
 
-    # Returns whether origin URLs have matching hosts and paths.
-    #
-    # origins_equal("git@github.com:foo/bar", "https://github.com/foo/bar") # => true
-    protected def origins_equal(origin_1, origin_2)
-      return true if origin_1 == origin_2
-      return false if origin_1.nil? || origin_2.nil?
+    # Returns whether origin URLs have differing hosts and/or paths.
+    protected def origin_changed?
+      return false if origin_url == git_url
+      return true if origin_url.nil? || git_url.nil?
 
-      re = Regex.union(
-        /[\w\.]+@(?<host>[\w\.]+):\/?(?<path>.*)/,         # git@github.com:foo/bar
-        /(ssh|https?):\/\/(?<host>[^:\/\s]+)\/(?<path>.*)/ # https://github.com/foo/bar
-      )
+      origin_parsed = parse_git_uri(origin_url)
+      git_parsed = parse_git_uri(git_url)
 
-      match_1 = re.match(origin_1)
-      match_2 = re.match(origin_2)
+      (origin_parsed.host != git_parsed.host) || (origin_parsed.path != git_parsed.path)
+    end
 
-      return false if match_1.nil? || match_2.nil?
+    # Parses a URI string, with additional support for ssh+git URI schemes.
+    private def parse_git_uri(raw_uri)
+      # Try normal URI parsing first
+      u = URI.parse(raw_uri)
+      return u if u.host && u.path
 
-      ["host", "path"].each do |element|
-        if match_1[element] != match_2[element]
-          return false
-        end
+      # Otherwise, assume and attempt to parse the scp-style ssh URIs
+      host, _, path = raw_uri.partition(':')
+
+      if host.includes?('@')
+        user, _, host = host.partition('@')
       end
 
-      true
+      # Normalize leading slash, matching URI parsing
+      if !path.starts_with?('/')
+        path = '/' + path
+      end
+
+      URI.new(scheme: "ssh", host: host, path: path, user: user)
     end
 
     private def file_exists?(refs, path)
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -271,7 +271,7 @@ module Shards
       end
 
       # Normalize leading slash, matching URI parsing
-      if !path.starts_with?('/')
+      unless path.starts_with?('/')
         path = '/' + path
       end
 
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -260,8 +260,8 @@ module Shards
     # Parses a URI string, with additional support for ssh+git URI schemes.
     private def parse_git_uri(raw_uri)
       # Try normal URI parsing first
-      u = URI.parse(raw_uri)
-      return u if u.host && u.path
+      uri = URI.parse(raw_uri)
+      return uri if uri.absolute?
 
       # Otherwise, assume and attempt to parse the scp-style ssh URIs
       host, _, path = raw_uri.partition(':')
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -131,7 +131,7 @@ module Shards
 
     def local_path
       @local_path ||= begin
-        uri = parse_git_uri(git_url)
+        uri = parse_uri(git_url)
 
         path = uri.path.to_s[1..-1]
         path = path.gsub('/', File::SEPARATOR) unless File::SEPARATOR == '/'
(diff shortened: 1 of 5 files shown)


straight-shoota:
Instead of hard coding the normalization for individual repository providers, it would probably be better to have a more generic solution which would work with *any*  provider.
I think it should suffice to extract hostname and path from both ssh+git and https URLs.

kalafut:
I started that way but was a little concerned about the `git@...` portion, and whether that should be a distinguishing characteristic. i.e. whether `git@github.com:foo/bar` should equal `fred@github.com:foo/bar`. It sounds like that's not a concern, in which case I can definitely simplify this.

kalafut:
@straight-shoota I've generalized this comparison. As it turns out there are actually a _lot_ of allowed git URL schemes. The regexes should take care of the common ones, and exactly matching origins will always work (i.e. current behavior).

RX14:
Could we use `URI.parse` instead of the regex?

test/git_resolver_test.cr
```diff
@@ -38,6 +38,28 @@ module Shards
...
+      assert r.origins_equal("misc_pattern", "misc_pattern")
+      assert r.origins_equal(nil, nil)
+
+      # mismatches
+      assert !r.origins_equal("https://gitlabZ.com/foo/bar", "https://gitlab.com/foo/bar")
+      assert !r.origins_equal("git@github.com:foo/bar2", "https://gitlab.com/foo/bar")
+      assert !r.origins_equal("http://bitbucket.org/foo", "sshZ://bitbucket.org/foo")
+      assert !r.origins_equal("misc_pattern", "misc_pattern_2")
+      assert !r.origins_equal("http://bitbucket.org/foo", nil)
+      assert !r.origins_equal(nil, "http://bitbucket.org/foo")
```
Sija: please, use `refute` instead.

src/resolvers/git.cr
```diff
@@ -243,7 +243,34 @@ module Shards
     end
 
     private def origin_changed?
-      (@origin_url ||= capture("git ls-remote --get-url origin").strip) != git_url
+      @origin_url ||= capture("git ls-remote --get-url origin").strip
```
Sija: IMO this should be extracted to:

```cr
def origin_url
  @origin_url ||= capture("git ls-remote --get-url origin").strip
end
```

Then, there's no need for additional `origins_equal` method (its body can be pasted in here - into `origin_changed?`).

src/resolvers/git.cr
```diff
@@ -243,7 +243,34 @@ module Shards
...
+      match_1 = re.match(origin_1)
+      match_2 = re.match(origin_2)
+
+      return false if match_1.nil? || match_2.nil?
+
+      ["host", "path"].each do |element|
+        if match_1[element] != match_2[element]
+          return false
+        end
+      end
```
Sija: Please, no `Regex` is needed here, `URI.parse` should be more than enough.

test/git_resolver_test.cr
kalafut: 👍

src/resolvers/git.cr
kalafut: Yes, I think that refactor will work fine.

kalafut:
@Sija @RX14 Thanks for the review! `URI.parse` does not handle the SSH URIs that are in the scp style (e.g. `git@github.com:example/project.git`), so something additional was needed. The pair of regexes was compact, but I don't really like adding regexes either if I can avoid it.

I'd also missed @straight-shoota's comment about `local_path` being broken too for these patterns. As a result, I've redone things a bit to make a small git URI parser that uses `URI.parse` but can also deal with some other formats, while still returning a URI object for consistency. This It still feels little sketchy, though there are no regex's and it's probably better over all. It can be using for `origin_changed?` and `local_path`.

I need to add some other tests but with then push it up for y'all to look at.

src/resolvers/git.cr
kalafut: See comment below

kalafut:
I've substantially updated the PR based on feedback, and realizing that `local_path` is also sort of broken. No more ugly regexes either.

straight-shoota:
The implementation of `parse_git_uri` could be much simpler:
```cr
private def parse_git_uri(raw_uri)
  uri = URI.parse(raw_uri)
  return uri if uri.absolute?
  
  URI.parse("ssh:#{raw_uri}")
end
```

kalafut:
Using `.absolute` is a good change.  I don't see how the recommendation handles the main case of git+ssh urls, however. That would just leave everything in the path. e.g. https://play.crystal-lang.org/#/r/8grl

straight-shoota:
Yeah, you're right.

kalafut:
Looks like parse + `absolute` can get confused too:

```
uri = URI.parse("github.com:foo/bar") # scheme => github.com, host => nil, path => foo/bar
p uri.absolute? # => true
```

But that's only If the user (`git@...`) isn't included. While that style is accepted by git, it would seem to be a pretty unlikely real world case. I'd still be OK with `absolute` and not handling those user-less ssh URLs.

straight-shoota:
I suppose shards should allow the exact same formats as accepted by git, so user-less scp-like format `example.com:foo/bar` should be valid. It seems like git treats opaque URLs as scp-like format, so expanding the condition to `uri.absolute? && !uri.opaque?` should be sufficient.

kalafut:
Yep that seems to work OK.

src/resolvers/git.cr
```diff
@@ -242,8 +242,40 @@ module Shards
       false
     end
 
-    private def origin_changed?
-      (@origin_url ||= capture("git ls-remote --get-url origin").strip) != git_url
+    def origin_url
```
ysbaddaden:
```suggestion
    private def origin_url
```
No need to be public AFAIK.

src/resolvers/git.cr
```diff
@@ -242,8 +242,40 @@ module Shards
...
+      return true if origin_url.nil? || git_url.nil?
+
+      origin_parsed = parse_git_uri(origin_url)
+      git_parsed = parse_git_uri(git_url)
+
+      (origin_parsed.host != git_parsed.host) || (origin_parsed.path != git_parsed.path)
+    end
+
+    # Parses a URI string, with additional support for ssh+git URI schemes.
+    private def parse_git_uri(raw_uri)
```
ysbaddaden:
```suggestion
    private def parse_uri(raw_uri)
```

The `git` naming feels redundant —we're already in the `GitResolver` class.

src/resolvers/git.cr
```diff
@@ -242,8 +242,40 @@ module Shards
...
+
+      # Otherwise, assume and attempt to parse the scp-style ssh URIs
+      host, _, path = raw_uri.partition(':')
+
+      if host.includes?('@')
+        user, _, host = host.partition('@')
+      end
+
+      # Normalize leading slash, matching URI parsing
+      if !path.starts_with?('/')
```
ysbaddaden:
```suggestion
      unless path.starts_with?('/')
```

test/git_resolver_test.cr
```diff
@@ -38,6 +38,49 @@ module Shards
...
+        dependency["git"] = origin
+        refute library.origin_changed?, origin
+      end
+
+      # These alternatives should all trigger origin as changed
+      changed_origins = [
+        "https://github.com/foo/bar2",
+        "https://github.com/foos/bar",
+        "https://githubz.com/foo/bar",
+        "file:///github.com/foo/bar",
```
ysbaddaden: This array is missing a ssh+git example.

src/resolvers/git.cr
straight-shoota: Doesn't hurt either, though...

src/resolvers/git.cr
kalafut: 👍

test/git_resolver_test.cr
kalafut: Added tests.

kalafut:
All comments addressed.

ysbaddaden:
Thanks!

PR #317, opened 2020-01-29
Format code (no functional change)

kalafut:
There are a few small formatting diffs that kept popping up during my other PR.


RX14:
We should CI `crystal tool format`, really.

Issue #29, opened 2015-09-09
New command: validate


ysbaddaden:
Since the `shard.yml` has a [SPEC](https://github.com/ysbaddaden/shards/blob/master/SPEC.md) it would be nice to have a validator that ensures that a given `shard.yml` file is valid.

PR #321, opened 2020-02-07
SPEC: Allow periods in shard name

straight-shoota:
This is a proposal to allow period (`.`) in the name of a spec.

The main reason to allow this is because its already in use (since there is no strict enforcement anyways). There are a number of shards ending with `.cr`, for example https://github.com/TPei/progress_bar.cr, https://github.com/krthr/haye.cr, https://github.com/imdrasil/hermes.cr
This is somewhat of an attempt to "legalize" those names.

I'm not aware of any specific reason for prohibiting this character. There might be some source of confusion with respect to file extensions. But since shard names are typically used as directories, I don't see any real issues there.
The general idea is to keep the allowed character limited, but basic special characters should be allowed (dashes and underscores are already allowed).

/cc crystal-lang/crystal#8737

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -68,8 +68,8 @@ The name of library (String, required).
 - It should be lowercase (a-z).
 - It should not contain `crystal`.
 - It may contain digits (0-9) but not start with one.
-- It may contain underscores or dashes but not start/end with one.
-- It must not have consecutive underscores or dashes.
+- It may contain underscores, dashes or periods but not start/end with one.
+- It must not have consecutive underscores, dashes or periods.
 
 Examples: `minitest`, `mysql2`, `battery-horse`.
 


ysbaddaden:
I don't see any benefit for allowing dots in shard names. I see confusion, especially when the shard name ends with `.cr` that is similar to having `crystal` appear in your shard name, which we already recommend _against_.

Sija:
@straight-shoota IMO the repo name and shard name are two different things - there's nothing wrong in having a `.` within the repo name but in the shard name...?

straight-shoota:
@Sija It's not just the repo names. The referenced repos define the shard name ending with `.cr` in their `shard.yml`.

Sija:
IMO that's a bit too much.

straight-shoota:
Thanks for your input. There seems to be no support for periods in shard names.

We could consider taking measures to prohibit them altogether. That would however break existing shards with periods in their name and shards using them as dependencies.
On shardbox.org there are six shards with periods, all ending in `.cr`:

* https://github.com/TPei/progress_bar.cr
* https://github.com/libpixel/libpixel.cr
* https://github.com/krthr/haye.cr
* https://github.com/imdrasil/hermes.cr
* https://github.com/krthr/pope.cr
* https://github.com/jaredsmithse/ynab.cr

None of them is a dependency of any shard known on shardbox.org. So at least it wouldn't have a big impact if those ceased to work.

I'd definitely try to avoid getting new shards with invalid names. I'll update `crystal init` to properly validate the name. `shards init` already sanitizes the folder name.

straight-shoota:
`shards validate` would also notify about invalid shard name (#29).

Also should we maybe issue warnings when such a shard is installed, similar to what #341 added for mismatching tag and SPEC version?

straight-shoota:
Closing. Periods in shard names should not be support.
Warnings can be added when a validation command is introduced.

PR #322, opened 2020-02-07
Use Molinillo to resolve dependencies

### LINKED ISSUE #303: Shards takes hours to resolve dependencies
Shards no longer crashes since #302 but now takes 77 minutes to resolve the following SPEC (copied from #282):

```yaml
name: project
version: 1.1.1

dependencies:
  shainet:
    github: NeuraLegion/shainet
  myhtml:
    github: kostya/myhtml
  cadmium:
    github: watzon/cadmium
    branch: master
  fix:
    github: NeuraLegion/fix
  har:
    github: NeuraLegion/har
    branch: HEAD
  msgpack:
    github: benoist/msgpack-crystal
  kemal:
    github: kemalcr/kemal
    branch: HEAD
  kemal-basic-auth:
    github: kemalcr/kemal-basic-auth
  selenium:
    github: ysbaddaden/selenium-webdriver-crystal
    branch: HEAD
  stumpy_png:
    github: stumpycr/stumpy_png
    version: "~> 4.4.1"
  stumpy_utils:
    github: stumpycr/stumpy_utils

development_dependencies:
  ameba:
    github: veelenga/ameba
  mysql:
    github: crystal-lang/crystal-mysql
```

The SAT solver speed isn't a problem: it's fast enough and CaDiCaL for example, a state-of-the CDCL SAT solver doesn't magically solve the issue (shards still takes forever to resolve dependencies).

The problem is likely the high number of variables and clauses we pass to the solver. We can hardcode the "can install only one version at a time" constraint into the solver, to reduce the number of clauses to a very small number, but it still takes forever to resolve because we didn't reduce the number of variables.

Shards 0.9 already has optimizations in place to reduce the number of variables and clauses: it will only consider accessible dependencies and versions. The problem of the above SPEC is that there isn't any restriction so shards will reach and consider all dependencies and all their versions are accessible. This leads to 139 variables for the above SPEC and a thousand clauses, which leads to 77 minutes to resolve.

Note that adding a few version constraints like `version: ~> x.y.0` to shainet, myhtml, ameba and mysql reduces those 77 minutes to 0.5 seconds. Yes, the performance improve is _that_ dramatic.

waj:
This PR integrates a [port](https://github.com/crystal-lang/crystal-molinillo) of the [Molinillo](https://github.com/CocoaPods/Molinillo) dependency solver.

I tried to keep the PR minimal, maintaining the semantics of the current commands. Only the error reporting is different.

Fixes #303


asterite:
Since nobody mentioned it yet it's probably not a concern, but... is it okay that shards now has a dependency? I'm thinking of how we distribute Crystal, distributions now must install an old shards binary as part of the process.

jhass:
An alternative to requiring an old shards binary is to provide release tarballs for shards that vendor in the dependencies. Basically tar up the result of `git clone shards`, `git checkout tag`,  `shards install`.

waj:
@Sija @RX14 i just did the requested changes
@ysbaddaden I removed the SAT solver, but please at least create a separate shard with it! :)

waj:
Regarding the shard dependency, last resort would be checking out molinillo code into `lib/molinillo` manually before the build

waj:
> Maybe tag a release for molinillo, too.

@ysbaddaden I was thinking we could tag molinillo once we did enough testing

ysbaddaden:
Or add this [Ruby-based bootstrap](https://gist.github.com/ysbaddaden/3ce4cd72a0ad1e8bd4533213785069e1) that will read `shard.lock` and install dependencies at their locked versions.

RX14:
I'd rather just vendor the molinillo resolver into shards. It shouldn't depend on itself.

RX14:
Also @ysbaddaden, that bootstrap is brittle since github refuses to allow you to download any sha1 which is not referred to by a git ref. For a bootstrap you probably have to `--mirror`.

waj:
I just merged this before it starts diverging from master too much. We can continue the discussion about how to integrate better with the molinillo shard.

RX14:
I think that distributing a tarball with `shards` "pre-run" for distros is fine.

j8r:
Another option is to use git subtree / submodule to include `molinillo` inside `shards`, without having to use `shards`.
It would essentially be vendoring, but also keep history.

anatol:
I am trying to build the latest `shards` release 0.10.0 and when I do `make CRFLAGS=--release` the build fails with

```
find: ‘lib/molinillo’: No such file or directory
crystal build src/shards.cr -o bin/shards --release
Showing last frame. Use --error-trace for full trace.

In src/molinillo_solver.cr:1:1

 1 | require "molinillo"
     ^
Error: can't find file 'molinillo'

If you're trying to require a shard:
- Did you remember to run `shards install`?
- Did you make sure you're running the compiler in the same directory as your shard.yml?
make: *** [Makefile:23: bin/shards] Error 1
```

Could you please clarify what is the correct set of the build instructions?

Sija:
@anatol It seems you haven't run `shards install` beforehand.

straight-shoota:
You need to install molinillo into `./lib/molinillo` folder. If you already have a shards executable available, you can just run `shards install` in the root directory. Otherwise `curl https://github.com/crystal-lang/crystal-molinillo/archive/v0.1.0.tar.gz | tar -xf - -C lib` should do.
We should probably add a target to the Makefile.

RX14:
We should add a bootstrap target to to the makefile so you don't need shards to compile shards.

RX14:
@anatol for the PKGBUILD, downloading `https://github.com/crystal-lang/crystal-molinillo/archive/v0.1.0.tar.gz` and then `mv "$srcdir"/crystal-molinillo-0.1.0 "$srcdir"/shards-$pkgver/lib/molinillo` should be your best bet.

straight-shoota:
I've added a bootstrap to #344

Issue #86, opened 2016-01-21
New command: info


ysbaddaden:
The info command asks the registry for informations about a package (repository url, last version, etc).

```
$ shards info minitest
    name: minitest
     url: https://github.com/ysbaddaden/minitest.cr.git
 version: 0.3.0
released: 2016-01-11T15:08:19.000Z
```

straight-shoota:
Additional to the original intent, I think this tool would be useful without querying any remote registry, just presenting information about locally available shards. Both are viable use cases.

Currently, we can query the installed shards using `shards list`, which presents a list of shard names and their installed versions.
But it would also be interesting to query more information about these (installed) shards. For example properties from `shard.yml` like `description`, `license` or `crystal`. Or all available versions. Or properties at a specific version.

The CLI interface could look like this: `shards list [<name> | <ref>] [<version>] [<property>...]`

The first argument can be the name of a shard, a URL or omitted. A name is resolved against the dependencies defined in local `shard.yml`. It can also be a remote reference like (`github:crystal-lang/shards`) to show information about a shard that's not assigned a name in the local context. Integration with a remote index like shardbox.org would also be possible. When omitted, the command applies to the shard in the current directory (like the current behaviour of `shards version`).

The second argument specifies the version and can be omitted. By default it resolves to the installed version for locally installed shards or latest for remote.

Following arguments specify which properties to show. `shards info --version` prints the version, `shards info --name` the name, `--repo` the repository URL, `--versions` lists available versions, `--dependencies` the dependencies etc.
When omitted, a human readable summary is shown, like the sample in the original comment.

An optional `--format` flag could switch output to JSON or other formats.

MatheusRich:
@ysbaddaden and @straight-shoota do you guys think it's possible to add charge licenses as well?

straight-shoota:
@MatheusRich What do you mean with "charge license"?

MatheusRich:
Whoops, that was a mistake, my bad. I meant "shard licenses"

straight-shoota:
I'm sorry, but this still doesn't clarify, what you mean. Where and how do you want to add what kind of license?

MatheusRich:
I'm asking if the command info could include the shard license, like MIT, GPL, etc. I think that would be nice and useful.

straight-shoota:
Yes, of course. The license is already mentioned as an example in https://github.com/crystal-lang/shards/issues/86#issuecomment-588208805

Issue #129, opened 2016-09-28
Leverage libgit2 instead of the git CLI tool


ysbaddaden:
Maybe it would be better to use [libgit2](https://libgit2.github.com/) ([API](https://libgit2.github.com/libgit2/)) instead of relying on the presence of the external CLI tool. It could eventually lead to a better design of GitResolver, a simpler way to detect and report errors (e.g. refs not found) and portability (i.e. windows).

straight-shoota:
There is a shard [crystal-git/libgit2](https://github.com/crystal-git/libgit2) with bindings, but it seems not to be actively maintained and stuck in the middle of a restructuring (https://github.com/crystal-git/libgit2/issues/2#issuecomment-338455484).

ysbaddaden:
I created this issue 2 years ago, and I'm not sure it's such a good idea no. Anyway it won't happen anytime soon (or even later). Shards doesn't need external dependencies to bootstrap itself (apart from Crystal) and I want to keep it that way.

Closing for the time being. If someone is willing to try to integrate libgit2, please try it and report successes and issues :-)

skinnyjames:
Hi all, I'm not sure if this path is a good one, but I've been working on library that wraps libgit2, and I plugged it into the `git_resolver` for this project. (Just cloning and fetching so far).

It's more of a proof a concept than a suggestion, but the tests are passing, which is nice.

If anyone was interested in checking it out, I PR'd against my fork of `shards`.
https://github.com/skinnyjames-forks/shards/pull/1/files

skinnyjames:
Following up on this, I accidentally made a [PR against shards](https://github.com/crystal-lang/shards/pull/586#issue-1718396453), but as Grits is moving forward I think it's worth revisiting this issue.

From the PR:

## Benchmarks

Overall true git clones appear faster than libgit2, but libgit2 clones consume significantly less memory.  If not employed already, concurrency may help.

A benchmark on a repo with many dependencies using `hyperfine` (_gshards_ is the release binary using libgit2).

`hyperfine 'rm -Rf lib && shards install --without-development' 'rm -Rf lib && gshards install --without-development'`

```
Benchmark 1: rm -Rf lib && shards install --without-development
  Time (mean ± σ):      4.076 s ±  0.263 s    [User: 1.844 s, System: 1.789 s]
  Range (min … max):    3.847 s …  4.669 s    10 runs
 
Benchmark 2: rm -Rf lib && gshards install --without-development
[...]

Issue #87, opened 2016-01-21
New command: search


ysbaddaden:
The search command searches the registry for shards matching a pattern:

```
$ shards search s
Search results:
  shards https://github.com/ysbaddaden/shards.git
  selenium-webdriver https://github.com/ysbaddaden/selenium-webdriver-crystal.git
```

PR #324, opened 2020-02-13
Add shards info command

straight-shoota:
This is a basic implementation of #86

It adds a new `shards info` command which can display information about a shard. For now, the features are very crude. The main purpose is to provide a basis for crystal-lang/crystal#8792 to extract the shard name and version from the working directory.
Additional features can be added later.


ysbaddaden:
For starters, this is unrelated to #86 which would search & return information about any shard from a hypothetical registry (i.e. remote shards, not the local project).

For a local SPEC, the info is already right there, and the SPEC format is fully documented and stable. Why not read it directly?

About the docs use-case: having `crystal docs` depend on `shards` makes even less sense to me than having it depend on `libyaml`. The documentation for ruby gems don't print the gem name either, and it's fine.

straight-shoota:
Then #86 completely missed the point to illustrate it's about querying some remote index. I understand it as being primarily concerned with querying information about *installed* shards.

This PR only reflects the shard in current working directory, but an enhancement would be to receive a shard name which would be resolved according to local `shard.yml`.
Querying a remote index would be another possible enhancement. But IMO the core functionality would be to retrieve information about shards and dependencies in a local scope.

Since the shards system works without a central registry, the only real source of truth about features in the ecosystem is the locally available implementation. It is based on unique combination of `shard.yml` configurations with git tags (for the primary resolver). You can obviously put that all together yourself to find information. But I believe the `shards` program should offer a way to reliably query information about locally installed shards, their releases and dependencies. This kind of introspection is already available because it's used internally. It should also be exposed in an external interface to allow other tools (like the docs generator) to easily use this information.

This PR is obviously only a first step into reaching that goal.

> having crystal docs depend on shards makes even less sense to me than having it depend on libyaml.

It's exactly the same reason why shards calls `git` commands instead of binding `libgit`.  `shards` is most usually installed with the compiler. Also it's not a hard dependency. If `shards` command is not available, the doc generator can still work.

Sija:
Similarly, RubyGems also have an `info` command, `gem info pry` here, fwiw:

```yaml
*** LOCAL GEMS ***

pry (0.12.2)
    Authors: John Mair (banisterfiend), Conrad Irwin, Ryan Fitzgerald,
    Kyrylo Silin
    Homepage: http://pryrepl.org
    License: MIT
    Installed at: /usr/local/lib/ruby/gems/2.6.0

    An IRB alternative and runtime developer console
```

straight-shoota:
@Sija Yes, that might be worth pointing out. Most package/dependency managers I'm aware of have such a tool.

ysbaddaden:
> #86 completely missed the point to illustrate it's about querying some remote index. 

Huh? it could hardly be more explicit: "The info command **asks the registry** for informations about a package". It's even linked to #85 which itself links to #87. It also explicitly specifies a shard name in the command line.

> It's exactly the same reason why shards calls git commands instead of binding libgit

Please don't make up historical background. Here is the actual reason:

Shards relies on the external `git` binary because it was _much easier_ to deal with a few simple git commands which I knew very well, than creating bindings from scratch to an external C library which I knew nothing about. We're talking about a couple hours vs days, if not weeks, of work. Yet, that could change, see #129. I often have to install `git` for `shards` to run in containers for example, and always forget to. Having a self contained shards binary would be nice, but it would also make it more complex (and probably introduce another external shard).

Here is the truth: I hate unrequited complexity and adding things for whatever reason and no real benefit. In my eyes, this whole PR is merely:

```yaml
YAML.parse("shard.yml")["name"]
```

I doubt this PR would exist if the SPEC was written in JSON. The "Crystal can't depend on libyaml" argument wouldn't exist anymore —which is already nonsensical to me since Shards already depends on libyaml and its "most usually installed with the compiler" anyway.

Note that the `shards version` command exists because it's easy to run from a macro —where we can't use the YAML bindings (nor JSON or whatever)— and solved an actual issue directly related to shards: to keep a single source of truth about a shard version.

Which means that this PR _could_ become interesting. I just currently fail to grasp it.

@Sija AFAIK the Ruby gem's info command can query locally installed gems which are installed globally, as well as gems from the rubygems registry. This can be useful.

asterite:
Anyone up to porting libyaml to Crystal? At least the parser+scanner and emitter.

There's already a port in Go so my guess is that it shouldn't be hard. Plus it should end up being shorter because of Crystal's syntax, blocks, etc.

That way we can get rid of libyaml and also parse `shards.yml` directly in the compiler for many things (for example docs).

straight-shoota:
There's actually no indication that #86 is about a *remote* registry. Since shards is decentralized, that thought didn't suggest itself. My understanding of *registry* in this context was the collection of locally available shards, as they are referred to by name in `shard.yml`.
But anyway, that's just an exercise in misinterpretation. I mainly wanted to point out that the feature description is completely valid without querying any kind of remote instance.

> Please don't make up historical background.

I just referred to https://github.com/crystal-lang/shards/issues/129#issuecomment-420186428 and I don't think anything, including your explanation here, contradicts my statement. Yes, we could change the compiler to use `libyaml`. But relying on `shards` binary is just easier by a lot. And it's also a good exercise in separating concerns.

> I hate unrequited complexity and adding things for whatever reason and no real benefit.

I honestly see no nothing of that sort. On the contrary, it would add unnecessary complexity to the doc generator if it had to deal with reading shards spec. Even if that was just a one-liner to implement, it still carries a lot of baggage.

I tried to keep this PR focused on a small use case in order to merge quickly what currently blocks advancing the doc generator.
But maybe the general usefulness of this tool is easier acknowledge when looking at the broader scope. I've detailed this in the original issue in https://github.com/crystal-lang/shards/issues/86#issuecomment-588208805. When shards are not only addressed by their name, but a repo URL, it seems very useful to also have command to query the name of that shard which would be already pretty complex to do without a dedicated tool.

@asterite Having libyaml as an external dependency contributes to making it difficult to use in the compiler. We could also use a very simplistic implementation that's barely good enough to reliably parse `name:` and `version:` properties from `shard.yml`. But that's not the point. IMO the main reason to *not* implement parsing `shard.yml` in the compiler/doc generator is separation of concerns. `shards` is the tool for dealing with shard-specific features. There's no point in re-implementing anything of that when the `shards` binary could easily be enhanced to deliver such kind of introspection in a machine-readable way.

PR #326, opened 2020-03-09
Show the shard's name when running scripts

bcardiff:
When running `$ shards install` it will now append the shard name where the script is defined

For example when running it in lucky it will show the precompiled_tasks for avram and make bin from ameba are run.

```
... stripped ...
Installing teeplate (0.8.0)
Installing blank (0.1.0)
Installing db (0.8.0)
Installing pg (0.20.0)
Postinstall script/precompile_tasks (avram)
Postinstall make bin (ameba)
Writing shard.lock
```

When the command fails, [only the command is displayed](https://github.com/crystal-lang/shards/blob/0cda1578a12f6d59254b81c2b7a2386f34a3ea44/src/script.cr#L10) with no path/dependency name.

### DIFF
--- a/src/resolvers/resolver.cr
+++ b/src/resolvers/resolver.cr
@@ -42,8 +42,8 @@ module Shards
 
     def run_script(name)
       if installed? && (command = installed_spec.try(&.scripts[name]?))
-        Shards.logger.info "#{name.capitalize} #{command} (#{dependency.name})"
-        Script.run(install_path, command)
+        Shards.logger.info "#{name.capitalize} of #{dependency.name}: #{command}"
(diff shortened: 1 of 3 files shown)


straight-shoota:
Maybe a more logical ordering would be to put the dependency name before the command? `Postinstall of avram: script/precompile_tasks`. This would also help to keep the names aligned in case the command wraps.

> When the command fails, only the command is displayed with no path/dependency name.

Isn't the shard name a significant information in case of a failure? I'd expect that to show when the command fails as well.

bcardiff:
Feedback addressed @straight-shoota

Issue #310, opened 2019-12-30
still Doesn't support shard renaming and CAN be fixed.


masukomi:
despite issue #272 being closed this is STILL a bug in 0.9.0 and it absolutely can be fixed (potential solution below)

This is a serious issue because it breaks 3 things:

* the ability for a shard owner to rename it (not horrible)
* the ability for a shard owner to recover from accidentally releasing a version with ANY change to the name (accidentally committed typo)
* the ability for anyone to fork a shard and change the name unless they ALSO wipe out all the previous tags. 

If the name changes (intentionally or accidentally) this shard can no longer be installed, nor can any future forks that attempt to rename it to make it clear it's a fork.

The problem is that this method 

```
def add(dependency : Dependency)
```
in `src/solver/graph.cr`

throws an error `unless dependency.name == spec.name` 

This is a bug because it does it for _every historical version_ and it is comparing them against what was specified. Thus if the past is different from the present an exception is raised. 

Here's a potential fix. The assumption of this fix is that we don't care about versions where the name doesn't match. Maybe it was the old name. Maybe there was an accidental typo committed. Let's just acknowledge it and move on. Eventually (in theory) we'll get to a version where the name _does_ match. 

```
      def add(dependency : Dependency)
        pkg = @packages[dependency.name] ||= Pkg.new(dependency)
        resolver = pkg.resolver

        versions_for(dependency, resolver).each do |version|
          next if pkg.versions.has_key?(version)

          if spec = resolver.spec?(version)
            if dependency.name != spec.name
              STDERR.puts("WARNING: shard name (#{spec.name}) doesn't match dependency name (#{dependency.name}) in version #{version}.\n   Skipping this version.")

              pkg.versions[version] = spec
              add(spec)
            end
          else
            # skip (e.g. missing shard.yml)
          end
        end
      end

```

I haven't made a patch for it because it causes all sorts of failures in the tests. 

I think the correct solution should probably also throw an error if NO versions have been added.

ysbaddaden:
> This is a bug because it does it for every historical version

This isn't a bug. Renaming a shard isn't supported, for reasons explained in #272.

Skipping versions for non matching names is interesting... but I can foresee a bunch of issuesnwith this approach. For example, what if I mistype the shard name in my shard.yml? I wont get a name mismatch error anymore, but a no versions available message (what?).

Please dont rename shards. But if you really must, please create a new repository. You can keep the history, but must delete all the version tags.

[...]

PR #327, opened 2020-03-10
Support shard renames

waj:
Currently a shard that was renamed in some point in history cannot be installed unless a version is specified. This is annoying, because for example this `shard.yml` fails with a misleading error:

```yml
name: test
version: 0.1.0

dependencies:
  lucky:
    github: luckyframework/lucky
```

```
$ shards install
Resolving dependencies
Fetching https://github.com/luckyframework/lucky.git
Error shard name (lucky_web) doesn't match dependency name (lucky)
```

This PR allows installing latest version of a shard (when no version is specified) or a previous version only if the name matches.

### DIFF
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -45,7 +45,7 @@ module Shards
 
       packages = [] of Package
       result.each do |v|
-        spec = v.payload.as(Spec) || raise "BUG: returned graph payload was not a Spec"
+        spec = v.payload.as?(Spec) || raise "BUG: returned graph payload was not a Spec"
         v.requirements.each do |dependency|
           unless dependency.name == spec.name
             raise Error.new("Error shard name (#{spec.name}) doesn't match dependency name (#{dependency.name})")
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -46,6 +46,11 @@ module Shards
       packages = [] of Package
       result.each do |v|
         spec = v.payload.as(Spec) || raise "BUG: returned graph payload was not a Spec"
+        v.requirements.each do |dependency|
+          unless dependency.name == spec.name
+            raise Error.new("Error shard name (#{spec.name}) doesn't match dependency name (#{dependency.name})")
+          end
+        end
         resolver = spec.resolver || raise "BUG: returned Spec has no resolver"
         version = spec.version
 
@@ -60,7 +65,11 @@ module Shards
       packages
     end
 
-    def name_for(dependency)
+    def name_for(spec : Shards::Spec)
+      spec.resolver.not_nil!.dependency.name
+    end
+
+    def name_for(dependency : Shards::Dependency)
       dependency.name
     end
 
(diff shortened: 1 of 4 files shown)


straight-shoota:
This basically implements my proposal from https://github.com/crystal-lang/shards/issues/310#issuecomment-569973096 so I agree with that change as it seems like the most expected behaviour and avoids serious ux issues.

We should still consider the critical voices from #310 and especially make sure that edge cases are handled properly.
For example: when `foo` was renamed to `bar` and the the dependency specifies `foo` with no version restriction, it seems like the current implementation would pick the latest release with the name `foo`. I don't think that's correct. It should still pick the latest version overall and then error when that version doesn't match the requested name. If you want to use an older version with `foo` name, you should have to explicitly specify a version.

waj:
I wasn’t aware of that conversation. Thank you for pointing it out. I think it makes sense but I’m not sure the implementation is that simple like yours. Anyway, tomorrow with I’ll analyze it and update this PR.

bcardiff:
We need to document the following known issue: A recently renamed shard must have an explicit branch or ref requirement. Otherwise since the lookup of all available versions and filtering for matching new name leads to an empty list, there is no installation possible.

waj:
@straight-shoota I just updated the PR and the description

src/molinillo_solver.cr
```diff
@@ -46,6 +46,11 @@ module Shards
       packages = [] of Package
       result.each do |v|
         spec = v.payload.as(Spec) || raise "BUG: returned graph payload was not a Spec"
```
straight-shoota: Just noticing this: `v.payload.as(Spec)` can't possibly be falsey because it raises if the type doesn't fit. The raise branch won't be reached ever. Should this be `.as?` instead?

straight-shoota:
> A recently renamed shard must have an explicit branch or ref requirement. Otherwise since the lookup of all available versions and filtering for matching new name leads to an empty list, there is no installation possible.

I don't follow. Why can't the available versions be limited to those matching the dependency name? This should at least work for the name at HEAD.

@waj Could you please add specs for using both old_name and new_name as a dependency *without* a version restriction? And the same for another dependency where the only version with new_name is not tagged.

bcardiff:
@straight-shoota after the push after my comment the implementation and the `test_install_new_when_shard_was_renamed` spec deals with that case. Before the filtering of versions was done in another place leading to the not supported scenario.

straight-shoota:
Oops, I missed that.
Still it would be great to have specs for untagged head and failure specs for `new_name@0.1.0`, `old_name` and `old_name@0.2.0`

src/molinillo_solver.cr
waj: Good catch!

waj:
@straight-shoota I just added the specs you requested. For the untagged head, it works when the "branch" is specified, because there is already tags in the repository. Actually I think this should be always the case, and the switch to "git ref" mode shouldn't be automatic when there is no tags, but that's a topic for another PR  🙂

straight-shoota:
Looks good to me.

But I'd like to get feedback from @ysbaddaden and @RX14 as well because of the concerns voiced in #310.

RX14:
My concern was that it wasn't worth implementing cause of the complexity. This is a small diff, I think it's up to @ysbaddaden now.

PR #328, opened 2020-03-10
Replace glob patterns in Makefile with `find` command

waj:
Currently the Makefile makes use of the `**` wildcard. But that actually doesn't have any effect. I replaced that with `find` calls, that I hope are fully portable 🤞


ysbaddaden:
AFAIK the `shell` function is a GNU make extension and _not portable_.

PR #331, opened 2020-03-11
Conditionally drop the usage of Errno in favor of File::Error

bcardiff:


Sija:
Some `TODO: remove after ...` would be good to have.

bcardiff:
For me all `compare_versions` are subject to be removed eventually. They will go when the mentioned version is wanted to be dropped. That might not depend on the version of the main lib or app. So it's hard to say, to be removed after x.y.z.

ysbaddaden:
Let's not bother with conditionals: master will only compile against crystal master, which is fine IMO.

j8r:
If one needs a latest shards and an older crystal compiler, which is not that common, (s)he can use a statically linked shards, from the official archives or compile it.

src/resolvers/path.cr
```diff
@@ -41,16 +41,29 @@ module Shards
...
-        real_install_path = File.real_path(install_path)
-      rescue errno : Errno
-        if errno.errno == Errno::ENOENT
-          return false
-        else
-          raise errno
+      {% if compare_versions(Crystal::VERSION, "0.34.0-0") > 0 %}
+        begin
+          real_install_path = File.real_path(install_path)
+        rescue ex : File::Error
```
waj: This should be `File::NotFoundError` instead, and no need to check the `os_error`

src/package.cr
```diff
@@ -71,15 +71,27 @@ module Shards
...
-        rescue ex : Errno
-          if {Errno::EPERM, Errno::EXDEV}.includes?(ex.errno)
-            FileUtils.cp(source, destination)
-          else
-            raise ex
+        {% if compare_versions(Crystal::VERSION, "0.34.0-0") > 0 %}
+          begin
+            File.link(source, destination)
+          rescue ex : File::Error
+            if {Errno::EPERM, Errno::EXDEV}.includes?(ex.os_error)
```
waj: Can't just catch any `File::Error` and try with `cp`? I mean without checking the `os_error`. I know it's not exactly the same semantics we had before, but if the hardlink fails for any reason try to use a copy.

bcardiff:
@waj feedback addressed.

I forgot to send/mention that to test this on 0.34.0 we need an update & release of minitest https://github.com/ysbaddaden/minitest.cr/pull/38

RX14:
Shards doesn't have to bootsrap itself from the previous crystal version, there's no need for it to compile on both.

bcardiff:
Shards CI runs on latest, no master. That and the possibility of using a new shards version leads to the need for shards to compile on the latest released Crystal and not just nightly.

I have restarted the osx failure. OSX travis build should go away and build on other provider.

This PR can be merged and later we handle the story for specs on Crystal master.

bcardiff:
Rebased on master. No further changes done.

I will refactor the logging on another PR.

PR #334, opened 2020-03-20
Replace minitest.cr in favor of std-lib spec

bcardiff:
This pr removes minitest to keep dependencies within crystal-lang org.

Since the integration spec use data initialized once arranche the spec to be `spec/unit` and `spen/integration`.

* The Makefile targets remains the same.
* I left as comments the custome failure messages.
* I left the helper assertion methods used in integration specs that have no matching in std-lib spec.
* To make typing happier, I introduced `run!` to be the alternative of `run` that captures the output.
* The MockResolver was dead code so it is removed.

Nothing against minitest.cr (sorry @ysbaddaden!) but is important to keep dependencies within the org.


bcardiff:
Yeah, there some advantage of minitest in some scenarios.

Another thing I miss in the current spec that is present in rspec is the `expect(expr).to ...` because it gives a cleaner prefix of an assertion. Something that minitest have.

---

Was there a reason to not capture all the output in `run` commands or was more an evolution of the code? Because we would make the output to always be captured.

ysbaddaden:
Yes, we should just always capture, it shouldn't be noticeably slower.

spec/integration/spec_helper.cr
```diff
@@ -0,0 +1,172 @@
...
+  def self.created_repositories?
+    @@created_repositories
+  end
+
+  def self.created_repositories!
+    @@created_repositories = true
+  end
+end
+
+Spec.before_each do
```
straight-shoota: Why does this run before each when there's a guard to only executed it once? Can't we just run the setup directly without a hook?

spec/integration/spec_helper.cr
```diff
@@ -0,0 +1,172 @@
...
+  create_file "binary", "bin/foobar", "#! /usr/bin/env sh\necho 'OK'", perm: 0o755
+  create_file "binary", "bin/baz", "#! /usr/bin/env sh\necho 'KO'", perm: 0o755
+  create_git_release "binary", "0.1.0", "name: binary\nversion: 0.1.0\nexecutables:\n  - foobar\n  - baz\n"
+  create_file "binary", "bin/foo", "echo 'FOO'", perm: 0o755
+  create_git_release "binary", "0.2.0", "name: binary\nversion: 0.2.0\nexecutables:\n  - foobar\n  - baz\n  - foo"
+
+  Shards::Specs.created_repositories!
+end
+
+private def assert(value, message, file, line)
```
straight-shoota: Could we replace this with a matcher overload that accepts an `message` argument? That would be a good enhancement for the stdlb spec (for all matchers). We could already add it right here and only need to remove the custom implementation once it's landed in stdlib, not touch the actual specs again.
`#should` already supports `file` and `line` arguments.

spec/integration/spec_helper.cr
```diff
@@ -0,0 +1,172 @@
...
+  assert File.exists?(path), "expected shard.lock to have been generated", file, line
+  locks = Shards::Lock.from_file(path)
+  refute locks.find { |d| d.name == name }, "expected #{name} dependency to not have been locked", file, line
+end
+
+def install_path(project, *path_names)
+  File.join(application_path, "lib", project, *path_names)
+end
+
+def debug(command)
```
straight-shoota: What is this used for?

spec/support/cli.cr
```diff
@@ -0,0 +1,107 @@
+Spec.before_each do
+  application_path.tap do |path|
```
straight-shoota: Why tap?

spec/integration/spec_helper.cr
bcardiff: The custom message story was raised somewhere. I made the helper private so it does not leak outside of this file. The custom messages in this file added some value so that's why I didn't want to loose it.

If that feature lands in the std-lib we can iterate the specs for sure. I see no need to do it right away here.

spec/versions_spec.cr
```diff
@@ -0,0 +1,205 @@
+require "./spec_helper"
+
+private def versions
```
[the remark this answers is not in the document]
bcardiff: It is used in multiple places... but ok...

spec/integration/spec_helper.cr
bcardiff: There is no `Spec.before_all` 🤔 . But we can remove that initialization check and move it to the top level.

spec/integration/spec_helper.cr
bcardiff: I guess it's a helper method to debug failure on commands. It was there I didn't want to disrupt anybody's workflow.

spec/support/cli.cr
bcardiff: Didn't notice that in the code migration. I'll remove it.

bcardiff:
@straight-shoota some feedback addressed. I didn't want to change/refactor the specs a lot. I think that can be handled later. The goal is to drop the non org dependency and allow compilation on master easily.

Improvements in the specs can come later IMO.

We need to fix build on master and promptly tag a molinillo / shards release so Crystal nightlies can build again.

bcardiff:
Regarding the commit history, I've done some rebases along the way and that bothers github ui. But we can squash all these commits if preferred. I commit it that way in case other PRs were merged in between so a rebase would be easier for me.

wontruefree:
I still would like to see a test unit style added to the std-lib.  I think it is a little weird that a project under the Crystal Org cannot have a test unit style.

asterite:
I agree. We should eventually enhance `spec` to provide test-unit like tests. This is not only a cosmetic matter: with test-unit you can define classes, inheritance, initializers (so better setup), share variables, etc.

But we can do it later, even after 1.0

wontruefree:
@asterite I am sure more people have asked but I made a post about it a while ago.
https://github.com/crystal-lang/crystal/issues/7042

asterite:
On second though, there's `minitest.cr` which can be used by any project. I guess it's find that crystal-lang org projects only use specs. That way any core contributor only needs to know spec.

spec/integration/spec_helper.cr
straight-shoota: Oh, I missed the private visibility. It's fine if this is only used in the same file.

straight-shoota:
This is definitely not the topic of this discussion.

bcardiff:
@straight-shoota are we good on your side to merge this as is?

spec/integration/spec_helper.cr
ysbaddaden: Yep. Useful when an integration test starts failing: add a debug run instead of trying to reproduce the test case or printing the output manually.

spec/integration/spec_helper.cr
RX14: There is... https://crystal-lang.org/api/master/Spec/Methods.html#before_all(&block)-instance-method

straight-shoota: That method only works inside a `describe` block. This seems a bit inconsistent, though. We should probably consider improving this API and maybe also add a top-level `before_all`.

ysbaddaden:
Yes, please squash to keep history nice. Also, as @RX14 pointed out, there seems to be `Spec.before_all` after all.

spec/integration/spec_helper.cr
straight-shoota: I just realized there is actually `Spec.before_suite`. And even we core members were not aware it exists :D

bcardiff: Done before_suite is used.

PR #336, opened 2020-03-27
Respect locks when adding new dependencies

waj:
I found this bug that confuses the Molinillo solver and it ends ignoring the lock in some situations like when adding a new dependency that have sub-dependencies already locked.

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -200,6 +200,25 @@ describe "install" do
     end
   end
 
+  it "install subdependency of new dependency respecting lock" do
+    create_git_repository "c"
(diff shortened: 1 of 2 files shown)


spec/integration/install_spec.cr
```diff
@@ -200,6 +200,25 @@ describe "install" do
...
+    create_git_release "c", "0.2.0", "name: c\nversion: 0.2.0\ndependencies:\n  d:\n    git: #{git_path("d")}\n    version: 0.2.0\n"
+    create_git_repository "d"
+    create_git_release "d", "0.1.0", "name: d\nversion: 0.1.0\n"
+    create_git_release "d", "0.2.0", "name: d\nversion: 0.2.0\n"
+
+    metadata = {dependencies: {c: "*", d: "*"}}
+    lock = {d: "0.1.0"}
+
+    with_shard(metadata, lock) do
+      debug "shards install"
```
bcardiff:
```suggestion
      run "shards install"
```

PR #340, opened 2020-03-30
Fix validation in production mode for dependencies referenced by commit

waj:
This is what @RX14 said on a previous PR (https://github.com/crystal-lang/shards/pull/337#issuecomment-605658181)

The logic was indeed wrong and I fixed it and added another test. I changed the order of the tests in the `validate` method because the solver adds a `version` on every lock, even for those with a `commit` [here](https://github.com/crystal-lang/shards/blob/fac08bce2c3482b10437eb8241bfa9565270ec0d/src/molinillo_solver.cr#L25).

Also, because the validation functions are now used only in production mode, the logic is simplified and they check only for exact match, version or commit.

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -200,6 +200,17 @@ describe "install" do
     end
   end
 
+  it "fails to install when dependency requirement (commit) changed in production" do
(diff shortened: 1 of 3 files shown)


RX14:
Thanks!

PR #344, opened 2020-04-04
Add lib to Makefile

straight-shoota:


.travis.yml
```diff
@@ -11,4 +11,4 @@ before_script: |
 
 script:
   - make test
-  - crystal tool format --check
+  - crystal tool format --check src spec
```
Sija, line 14: This change shouldn't be needed since `lib` is ignored from default.

Makefile
```diff
@@ -35,10 +35,16 @@ uninstall: phony
...
 test: test_unit test_integration
 
-test_unit: phony
-	$(CRYSTAL) spec ./spec/unit/*_spec.cr
+test_unit: phony lib
+	$(CRYSTAL) spec ./spec/unit/
 
-test_integration: bin/shards phony
-	$(CRYSTAL) spec ./spec/integration/*_spec.cr
+test_integration: bin/shards phony lib
```
RX14, line 41: `bin/shards` depends on `lib` so this shouldnt be neccesary

RX14:
I'm less sure of this now, I think the behaviour should be this:

- the makefile should warn if `lib` doesn't exist
- the makefile should provide a bootstrap target
- the warning should contain instructions to either run `shards` or `make bootstrap`.

a good bootstrap for the makefile is:
```make
MOLINILLO_VERSION = $(shell $(CRYSTAL) eval 'require "yaml"; puts YAML.parse(File.read("shard.lock"))["shards"]["molinillo"]["version"]')
MOLINILLO_URL = "https://github.com/crystal-lang/crystal-molinillo/archive/v$(MOLINILLO_VERSION).tar.gz"
```
with the target
```make
bootstrap:
	mkdir -p lib/molinillo
	curl -L $(MOLINILLO_URL) | tar -xz -C lib/molinillo --strip-components=1
```

straight-shoota:
@RX14 I've just enhanced the recipe for `lib` with bootstrap as a fallback if `shards install` fails. That way it should work without having to run a second command.
If you want to force the boostrap behaviour, you can run `make lib SHARDS=false`.

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ commands:
       - crystal/shards-install
       - run: make
       - run: make test
-      - crystal/format-check
+      - run: crystal tool format --check src spec
```
RX14, line 20: Why is this required? It really shouldn't be.

RX14:
I think this is fine

.circleci/config.yml
straight-shoota: See commit message on 4d559accab21a258e372c5c4adb8e0cabfc5ca32

> tmp/integrations is created by integration specs and contains non-formatted files

straight-shoota: Maybe `crystal/format-check` (or `crystal tool format`) should only apply to `spec src` by default. But this is also a very specific problem that shards contains ill-formatted files in its tree after the test have run.

Alternatives would be to run format check before `make test` or clean `tmp/integrations` afterwards.

j8r:
Is it possible to move forward? Updating the Alpine Linux shards package requires bootstrapping to get `molinillo`.

straight-shoota:
Rebased and squashed.

PR #349, opened 2020-04-10
Versioning rework

### LINKED ISSUE #325: Index out of bounds error in `shards outdated`
Running `bin/shards outdated` on shards repo (master at 38b3d21f) raises `Index out of bounds` exception:
```console
$ bin/shards outdated
Resolving dependencies
Fetching https://github.com/crystal-lang/crystal-molinillo.git
Fetching https://github.com/ysbaddaden/minitest.cr.git
Unhandled exception: Index out of bounds (IndexError)
  from /usr/share/crystal/src/indexable.cr:0:13 in 'first'
  from src/commands/outdated.cr:47:9 in 'analyze'
  from src/commands/outdated.cr:20:35 in 'run'
  from src/commands/command.cr:27:7 in 'run:prereleases'
  from src/cli.cr:61:11 in '->'
  from /usr/share/crystal/src/option_parser.cr:255:3 in 'parse'
  from /usr/share/crystal/src/option_parser.cr:185:5 in 'parse'
  from src/cli.cr:58:5 in 'run'
  from src/cli.cr:108:3 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:97:5 in 'main_user_code'
  from /usr/share/crystal/src/crystal/main.cr:86:7 in 'main'
  from /usr/share/crystal/src/crystal/main.cr:106:3 in 'main'
  from __libc_start_main
  from _start
  from ???
$ bin/shards --version
Shards 0.9.0 [38b3d21] (2020-02-13)
```

waj:
This ended up being a not so small PR and I have more changes coming, but I prefer to split here now that at least the tests are passing 😅 

This is mostly a refactor, although I fixed some cases of locks with commits that were either not working or working not totally on purpose.

For the user, there are two (non breaking) visible changes:
  * There is a `shard_name.version` file for each shard within the `lib` directory. This replaces the `.sha1` file that was generated for shards not coming from a release version.
  * The `shard.lock` file now contains `version` for both releases and git refs. For versions it's the same a before, but for git refs the version is stores something like: `0.9.0+git.commit.ed686ad3015f48f8b22cf4fbec80e8c8088ef44b`

Internally there are some refactors:
  * Versions contains metadata to represent references to a commit. No more need to check either `commit` or `version`.
  * The interface of the `Resolver` class changed slightly:
    * `versions_for(Dependency)` now returns the list of versions (releases or refs) matching the dependency requirement.
    * `spec(version)` works only with version strings, no more refs. (refs are transformed into versions first by calling the previous method)
    *   `install` is performed by version also, no more refs.
  * Lock files with old format can still be read and they will be transformed when needed.
  * `MolinilloSolver` is now pretty much agnostic to git refs. There is still a minor reference to git metadata that I plan to remove soon. Understanding of whether a version matches a requirement is delegated to the resolver.

All these changes (and the ones coming) will make it easier soon to add or fix features:
  * Know if certain commit still matches a ref
  * Provide information about outdated refs (and fix #325)
  * Install from locks without running the resolver
  * Better support for overrides of shards (i.e: a sub-dependency with a version with an explicit dependency in shard.yml)

### DIFF
--- a/spec/unit/git_resolver_spec.cr
+++ b/spec/unit/git_resolver_spec.cr
@@ -6,10 +6,13 @@ private def resolver(name)
 end
 
 module Shards
-  describe PathResolver do
+  describe GitResolver do
     before_each do
       create_git_repository "empty"
       create_git_commit "empty", "initial release"
+
+      create_git_repository "unreleased"
+      create_git_version_commit "unreleased", "0.1.0"
 
       create_git_repository "library", "0.0.1", "0.1.0", "0.1.1", "0.1.2"
       create_git_release "library", "0.2.0", shard: "name: library\nversion: 0.2.0\nauthors:\n  - julien <julien@portalier.com>"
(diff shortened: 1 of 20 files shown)


src/resolvers/resolver.cr
```diff
@@ -11,33 +11,70 @@ module Shards
...
+    abstract def latest_version_for_ref(ref : String?) : String?
+
+    def spec(version : String) : Spec
+      spec = Spec.from_yaml(read_spec(version))
+      spec.resolver = self
+      spec.version = version
+      spec
+    end
+
+    private abstract def read_spec(version : String)
```
straight-shoota, line 68: Is there a reason for making this this private? It's useful as external API, for example for shardbox which stores the entire spec file in the database.

spec/unit/git_resolver_spec.cr
```diff
@@ -24,11 +27,41 @@ module Shards
...
-      expect_raises(Error) { resolver("empty").read_spec }
-      resolver("library").read_spec.should eq("name: library\nversion: 0.2.0\nauthors:\n  - julien <julien@portalier.com>")
-      resolver("library").read_spec("0.1.1").should eq("name: library\nversion: 0.1.1\n")
-      resolver("library").read_spec("0.1.1").should eq("name: library\nversion: 0.1.1\n")
+    it "available releases" do
+      resolver("empty").available_releases.should be_empty
+      resolver("library").available_releases.should eq(["0.0.1", "0.1.0", "0.1.1", "0.1.2", "0.2.0"])
+    end
+
+    it "latest version for ref" do
```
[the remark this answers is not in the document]
waj: I don't follow. Tags are used for releases. When a branch is used, the versions returned correspond to git commits.

src/resolvers/resolver.cr
waj: The only reason is that `Resolver#spec` would also set the expected version in the returned `Spec` to have precise information about where that spec came from.

Are you using this API in shardbox? I can make it public, but can't you just use the `spec` method instead?

straight-shoota: Shardbox uses both. `spec` to get the parsed data but also `read_spec` to store the raw content of `shard.yml`. Parsing with `Spec.from_yaml` ignores unknown properties, thus loses information that might be useful to have.

spec/unit/git_resolver_spec.cr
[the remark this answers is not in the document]
straight-shoota: There is no example where `latest_version_for_ref(nil)` returns a different result than `latest_version_for_ref(branch)` where `branch` is an actual branch.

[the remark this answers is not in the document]
waj: Oh, I see. `latest_version_for_ref(nil)` returns the version corresponding to the `HEAD`. I'll add that case.

src/resolvers/resolver.cr
waj: Ok, I'll make it public

PR #354, opened 2020-04-16
Rework of dependency and requirements

waj:
This is a refactor of dependency requirements (specifications of which versions are suitable for each dependency).

Currently, each dependency has different fields (`version`, `branch`, `tag`, `commit`) that declares the requirement. But this model is imprecise (allows invalid states), difficult to reason in code (`version` is sometimes a pattern, others an exact version) and requires constant parsing of the `version` field (for example to differentiate patterns, or exact version).

With this change, each `Dependency` now has only three fields: `name`, `resolver` and `requirement`. Also note that the relationship between `Dependency` and `Resolver` is now inverted. (In a future refactor I also plan to remove the `name` from `Resolver` making it totally independent from the dependency that it pretends to install).

`Requirement` is actually a union of these types (structs):
  * `Version`: represents an exact version already resolved. Used in dependencies read from lock files and internally during dependency resolution.
  * `VersionReq`: a version pattern. This corresponds to the `version` field entered in the `shard.yml`
  * `Ref`: an abstract struct representing any reference specific for the resolver
  * `Any`: a wildcard, specifying that any version is suitable

Currently we only have only one resolver (git) that defines subclasses of `Ref`:
  * `GitBranchRef`: corresponds to the `branch` attribute in `shard.yml`
  * `GitTagRef`: corresponds to the `tag` attribute in `shard.yml`
  * `GitCommitRef`: corresponds to the `commit` attribute in `shard.yml`
  * `GitHeadRef`: used internally when there is no version specified and no tags

The parsing of `Ref` is delegated to the resolver.
Also, printing of exact versions is delegated to resolvers, thus producing better output without introducing ad-hoc code.

### DIFF
--- a/spec/unit/spec_spec.cr
+++ b/spec/unit/spec_spec.cr
@@ -5,7 +5,7 @@ module Shards
     it "parse minimal shard" do
       spec = Spec.from_yaml("name: shards\nversion: 0.1.0\n")
       spec.name.should eq("shards")
-      spec.version.should eq("0.1.0")
+      spec.version.should eq(version "0.1.0")
       spec.description.should be_nil
       spec.license.should be_nil
       spec.authors.should be_empty
@@ -58,45 +58,25 @@ module Shards
           branch: master
         local:
           path: /var/clones/local
-          tag: unreleased
       YAML
 
       spec.dependencies.size.should eq(3)
 
       spec.dependencies[0].name.should eq("repo")
-      spec.dependencies[0].path.should be_nil
-      spec.dependencies[0].git.should eq("https://github.com/user/repo.git")
-      spec.dependencies[0].version.should eq("1.2.3")
-      spec.dependencies[0].refs.should be_nil
+      spec.dependencies[0].resolver.should eq(GitResolver.new("repo", "https://github.com/user/repo.git"))
+      spec.dependencies[0].requirement.should eq(version_req "1.2.3")
 
       spec.dependencies[1].name.should eq("example")
-      spec.dependencies[1].path.should be_nil
-      spec.dependencies[1].git.should eq("https://example.com/example-crystal.git")
-      spec.dependencies[1].version.should eq("*")
-      spec.dependencies[1].refs.should eq("master")
+      spec.dependencies[1].resolver.should eq(GitResolver.new("example", "https://example.com/example-crystal.git"))
+      spec.dependencies[1].requirement.should eq(branch "master")
 
       spec.dependencies[2].name.should eq("local")
-      spec.dependencies[2].path.should eq("/var/clones/local")
-      spec.dependencies[2].git.should be_nil
-      spec.dependencies[2].version.should eq("*")
-      spec.dependencies[2].refs.should eq("unreleased")
-    end
-
-    it "fails dependency with duplicate resolver" do
-      expect_raises Shards::ParseError, %(Duplicate resolver mapping for dependency "foo" at line 6, column 5) do
-        Spec.from_yaml <<-YAML
-          name: orm
-          version: 1.0.0
-          dependencies:
-            foo:
-              github: user/repo
-              gitlab: user/repo
-          YAML
-      end
+      spec.dependencies[2].resolver.should eq(PathResolver.new("local", "/var/clones/local"))
+      spec.dependencies[2].requirement.should eq(Any)
     end
 
     it "fails dependency with missing resolver" do
-      expect_raises Shards::ParseError, %(Missing resolver for dependency "foo" at line 4, column 3) do
+      expect_raises Shards::ParseError, %(Unknown resolver "branch" for dependency "foo" at line 5, column 5) do
         Spec.from_yaml <<-YAML
           name: orm
           version: 1.0.0
@@ -107,8 +87,9 @@ module Shards
       end
     end
 
-    it "accepts dependency with extra attributes" do
-      spec = Spec.from_yaml <<-YAML
+    it "fails dependency with extra attributes" do
+      expect_raises Shards::ParseError, %(Unknown attribute "extra" for dependency "foo" at line 6, column 5) do
+        Spec.from_yaml <<-YAML
         name: orm
         version: 1.0.0
         dependencies:
@@ -116,8 +97,20 @@ module Shards
             github: user/repo
             extra: foobar
         YAML
-      dep = Dependency.new("foo", git: "https://github.com/user/repo.git")
-      spec.dependencies[0].should eq dep
+      end
+    end
+
+    it "fails dependency with unexpected attributes" do
+      expect_raises Shards::ParseError, %(Unknown attribute "extra" for dependency "foo" at line 6, column 5) do
+        Spec.from_yaml <<-YAML
+        name: orm
+        version: 1.0.0
+        dependencies:
+          foo:
+            path: ../path
+            extra: foobar
+        YAML
+      end
     end
 
     it "parse development dependencies" do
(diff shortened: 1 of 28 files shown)


spec/unit/spec_spec.cr
```diff
@@ -58,66 +58,59 @@ module Shards
...
       spec.dependencies[2].name.should eq("local")
-      spec.dependencies[2].path.should eq("/var/clones/local")
-      spec.dependencies[2].git.should be_nil
-      spec.dependencies[2].version.should eq("*")
-      spec.dependencies[2].refs.should eq("unreleased")
+      spec.dependencies[2].resolver.should eq(PathResolver.new("local", "/var/clones/local"))
+      spec.dependencies[2].requirement.should eq(Any)
     end
 
-    it "fails dependency with duplicate resolver" do
```
bcardiff, line 85: Why is the "fails dependency with duplicate resolver" spec no longer needed?

src/config.cr
```diff
@@ -7,7 +7,7 @@ module Shards
 
   VERSION_REFERENCE     = /^v?\d+[-.][-.a-zA-Z\d]+$/
   VERSION_TAG           = /^v(\d+[-.][-.a-zA-Z\d]+)$/
-  VERSION_AT_GIT_COMMIT = /\+git\.commit\.([0-9a-f]+)$/
+  VERSION_AT_GIT_COMMIT = /^(\d+[-.][-.a-zA-Z\d]+)\+git\.commit\.([0-9a-f]+)$/
```
bcardiff, line 10: Since this is using `^`, should it allow (or require) the `v` prefix in the version of the commit?

Shouldn't other tags than `vM.m.p` work? It seems the previous regex allowed them.

src/requirement.cr
```diff
@@ -0,0 +1,36 @@
...
+    def prerelease?
+      Versions.prerelease? @value
+    end
+  end
+
+  abstract struct Ref
+  end
+
+  module Any
+    extend self
```
bcardiff, line 32: Unit! ❤️

src/resolvers/git.cr
```diff
@@ -288,7 +361,19 @@ module Shards
...
+    def report_version(version : Version) : String
+      git_version = parse_git_version(version)
+      if commit = git_version.commit
+        "#{git_version.value} at #{commit[0...7]}"
+      else
+        version.value
+      end
+    end
+
+    register_resolver "git", GitResolver
```
bcardiff, line 374: Just checking, we are not losing the work done in #315 by registering the same resolver class for different keys, right?

src/resolvers/path.cr
```diff
@@ -56,15 +48,15 @@ module Shards
       {% end %}
     end
 
-    def available_releases : Array(String)
-      [spec.version] of String
+    def available_releases : Array(Version)
+      [spec(nil).version]
```
bcardiff, line 52: I guess the default value in `def spec(version = nil)` can go away.

spec/unit/spec_spec.cr
waj: Because of the way each dependency attributes are parsed now, any other attribute will be reported as "unknown attribute" instead. After the first key/value is parsed, the rest of the parameters are delegated to the resolver.

src/config.cr
waj: No, that's a pattern for a version with commit metadata (`1.2.3+git.commit.61e81552724e872e94fb52f95eae02e0e7fa6ffd`)

Tags without the `v` prefix are not allowed. Seems like it was supported once but it's not working now. I'm not sure if I did break it but there are not tests anyway. I'll try to fix it soon.

src/resolvers/git.cr
waj: That algorithm works on the full URL, so no.

src/resolvers/path.cr
waj: Maybe. Actually I want to refactor a little bit more the methods that read specs from files. There is some duplication now.

spec/unit/spec_spec.cr
straight-shoota: Please keep that sample. I suppose it would error with `Unknown attribute "gitlab"`? Which feels weird because `gitlab` is a known attribute.

straight-shoota:
The internal refactoring looks nice.

But I'm hesitant about the changes to parsing dependencies. They violate quite a few properties that 
have been established in #306 and previous discussions. For example, it was agreed that extraneaous attributes in dependency mappings should be ignored to enable forward compatibility. And that YAML parsing should be isolated and not depend on resolvers. Requiring the resolver url to be the first entry in the dependency mapping is also a huge breaking change. I can't see how we could possibly accept that. The SPEC does not define a specific order of keys, so the implementation must not assume any. 
Now this breaks with all of that, and I don't see any immediate benefit wich would justify deriving from the previous decisions.
Maybe you can shed some light on that? Is it necessary to break behaviour and why?

I've looked at the dependency mappings available on shardbox.org to get some stats: There are 720 dependencies (about 11%) where the resolver is not the first k/v pair. These would be broken with this change. When taking only dependencies defined by latest releases into account, the number is still 122 broken dependencies (but that's even 20%).

Additional, undefined properties on the other hand are less frequent. There are only 35 instances (about 0,5 %) and they all seem to be misspellings of `version`. So in this regard, a breaking change would be acceptable. But it would break forward compatibility in case we intend to introduce additional properties at some point.

waj:
Big oops! Thank you @straight-shoota for being right there. I thought it could be acceptable to break compatibility because fixing the yml would be an easy thing, but that would also break the parsing of previous versions which is not acceptable. I don't know what I was thinking... sorry and thanks.

I'll change the parsing to keep it compatible. Probably by parsing each dependency attributes to a hash, figure out the resolver type and pass that hash to the resolver to define the requirement.

Now, what do we win by allowing the resolver key to not be the first entry in each map? Maybe it's late to change that, but I think it improves readability of the spec.yml.

RX14:
I just want to say having shardbox to give statistics about these kinds of changes has been amazing and effected real data-driven change to shards. Thank you so much @straight-shoota!

waj:
Absolutely, shardbox is awesome!

I just pushed the changes that revert the (in)compatibility of the parsing.

I also replaced the `Dependency.new(pull)`  with `Dependency.from_yaml` because I didn't like that little game with tuples to workaround issues with nilables.

src/resolvers/git.cr
```diff
@@ -144,20 +186,47 @@ module Shards
...
 
     def git_url
-      dependency.git.not_nil!
+      source.strip
     end
 
-    private def git_refs(version)
-      case version
+    def parse_requirement(params : Hash(String, String)) : Requirement
+      params.each do |key, value|
```
bcardiff, line 193: Should we raise an `Ambiguous dependency requirement for ...` if more than one branch, tag, commit is specified instead of honoring the first one?

src/config.cr
bcardiff: I believe the `v` is required for shard tags. I am missing where the `v` that is in the tag is removed before defining the `package.version.value` to write the lock file.

bcardiff:
I've just noticed the lock file generated after this refactor as follows. Which is semantically fine, but just wondering if the change is intended or not.

```patch
version: 1.0
shards:
  molinillo:
---    github: crystal-lang/crystal-molinillo
+++    git: https://github.com/crystal-lang/crystal-molinillo.git
    version: 0.1.0
```

src/resolvers/git.cr
straight-shoota: Alternatively, pick the most specific one `version > tag > branch`.

straight-shoota:
@bcardiff I guess it doesn't matter much either way. Using explicit `git` is a bit more verbose, but might have a slight advantage towards forward compatibility. If we add more shortcut keys, older shards version would still be able to install from the lock file.

bcardiff:
Yes, I agree it can still work and it might even be a good choice. But since it wasn't stated as an intended change, I wanted to double-check.

waj:
Actually that changed since #306, not my fault 😇

I think it's ok, if we think the lock file is saving a normalized definition of the dependencies

src/config.cr
waj: `package.version` is a `Version`, not a tag. One of the intentions with theses changes is having something more meaningful than just strings everywhere.

spec/unit/spec_spec.cr
waj: Done, it's back!

src/resolvers/git.cr
waj: Actually a `tag` could be more specific than a `version` with a pattern. But they also could be conflicting if a version is not reachable in a branch.

If anything, I would say we could raise instead of considering just the first one. @straight-shoota do you think this would be a breaking change?

We could implement a stricter parser if we want, for the current shard.yml, but be more relaxed when the solver is scanning released versions.

waj: I'd say we could define this later. There is currently not defined behavior with conflicting attributes and this PR is not changing that.

bcardiff: Ok. I’m out of feedback :-)

straight-shoota:
I would have appreciated to get a chance at properly reviewing the updated changes :/

src/resolvers/git.cr
straight-shoota: JFTR: Shardbox database shows no dependencies with more more than one requirement restriction. So breaking things is not a concern.

PR #358, opened 2020-04-17
Format dependency and requirements

waj:
Since #354, the reporting of conflicts didn't look nice. Now it looks back again like this:

```
$ shards install
Resolving dependencies

- `db (0.2.0)` required by `shard.yml`
- `db (~> 0.1.0)` required by `mysql 0.1.0`
```


waj:
> Could you add an integration spec to ensure changes to the user experience won't go unnoticed again?

@straight-shoota done!

PR #364, opened 2020-04-22
Explicitly state build_options in help output, fixes #363

### LINKED ISSUE #363: shards build in release/production mode
If we look at the `shards --help` output, it says that we can pas `<options>` to the build argument:
```bash
build [<targets>] [<options>]  - Build the specified <targets> in `bin` path.
```

The options (that are defined in the same help output) are:
```
Options:
    --no-color                       Disable colored output.
    --version                        Print the `shards` version.
    --production                     Run in release mode. No development dependencies and strict sync between shard.yml and shard.lock.
    --local                          Don't update remote repositories, use the local cache only.
    -v, --verbose                    Increase the log verbosity, printing all debug statements.
    -q, --quiet                      Decrease the log verbosity, printing only warnings and errors.
    -h, --help                       Print usage synopsis.
```

It doesn't say that the build options aren't the ones that are specified in the `shards --help` output, so I assumed (and I'm sure many would fall in the same hole) that `shard build --production` would get me a release ready binary file, but it doesn't, because the correct option for that is  `--release` and `--production` is never actually passed to `crystal build` internally.

So, my suggestions are either:
1. Explicitly specify that the `<options>` in the `build` arg belong to `crystal build` options and not to `shards` options.
*or*
2. Pass `--production` arg to `crystal build` arguments (but of course, as `--release`). 
*or*
3. Remove the ability to build source code thru shards (since it's just a dep manager) and advise to always use `crystal build` instead

Darwinnn:
The `[<options>]` in the `build` argument was a bit misleading, it's more clear now on what options to use with `build`

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -7,7 +7,7 @@ module Shards
       shards [<options>...] [<command>]
 
       Commands:
-          build [<targets>] [<build_options>]  - Build the specified <targets> in `bin` path.
+          build [<targets>] [<build_options>]  - Build the specified <targets> in `bin` path, all build_options are delegated to `crystal build`.
           check                                - Verify all dependencies are installed.
           init                                 - Initialize a `shard.yml` file.
           install                              - Install dependencies, creating or using the `shard.lock` file.
@@ -17,10 +17,7 @@ module Shards
           prune                                - Remove unused dependencies from `lib` folder.
           update [<shards>]                    - Update dependencies and `shard.lock`.
           version [<path>]                     - Print the current version of the shard.
-      
-      Build options:
-          Use "crystal build --help" to see available build options.
-      
+
       General options:
       HELP
     puts opts
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -7,18 +7,21 @@ module Shards
       shards [<options>...] [<command>]
 
       Commands:
-          build [<targets>] [<options>]  - Build the specified <targets> in `bin` path.
-          check                          - Verify all dependencies are installed.
-          init                           - Initialize a `shard.yml` file.
-          install                        - Install dependencies, creating or using the `shard.lock` file.
-          list [--tree]                  - List installed dependencies.
-          lock [--update] [<shards>]     - Lock dependencies in `shard.lock` but doesn't install them.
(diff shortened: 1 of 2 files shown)


src/cli.cr
```diff
@@ -7,18 +7,21 @@ module Shards
...
+          install                              - Install dependencies, creating or using the `shard.lock` file.
+          list [--tree]                        - List installed dependencies.
+          lock [--update] [<shards>]           - Lock dependencies in `shard.lock` but doesn't install them.
+          outdated [--pre]                     - List dependencies that are outdated.
+          prune                                - Remove unused dependencies from `lib` folder.
+          update [<shards>]                    - Update dependencies and `shard.lock`.
+          version [<path>]                     - Print the current version of the shard.
+      
+      Build options:
+          Use "crystal build --help" to see available build options.
```
straight-shoota, line 22: I don't think this should be mentioned here. There are no explanations for other command's arguments either.
That would better fit for a more detailled `shards build --help` if that was a thing.

Darwinnn:
> Could you also update the manpage to use the terms `build_options` and `GENERAL OPTIONS`?

Oh I missed that, thanks! Updated.

src/cli.cr
Darwinnn: Since `shards build --help` is not a thing, I'm assuming we have to somehow clarify that `build_options` have nothing to do with `GENERAL OPTIONS`, same way it's stated in the manpage.

I also don't see a way how we could implement `shards build --help` without copying the info from `crystal build --help` directly and then somehow keeping these two in sync. 

Any ideas?

straight-shoota: Well `shards build --help` wouldn't list the options explicitly but show more details information about the command itself, just like the `build` entry on the manpage.

Darwinnn: Okay, we can specify it in the `build` arg itself, does this seem suitable?
```
build [<targets>] [<build_options>]  - Build the specified <targets> in `bin` path, all build_options are delegated to `crystal build`.
```

Darwinnn: Thanks @RX14! I've updated the code accordingly.

PR #366, opened 2020-04-27
Replace .version files with a single .shards.info

waj:
Since #349 every installed shard is accompanied by a `.version` file with the real installed version. This clutters the `lib` directory. This PR replaces all those files with a single `lib/.shards.info` file. This file is also a YAML so the contents could be extended in the future with any required metadata.

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -49,11 +49,11 @@ describe "install" do
     end
   end
 
-  it "won't fail if version file is missing (backward compatibility)" do
+  it "won't fail if info file is missing (backward compatibility)" do
     metadata = {dependencies: {web: "*"}}
     with_shard(metadata) do
       run "shards install"
-      File.delete("lib/web.version")
+      File.delete("lib/.shards.info")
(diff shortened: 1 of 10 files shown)


waj:
Note that the previous change is not released yet. So a migration path is unnecessary.

Sija:
I'd suggest naming the file `.shards.yml` (or similar but with the `.yml` extname), since it is an YML file after all.

waj:
@Sija I think that name could be confused with `shard.yml`. Also, the `shard.lock` file is also a yml and yet it has a different extension.

bcardiff:
@Sija since the two files have different semantics and different schema I don't think is good to name them so similar. `.shards.info` is fine.

straight-shoota:
@Sija This is only for internal use and the fact that it's based on a YAML format is an implementation detail. We could decide to change the representation at any point and still keep the same file name.

straight-shoota:
We could consider naming the file `shards.installed` or `shards.versions` to better express the purpose. But I'm fine with `shards.info`.

PR #367, opened 2020-04-28
Change install dir to `.crystal/shards`

### LINKED ISSUE #258: name clash for root directory 'lib'
I know it has already been discussed [here](https://github.com/crystal-lang/shards/issues/198), but I wanted to know if you would consider maybe changing the directory used for shards as something less problematic when crystal is used in a mono-repo with several languages?

Something like `crystal_shards` (similar to `node_modules`) might be a good candidate, what do you think?

I'm asking because `lib` is often used in projects I work on: mostly in Ruby, but also in NodeJS and PHP (and I suspect in some other languages that I don't use). I know it's a bit late to change something like that, but Crystal has not reached 1.0 yet and I thought it could be worth it to ask before that time.

waj:
This version will display a warning when the `lib` directory is found but not the `crystal_shards`.

Closes #258

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -45,7 +45,7 @@ describe "install" do
     with_shard(NamedTuple.new) do
       Dir.mkdir "lib"
       stdout = run "shards install --no-color"
-      stdout.should contain(%(W: Shards now installs dependencies into the '_crystal/shards' directory. You may move or delete the legacy 'lib' directory.\n))
+      stdout.should contain(%(W: Shards now installs dependencies into the '.crystal/shards' directory. You may move or delete the legacy 'lib' directory.\n))
     end
   end
 
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -45,14 +45,14 @@ describe "install" do
     with_shard(NamedTuple.new) do
       Dir.mkdir "lib"
       stdout = run "shards install --no-color"
-      stdout.should contain(%(W: Shards now installs dependencies into the 'crystal_shards' directory. You may move or delete the legacy 'lib' directory.\n))
+      stdout.should contain(%(W: Shards now installs dependencies into the '_crystal/shards' directory. You may move or delete the legacy 'lib' directory.\n))
     end
   end
 
   it "doesn't show warning if both legacy and current install path are present" do
     with_shard(NamedTuple.new) do
       Dir.mkdir "lib"
-      Dir.mkdir Shards::INSTALL_DIR
+      Dir.mkdir_p Shards::INSTALL_DIR
       stdout = run "shards install --no-color"
       stdout.should eq(%(I: Resolving dependencies\n))
     end
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -397,7 +397,11 @@ describe "install" do
 
   it "runs postinstall with transitive dependencies" do
     with_shard({dependencies: {transitive: "*"}}) do
-      run "shards install"
+      # Passing this env is needed until the next Crystal release
+      env = {
+        "CRYSTAL_PATH" => "#{Shards::INSTALL_DIR}:#{`crystal env CRYSTAL_PATH`.chomp}",
+      }
+      run "shards install", env: env
       binary = install_path("transitive", "version")
       File.exists?(binary).should be_true
       `#{binary}`.should eq("version @ 0.1.0\n")
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -423,11 +423,7 @@ describe "install" do
 
   it "runs postinstall with transitive dependencies" do
     with_shard({dependencies: {transitive: "*"}}) do
-      # Passing this env is needed until the next Crystal release
-      env = {
-        "CRYSTAL_PATH" => "#{Shards::INSTALL_DIR}:#{`crystal env CRYSTAL_PATH`.chomp}",
-      }
-      run "shards install", env: env
+      run "shards install"
       binary = install_path("transitive", "version")
       File.exists?(binary).should be_true
       `#{binary}`.should eq("version @ 0.1.0\n")
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -53,7 +53,7 @@ describe "install" do
     metadata = {dependencies: {web: "*"}}
     with_shard(metadata) do
       run "shards install"
-      File.delete("lib/.shards.info")
+      File.delete("#{Shards::INSTALL_DIR}/.shards.info")
       run "shards install"
     end
   end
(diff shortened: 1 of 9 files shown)


ysbaddaden:
Please no. This is the _ugliest_ solution.

asterite:
Beauty is in the eye of the beholder.

In node there's `node_modules` which is very explicit. `crystal_shards` is ugly but very explicit.

By the way, we discussed this thoroughly in the last core team meeting so this is kind of decided. We can continue discussing, but we already weighted all the pros and cons in that meeting.

waj:
I hope you mean the chosen name, and not my code 😄

ysbaddaden:
@asterite NPM/Node is a living example of what a dependency management system **must never do**.

@waj Yes, the code is fine. The chosen name is just the worst choice.

Anyway, that's the last death stroke. I'll remove myself from everything crystal related.

asterite:
@ysbaddaden What other name we can choose? We thought about `.crystal/shards` but it's really not nice that the directory is hidden because you sometimes want to go take a look at the code. It's also nice that's it's not hidden so you know what's going on. For instance, `lib` wasn't hidden.

What name did you have in mind?

asterite:
I also mentioned in the meeting that maybe the name "lib" is affecting one or two people that decided to put all their projects in a single directory. For 99.99% of the users the name "lib" is fine. So another option could be not doing anything at all. But "crystal_shards" solves the problem for that minority, so maybe it's good.

ysbaddaden:
I agreed with the change, using `lib` led to conflicts, and it's not nice. But there were much better proposal in the issue. Something like `vendor/crystal` for example, following on Ruby's Bundler as usually recommended (`bundle --path vendor`).

straight-shoota:
Let's continue the discussion about the name in #258

patleb:
@ysbaddaden, I agree that `npm` isn't great, but as long as `shards` puts the dependencies into the project directory under whatever name, it's always at risk of conflict: unless it's using a name that has very little chances to be used (`vendor` isn't one). So, maybe the problem is that `shards` behaves more like `npm` than `bundler` when it comes to the dependencies root path? I think that `crystal_shards` is a sensible choice as a first solution to the problem, but maybe `shards` should take more inspiration from `bundler` which puts dependencies outside by default and gives the option to put them wherever you want by specifying the path if needed? I know that this isn't a trivial amount of work since you would need to manage how dependencies are scoped/shared between different `shards.yml` files, but it could be considered as a second iteration solution or a future feature maybe.

ysbaddaden:
@patleb Sigh. I _never_ said `vendor`, I said `vendor/crystal`.

Anyway. I don't care. Please don't tag me on crystal issues anymore.

patleb:
ysbaddaden, sorry if I offended you. I was referring to the comment in the related issue made about `composer` using `vendor` for its dependencies. I assumed you read it and that you were an active contributor, my bad.

oprypin:
This should not have been merged yet. The change to Crystal itself would need to be merged (and discussed!) first.

Shards@master already puts libs into a directory that Crystal@master has no idea about yet. And in more of a long term consideration, there will be no grace / overlap period where a newer Shards would work with an older Crystal. It's significant because the failure mode is completely implicit and confusing: Shards silently succeeds and Crystal "silently" doesn't see any libs.

Sija:
@oprypin It should work anyway since there is a `:lib` fallback.

patleb:
@oprypin, first, would you mind explaining what you mean by "silently" please? Otherwise, I'm gonna assume that you are misrepresenting the situation to make it look worse than it is, which isn't ok.

Second, it has been discussed and I don't think that I agree with your suggestion to throw the baby with the water. What you are talking about is another matter entirely: version dependency. Shards has an implicit dependency on Crystal internals which could be solved, but it's unrelated to the fact that `lib` isn't a sensible default to put shards in.

Third, as I said, what you're talking about is something internal to Crystal and, by extension, Shards. I don't mean to be judgemental, but relying on internals isn't really a good practice. You could ask to make it easier or less brittle to work with internals of a language/framework, but you should take responsibility first when you're using something considered private. I don't think that the crystal team should be responsible of decisions made by other tools using their internals.

Finaly, I really like that this pull request has defined the constant `Shards::INSTALL_DIR`, this gives other tools a reference to were are the `shards` installed by default.

oprypin:
@patleb My point was,

User starts a project,
adds a dependency,
`shards install` downloads deps into `.crystal/shards`,
(possibly some time passes), user starts relying on that dep,
`crystal build` (or even `shards build`) reports that the library is missing,
user double-checks that the libraries are all there, as intended,
is confused why Crystal just refuses to see them.

oprypin:
> what you're talking about is something internal to Crystal and, by extension, Shards

> you should take responsibility first when you're using something considered private

To me the important part is to consider every situation in a context where Shards doesn't exist. I don't follow your conclusions. Is it that I should not be able to manually add the code of a dependency? What exactly is private -- the ability to create a file as `lib/foo/src/foo.cr`? Changing that to `.crystal/shards/lib/foo/src/foo.cr` would certainly make it *feel* private, but it *shouldn't be*.

patleb:
@oprypin, thx for clarifying, by "silent", you meant that it fails with an unspecific error message. I agree that it's not good, but it's not unsolvable and the current solution was a warning message about migrating the path to `.crystal/shards` and I think that a crystal version check should have been done as well.

As I said, what you're talking about is another problem that came up (implicit dependency on `crystal` internals, which could change through time depending on the version), but it's not what this issue was opened for. I think that it's great that you brought the problem up, but you forced the situation to go back instead of moving forward by solving the other issue (either in this pull request or in another one).

> Is it that I should not be able to manually add the code of a dependency?... What exactly is private -- the ability to create a file as lib/foo/src/foo.cr?

No, what I meant is that the install path is part of the internals and crystal gives a way to configure this. If you want to use the direct value of the internal, you should take into account that the value could change. The correct way (less brittle) is to use the configurable way.

> Changing that to .crystal/shards/lib/foo/src/foo.cr would certainly make it feel private, but it shouldn't be.

Yes, it should be private: an external tool (`shards`) is responsible for it. I think that making `crystal` or `shards` using `lib` for its dependencies led to the confusion you're having about the directory role. @j8r brought the idea of having another default lookup path for vendoring. I think that it's a good idea, but it's not what this issue/pull request was about. The problem, at the moment, is that `lib` could be used for the dependencies managed by `shards`, user libraries managed by the user and vendor libraries also managed by the user or by other tools (but, there will most certainly be some problems down the road which is why I opened the issue). In the world of Ruby and more specifically Rails, this problem is solved by having two more folders: `lib` and `vendor`. But, this is not the responsibility of Ruby nor Bundler to address this issue and these paths aren't added either to the lookup path... do you see the pattern here?

patleb:
Overall, your problems are what I was trying to prevent by opening the issue: clashing with other tools. You're experencing exactly what happens when other people rely on a directory that should be private.

PR #369, opened 2020-04-29
Install dependencies in reverse topological order

### LINKED ISSUE #360: Postinstalls in nested dependencies should run before dependent shards
Given this dependency structure:

- app
  - dep-a
    - dep-b

If `dep-a` and `dep-b` both have `postinstall` scripts, we want `dep-b`'s `postinstall` to run first since `dep-a` may depend on it, just as we would if we ran `shards build` from within in `dep-a`'s own directory locally.

I attempted to solve with #359 but it is not a guaranteed solution — since `app` and `dep-a` could both list `dep-b` as a dependency separately, its placement in the flattened list of dependencies is not guaranteed.

waj:
Fixes #360

### DIFF
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -67,7 +67,7 @@ module Shards
           .resolve(deps, base)
 
       packages = [] of Package
-      result.each do |v|
+      tsort(result).each do |v|
         next unless v.payload
         spec = v.payload.as?(Spec) || raise "BUG: returned graph payload was not a Spec"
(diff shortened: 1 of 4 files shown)


src/molinillo_solver.cr
```diff
@@ -91,6 +91,28 @@ module Shards
       packages
     end
 
+    private def tsort(graph)
+      sorted_vertices = typeof(graph.vertices).new
```
asterite, line 95: 😍

PR #370, opened 2020-04-29
Fix: shard.lock is always overwritten even when dependencies are up to date

waj:
Since #354 unchanged locks are undetected and always overwritten after the "install" command


asterite:
map block to_h is simply to_h block

waj:
@asterite is a little bit rude (and intimidating) to read the mind of others

asterite:
:-)

Sorry for the brevity, cellphone

PR #372, opened 2020-04-30
Ensure the compiler --release flag if we build with --production

jwaldrip:
The documentation was misleading. So I thought it would be proper to automatically add the compiler release flag when building with the shards --production flag.

### DIFF
--- a/src/commands/build.cr
+++ b/src/commands/build.cr
@@ -25,11 +25,10 @@ module Shards
       private def build(target, options)
         Log.info { "Building: #{target.name}" }
 
-        args = [
-          "build",
-          "-o", File.join(Shards.bin_path, target.name),
-          target.main,
-        ]
+        args = ["build", "-o", File.join(Shards.bin_path, target.name)]
+        args << "--release" if Shards.production?
+        args << target.main
+
         options.each { |option| args << option }
         Log.debug { "crystal #{args.join(' ')}" }
 


src/commands/build.cr
```diff
@@ -25,11 +25,10 @@ module Shards
       private def build(target, options)
         Log.info { "Building: #{target.name}" }
 
-        args = [
-          "build",
-          "-o", File.join(Shards.bin_path, target.name),
-          target.main,
-        ]
+        args = ["build", "-o", File.join(Shards.bin_path, target.name)]
+        args << "--release" if Shards.production
```
Blacksmoke16, line 29:
```suggestion
        args << "--release" if Shards.production?
```

src/commands/build.cr
```diff
@@ -25,11 +25,10 @@ module Shards
       private def build(target, options)
         Log.info { "Building: #{target.name}" }
 
-        args = [
-          "build",
-          "-o", File.join(Shards.bin_path, target.name),
-          target.main,
-        ]
+        args = ["build", "-o", File.join(Shards.bin_path, target.name)]
+        args << "--release" if Shards.production?
```
Sija, line 29:
```suggestion
        args << "--release" if Shards.production? && !options.includes?("--release")
```

straight-shoota:
I'm not sure about this. `--production` affects dependency resolution, `--release` affects the compiler process. They don't necessarily have the same use cases. For example in testing and staging environments you might want to use `--production` to resolve dependencies like in production but don't bother about a release build.

Blacksmoke16:
@straight-shoota It that case they can just do `shards install --production`.  However, by using the `build` command with `--production`, IMO the intent is to create a production binary.  In which case `--release` makes sense.  I'd also argue that if you're using a binary for use in testing/staging you ought to be using it as you would in production, i.e. with `--release`.

jhass:
Yes, I think the confusion/wrong expectations here are from its name. Adding  `--release`, renaming `--production` to `--strict` or maybe `--frozen` sounds good to me. And then maybe we can reintroduce `--production`  as a shorthand for the combination of those, at some point.

asterite:
I think we should have just one flag, either `--release` or `--production` that does a combination of those two things (this PR is fine).

It's the same reason why we don't have `-O1`, `-O2`, `-O3` in the compiler. We just have `--release` and that means "this is the build that we are going to deploy, and it better be fast". Having just one flag that means that in shards is ideal in my opinion.

I don't understand why you'd want to deploy a non-optimized build in staging. Then QA finds that stuff is slow for no reason, maybe report it... staging and production should be as similar as possible.

j8r:
I agree with @straight-shoota . I would also like to add that the name `--production` is not a very good name for a flag which installs only "main" libraries (those in the `dependencies` key).

It think, it would be clearer to do this way:
`--no-dev`: do not install any dev packages (those in in `development_dependencies`)
`--release`: compile in release mode
`--production`: combines both `--no-dev` and `--release`

PR #375, opened 2020-05-02
Remove unnecessary require to break loop

straight-shoota:
There was a require loop `resolvers/resolver > spec > config > info > lock > package > resolvers/*` which would cause code like the following to fail:

```Cr
require "./src/resolvers/git"
require "./src/resolvers/path"
```
```
In src/resolvers/path.cr:4:24

 4 | class PathResolver < Resolver
                          ^-------
Error: undefined constant Resolver
```

Lock doesn't depend on Package, so there's no need to require it there.

### DIFF
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -1,5 +1,6 @@
 require "./ext/yaml"
 require "./requirement"
+require "./resolvers/resolver"
 
 module Shards
   class Dependency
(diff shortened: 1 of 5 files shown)


jhass:
Mmh, CI seems to disagree? :)

straight-shoota:
Yeah, it's more messed up. Now it should be good.

PR #378, opened 2020-05-03
Allow Makefile envvars to be overwritten from a command line

anatol:
Signed-off-by: Anatol Pomozov <anatol.pomozov@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,17 @@
 .POSIX:
 
-CRYSTAL = crystal
-SHARDS = shards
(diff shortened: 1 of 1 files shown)


RX14:
Shouldn't `CRYSTAL` and `SHARDS` be overwritable too?

anatol:
If you think that build process needs to customize it then yes. These two envvars need to be overwritable.

RX14:
It'd be useful, like `CC`.

PR #381, opened 2020-05-05
only write lock if there are changes

repomaa:
This will avoid changing the mtime of shard.lock and therefor enabling tasks in makefiles that run shards install and depend on shard.yml and shard.lock. Currently those kinds of tasks would be rerun every time because every shards install will write shard.lock (regardless if there are changes or not).

### DIFF
--- a/src/lock.cr
+++ b/src/lock.cr
@@ -42,6 +42,14 @@ module Shards
     end
 
     def self.write(packages : Array(Package), path : String)
+      File.open("/tmp/shards.log", "a+") do |log|
+        log.puts File.exists?(path)
+        log.puts packages.inspect
+        log.puts from_file(path).inspect if File.exists?(path)
+      end
+
+      return if File.exists?(path) && unchanged?(packages, from_file(path))
+
       File.open(path, "w") do |file|
         write(packages, file)
       end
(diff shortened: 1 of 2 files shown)


waj:
Hi @repomaa, thanks for this. Actually I think there is a better approach already used in the `Install` command to avoid re-reading the `shard.lock` file. It would be probably a good idea to move `generate_lockfile?` and `outdated_lockfile?` to `Command` or even make `write_lockfile` to call them and do nothing when the file is already up to date.

src/lock.cr
```diff
@@ -42,6 +42,14 @@ module Shards
     end
 
     def self.write(packages : Array(Package), path : String)
+      File.open("/tmp/shards.log", "a+") do |log|
+        log.puts File.exists?(path)
+        log.puts packages.inspect
+        log.puts from_file(path).inspect if File.exists?(path)
+      end
```
waj, line 49: is this a leftover? 🙂

straight-shoota:
I don't think this solution is sufficient. When there's a dependecy for `shard.lock` on `shard.yml` in a Makefile, it needs to be ensured that `shard.lock` is touched to have a newer mtime than `shard.yml`. Even if the content didn't change.

straight-shoota:
This was eventually superseded by #444. Thanks @repomaa

PR #382, opened 2020-05-05
make sure tags aren't signed

repomaa:
this prevents errors if users running specs have set up gpg signing in
their git config


straight-shoota:
#245 already introduced that for commits, but tags seem to have been forgotten.

PR #386, opened 2020-05-09
Fix commit hash comparison with abbreviated hash

### LINKED ISSUE #384: Commit version restriction cannot be resolved
I get a strange error on shards master. Using this dependency:
```yaml
dependencies:
  shards:
    github: crystal-lang/shards
    commit: "5666481"
```
```console
$ shards install
Resolving dependencies
Fetching https://github.com/crystal-lang/shards.git

- `shards (commit 5666481)` required by `shard.yml`
```

The shards binary is built from current master `Shards 0.10.0 [ee74934] (2020-05-06)`.

5666481 is definitely a valid commit on this repo, so I see no reason why this should fail.

The latest release build (`Shards 0.10.0 [4091ac5] (2020-04-06)`) works as expected.

straight-shoota:
This also enables full-length commit hash reporting for dependencies. Maybe we could use abbreviated hashes in non-critical situations (such as error messages), but I think it might actually be better to always be explicit. Otherwise there's no indication whether the actual hash value is abbreviated or it's only abbreviated for display.

Resolves #384

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -46,6 +46,10 @@ module Shards
     def initialize(@commit : String)
     end
 
+    def =~(other : GitCommitRef)
+      commit.size >= 7 && other.commit.starts_with?(commit)
+    end
+
     def to_git_ref
       @commit
     end
@@ -161,7 +165,7 @@ module Shards
     def matches_ref?(ref : GitRef, version : Version)
       case ref
       when GitCommitRef
-        git_ref(version) == ref
+        ref =~ git_ref(version)
       else
         # TODO: check branch and tags
         true
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -54,8 +54,8 @@ module Shards
       @commit
     end
 
-    def to_s(io)
-      io << "commit " << @commit[0...7]
+    def to_s(io, *, commit_length = nil)
+      io << "commit " << @commit[0...commit_length]
     end
 
     def to_yaml(yaml)
--- a/spec/unit/dependency_spec.cr
+++ b/spec/unit/dependency_spec.cr
@@ -69,7 +69,7 @@ module Shards
       parse_dependency({foo: {git: "", version: "~> 1.0"}}).to_s.should eq("foo (~> 1.0)")
       parse_dependency({foo: {git: "", branch: "feature"}}).to_s.should eq("foo (branch feature)")
       parse_dependency({foo: {git: "", tag: "rc-1.0"}}).to_s.should eq("foo (tag rc-1.0)")
-      parse_dependency({foo: {git: "", commit: "4478d8afe8c728f44b47d3582a270423cd7fc07d"}}).to_s.should eq("foo (commit 4478d8a)")
+      parse_dependency({foo: {git: "", commit: "4478d8afe8c728f44b47d3582a270423cd7fc07d"}}).to_s.should eq("foo (commit 4478d8afe8c728f44b47d3582a270423cd7fc07d)")
     end
   end
 end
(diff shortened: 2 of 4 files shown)


waj:
Thanks @straight-shoota! I was working on a similar fix already 😅, but I like the operator you used for the comparison.

However, I'm not sure about displaying the full hash when installing. I think it's unnecessary because if it works correctly it's because there was a match with the ref. I might only want to see the value so I can make a quick check if the right commits was selected (and GitHub shows the first 7 digits as well, so the visual match is easier). Bundler does a similar thing. It shows the first 7 chars of the hash, no matter how many digits were given in the Gemfile:
```
Using json 2.3.0 from https://github.com/flori/json (at master@da50acc)
```
When there is an error, it's different, because the user needs to understand why the value didn't match:
```
Revision ffffffffffffffff does not exist in the repository https://github.com/flori/json. Maybe you misspelled it?
```

The two values live in different objects in Shards. The value that was matched is always a `Version` (a unequivocal identifier of a single version), while the value given by the user in `shards.yml` is a `GitCommitRef` in this case. In other words, I think commit hash from `Version`, which is always full no matter what the user provides, must be abbreviated (look at `GitResolver#report_version`), while the `GitCommitRef` which is only printed on diagnostics or error messages, must be always full.

src/resolvers/git.cr
```diff
@@ -46,12 +46,16 @@ module Shards
     def initialize(@commit : String)
     end
 
+    def =~(other : GitCommitRef)
+      commit.size >= 7 && other.commit.starts_with?(commit)
```
waj, line 50: Why impose this minimum limit? I think git accepts a minimum of 4 digits if there is no conflicts.

src/resolvers/git.cr
```diff
@@ -46,12 +46,16 @@ module Shards
...
+    end
+
     def to_git_ref
       @commit
     end
 
-    def to_s(io)
-      io << "commit " << @commit[0...7]
+    def to_s(io, *, commit_length = nil)
```
waj, line 57: I don't see where this `commit_length` argument is in use

src/resolvers/git.cr
straight-shoota: At this point it's not trivial to check whether there are no conflicts. With a minimum size of 7 the probability for a conflict is very low.

I don't think that's the correct solution though. If we keep this behaviour it should already fail when reading a hash that is too short for comparison.
But a better solution would probably to check whether the abbreviated commit is actually conflict-free by asking git to resolve it. Then we can be sure that the starts_with? match is valid.

src/resolvers/git.cr
straight-shoota: I originally wanted to use that to display different lengths depending on context, but decided that's too complex for this fix PR.

src/resolvers/git.cr
waj: Because at that point it shouldn't make any assumptions at all. It would be correct even allowing no minimum size for the hash. The resolution will fail at a different place if there is not matching commit or if there is a conflict.

The second part is already being done, although it could handle the error differently to show a more meaningful message. For example, when the hash is incorrect or have conflicts it's currently showing:

```

- `db (commit abcdef)` required by `shard.yml`
```

That's because `GitResolver` already returned an empty set for the available versions.

src/resolvers/git.cr
waj: Please, check my comment: https://github.com/crystal-lang/shards/pull/386#issuecomment-626090079

Actually I think the right way to solve this is raise an error with a better message at `latest_version_for_ref` when there is no available matches, (instead of returning `nil`). That message should contain the exact input read from the `shard.yml`. And leave here the abbreviated commit. At that point it will be used only to show conflicts with other dependencies, but already knowing the commit exists.

straight-shoota:
@waj do you want to take over? You probably know better how to deal with this.
I just though it was quick and easy and wanted to help out 😆

waj:
@straight-shoota absolutely! 😄 I'll schedule some time to write some docs too.

waj:
Superseded by #388 and #389

PR #387, opened 2020-05-10
Add error on duplicate arguments in shard.yml

### LINKED ISSUE #376: Duplicate keys in spec
When shard.yml contains duplicate keys for example two entries called `dependencies`, the spec parser concatenates the values:

```cr
yaml = <<-YAML
name: foo
version: 0.1.0

dependencies:
  foo:
    github: foo/foo
dependencies:
  bar:
    github: bar/bar
YAML
spec = Shards::Spec.from_yaml yaml

spec.dependencies.map(&.name) # => ["foo", "bar"]
```

This is unexpected and differes from any other YAML parser where previous values are simply overriden. For example `YAML.parse` behaves that way:
```cr
YAML.parse(yaml)["dependencies"].as_h.keys # => ["bar"]
```

As a result of this behaviour normalized YAML (for example passing `yaml` through `YAML.parse(yaml).to_yaml` parses differently than the ill-formed spec with duplicate keys.

I'd expect the parser to behave like other YAML parses and only use the last value. That's how it worked before #349. We could also consider to fails parsing such a a spec. Duplicate keys are clearly an error in the shard spec and should be fixed by the author. But issueing a warning is probably better.

straight-shoota:
Resolves #376

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -56,6 +56,7 @@ dependencies.
 
 - It must be a valid YAML file.
 - It must be UTF-8 encoded.
+- It must not contain duplicate attributes.
 - It should be indented with 2 spaces.
 - It should not use advanced YAML features, only simple hashes, arrays and strings.
 
--- a/spec/unit/spec_spec.cr
+++ b/spec/unit/spec_spec.cr
@@ -264,5 +264,47 @@ module Shards
       YAML
       expect_raises(ParseError) { Spec.from_yaml(str) }
     end
+
+    it "errors on duplicate attributes" do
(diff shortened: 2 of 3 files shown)


waj:
This is ok, and I'll merge it right away because it solves the issue. But I wonder if the `YAML` parser shouldn't have an option to be strict in this case. Note that this solution only handles duplicates at the top level, but it ignores on other places.

straight-shoota:
Yeah, I'd also like to add a feature to the parser. Same goes for `JSON`, too, I guess.

PR #389, opened 2020-05-12
Make errors with git refs more descriptive

waj:
Currently when a git ref doesn't exist, the resolver just returns an empty array for the available versions. This makes the solver to fail with a confusing message. For example:
```

- `db (branch foo)` required by `shard.yml`
```

With this PR, the error is handled more prematurely, thus providing better info:
```
Could not find branch foo in the repository https://github.com/crystal-lang/crystal-db.git for shard "db"
```

Also, if a commit hash is used, the full hash is displayed:
```
Could not find commit fd686ad3015f48f8b22cf4fbec80e8c8088ef44b in the repository https://github.com/crystal-lang/crystal-db.git for shard "db"
```

Related to #384, #386

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -605,4 +605,36 @@ describe "install" do
       stdout.should contain(%(W: Shard "noshardyml" version (0.1.0) doesn't have a shard.yml file))
     end
   end
+
+  it "shows error when branch does not exist" do
+    metadata = {dependencies: {web: {git: git_url(:web), branch: "foo"}}}
+    with_shard(metadata) do
+      ex = expect_raises(FailedCommand) { run "shards install --no-color" }
+      ex.stdout.should contain(%(E: Could not find branch foo in the repository #{git_url(:web)} for shard "web"))
+    end
+  end
+
+  it "shows error when tag does not exist" do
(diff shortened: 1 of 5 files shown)


spec/integration/install_spec.cr
```diff
@@ -605,4 +605,36 @@ describe "install" do
       stdout.should contain(%(W: Shard "noshardyml" version (0.1.0) doesn't have a shard.yml file))
     end
   end
+
+  it "shows error when branch does not exist" do
+    metadata = {dependencies: {web: {git: git_url(:web), branch: "foo"}}}
+    with_shard(metadata) do
+      ex = expect_raises(FailedCommand) { run "shards install --no-color" }
+      ex.stdout.should contain(%(E: Could not find branch foo in the repository #{git_url(:web)} for shard "web"))
```
straight-shoota, line 613: I'd suggest to change the order of shard and url: 
```suggestion
      ex.stdout.should contain(%(E: Could not find branch foo for shard "web" in the repository #{git_url(:web)}))
```

The shard name is typically the more relevant information because that's where you look it up in the dependencies specification. Shard names are usually relatively short while git URLs can be pretty long, so it's easier to read that way.

Issue #365, opened 2020-04-27
RFC: shard.yml crystal semantic


bcardiff:
## Motivation

I think it is important to allow shards to work and deal with dependencies that will require some range of language/std-lib version to work with. Up until now we frameworks usually force to run on Crystal latest only for convenience as the ecosystem mature and gets bigger.

## Current status

Currently [SPEC](https://github.com/crystal-lang/shards/blob/master/SPEC.md) says:

> The last known Crystal version that is capable to compile the Shard (String).
> Purely informational.

* I've seen shards with no crystal version key
* I've seen shards that do not change the crystal version in the `shard.yml`
* I use it in crystal-db as the minimum required version

## Thoughts

I think we should enforce a different semantic on the crystal version mentioned in `shard.yml`.

For a given shard release, if we restrict that it will work with `crystal <= upper_bound` this means that:

* shards resolver could directly discard old releases that will not work with the current crystal version used in the project
* will push the community to keep an eye and maintain the shard more thoroughly once a `crystal > upper_bound` is released.
  * Some of the updates might by lifting the `upper_bound` if everthing is fine.
  * might delay the whole ecosystem to come up to date and be available to the latest crystal version.

For a given shard release, if we restrict that it will work with `crystal >= lower_bound` this means that:

* shards will be able to avoid upgrading the shard beyond what is supported in the current crystal version used in that project. (this is important if the current crystal version of the project is not the latest crystal release).

Depending on the `upper_bound` chosen it might become a burden for the ecosystem to upgrade things on every minor crystal release.

Picking an `upper_bound` should balance between a guess on what could happen and committing to a maintenance for lifting.

Although shards packages are not required to adhere to semver, the compiler and std-lib could / should.

## Proposal

Even if we don't implement the whole logic now in shards, it is good to state what is the semantic  of the crystal key.

If we assume Crystal & std-lib will adhere to semver then we can push the burden of forcing the ecosystem to republish shards on mayor releases only.

So, I would like to propose that `crystal: x.y.z` be treated as `~> x.y, >= x.y.z` (ie: `>= x.y.z, < (x+1).0.0`). When `x >= 1` this is what semver should guarantee, right?

[...]

PR #394, opened 2020-05-19
Support intersection of versions

waj:
This adds support for version specifiers like: `>= 1.0.0, < 2.0`
I need this to implement #365, but I think it's also nice addition. Other dependency managers also support expressions like that.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -167,7 +167,7 @@ A version requirement (String).
 - It may be a version number.
 - It may be `"*"` if any version will do.
 - The version number may be prefixed by an operator: `<`, `<=`, `>`, `>=` or `~>`.
-- Many expressions can be separated by commas.
+- Multiple requirements can be separated by commas.
 
 Examples: `1.2.3`, `>= 1.0.0`, `>= 1.0.0, < 2.0` or `~> 2.0`.
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -167,8 +167,9 @@ A version requirement (String).
 - It may be a version number.
 - It may be `"*"` if any version will do.
 - The version number may be prefixed by an operator: `<`, `<=`, `>`, `>=` or `~>`.
+- Many expressions can be separated by commas.
 
-Examples: `1.2.3`, `>= 1.0.0` or `~> 2.0`.
+Examples: `1.2.3`, `>= 1.0.0`, `>= 1.0.0, < 2.0` or `~> 2.0`.
 
 Most of the version operators, like `>= 1.0.0`, are self-explanatory, but
 the `~>` operator has a special meaning, best shown by example:
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -41,6 +41,14 @@ describe "install" do
     end
   end
 
+  it "resolves intersection" do
+    metadata = {dependencies: {web: ">= 1.1.0, < 2.0"}}
+    with_shard(metadata) do
+      run "shards install"
+      assert_installed "web", "1.2.0"
+    end
+  end
+
   it "fails when spec is missing" do
     Dir.cd(application_path) do
       ex = expect_raises(FailedCommand) { run "shards install --no-color" }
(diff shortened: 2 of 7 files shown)


SPEC.md
```diff
@@ -167,8 +167,9 @@ A version requirement (String).
 - It may be a version number.
 - It may be `"*"` if any version will do.
 - The version number may be prefixed by an operator: `<`, `<=`, `>`, `>=` or `~>`.
+- Many expressions can be separated by commas.
```
straight-shoota, line 170:
```suggestion
- Multiple requirements can be separated by commas.
```

Sija:
@waj FWIW, I believe that a _compound_ keyword is more apt in this case than _intersection_. I'd use _compound requirement_ - as in bundler, see:

https://github.com/rubygems/rubygems/blob/6f1037c67557a33a81ead0a6b2b5867d601f0d08/bundler/spec/install/gems/resolving_spec.rb#L241

Blacksmoke16:
Compose calls them [version ranges](https://getcomposer.org/doc/articles/versions.md#version-range).

They also seem to allow `||` together ranges.  `>=1.0 <1.1 || >=1.2`.

Sija:
@Blacksmoke16 I don't know about the Composer, but they're not _ranges_, that's for sure :)

Blacksmoke16:
@Sija How is it not? `>= 1.0.0, < 2.0` would be `1.0.0...2.0.0`, i.e. any version greater than or equal to `1.0.0` but less than `2.0.0`.

Sija:
@Blacksmoke16 because `>= 1.0.0, < 2.0, != 1.1`

Blacksmoke16:
@Sija Still could be interpreted as `1.0.0...1.1.0 || 1.2.0...2.0.0`. 🤷.

Sija:
@Blacksmoke16 Yeah, that's a _compound_ (range) statement you've got there...

waj:
I prefer the word "intersection" because it's more explicit about how the expressions are composed. In the future, if we add other operators like `||` or `...` the text can be rephrased.

Sija:
@waj That's not _intersection_ though... And the comma used there is analogous to `&&` (and) logical operator.

waj:
@Sija It's the intersection of the version sets defined by each expression. In math intersection and logical and are very close concepts. Also, I'm not the only one using this term for this: https://dart.dev/tools/pub/dependencies#traditional-syntax

Sija:
@waj Not if you add `!=` operator to the mix.

bcardiff:
@Sija It’s intersection. All comma separated constraints need to be satisfied. Even if != is supported. https://www.siyavula.com/read/maths/grade-10/probability/14-probability-03

If or/union is added later is another matter.

Sija:
> ### Intersection
> The intersection of two sets is a new set that contains all of the elements that are in both sets. The intersection is written as 𝐴∩𝐵 or “𝐴 and 𝐵”.

Quite likely I'm misunderstanding the term in this context but I don't see any (_intersecting_) _sets_ here. I mean we have constraints, yes, but not sets.

By set intersection I'd understand having `1.0-1.2, 1.1-1.3`. OTOH `>= 1.0.0, < 2.0` defines a _set_ of versions between `1.0` and `2.0`, which is closer to a range in this case than to a _intersection_... adding multiple closes turns that into a _compound_ statement.

But maybe my brain is playing tricks on me 🤷‍♂️

waj:
@Sija do you realise we're discussing for a word it doesn't even appear in the documentation, right? It's just the issue name and a spec description at most 😄 

It's not brain tricks. Just different ways to see the same. When you say `>= 1.0.0` that's a range too. Just an unbounded range. And a range is a set. And sets can be intersected to create a new set. But while all ranges are sets, not all sets are ranges. For example, `>= 1.0.0, < 2.0, != 1.2.3` cannot be expressed as a single range (note that `!=` is currently not supported). Still can be interpreted as the intersection of three sets. Or a union of two ranges if you want, but for me it's harder to think in ranges because is more complicated to generalize.

Sija:
With this merged in, it would be good to add support for `!=` operator.

PR #395, opened 2020-05-22
Use `crystal` property to filter shard compatibility

waj:
This PR implements what's discussed in #365 

I had to also make the assumption that when the `crystal` property is not present, the expression `< 1.0.0` is assumed. This is because old shards in the past didn't specify the `crystal` version. If we don't make this assumption, once Crystal 1.0.0 is released, old shard versions will be installed by default. This has the implication that the parameter is now mandatory after `1.0.0` but I think that's a good thing and according to what was discussed.

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -666,6 +666,14 @@ describe "install" do
     end
   end
 
+  it "install version ignoring current crystal version if --ignore-crystal-version" do
+    metadata = {dependencies: {incompatible: "*"}}
+    with_shard(metadata) do
+      run "shards install --ignore-crystal-version", env: {"CRYSTAL_VERSION" => "0.3.0"}
+      assert_installed "incompatible", "1.0.0"
+    end
+  end
+
   it "doesn't match crystal version for major upgrade" do
     metadata = {dependencies: {incompatible: "*"}}
     with_shard(metadata) do
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -673,4 +673,12 @@ describe "install" do
       refute_installed "incompatible"
     end
   end
+
+  it "does match crystal prerelease" do
+    metadata = {dependencies: {incompatible: "*"}}
+    with_shard(metadata) do
+      run "shards install", env: {"CRYSTAL_VERSION" => "1.0.0-pre1"}
+      assert_installed "incompatible", "1.0.0"
+    end
+  end
 end
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -8,7 +8,7 @@ describe "install" do
     }
 
     with_shard(metadata) do
-      run "shards install"
+      debug "shards install"
 
       # it installed dependencies (recursively)
       assert_installed "web", "2.1.0"
@@ -184,8 +184,8 @@ describe "install" do
 
   it "resolves dependency spec at locked commit" do
     create_git_repository "locked"
-    create_git_release "locked", "0.1.0", "name: locked\nversion: 0.1.0\n"
-    create_git_release "locked", "0.2.0", "name: locked\nversion: 0.2.0\ndependencies:\n  pg:\n    git: #{git_path("pg")}\n"
+    create_git_release "locked", "0.1.0"
+    create_git_release "locked", "0.2.0", {dependencies: {pg: {git: git_path("pg")}}}
 
     metadata = {
       dependencies: {
@@ -438,9 +438,9 @@ describe "install" do
 
   it "fails with circular dependencies" do
     create_git_repository "a"
-    create_git_release "a", "0.1.0", "name: a\nversion: 0.1.0\ndependencies:\n  b:\n    git: #{git_path("b")}"
+    create_git_release "a", "0.1.0", {dependencies: {b: {git: git_path("b")}}}
     create_git_repository "b"
-    create_git_release "b", "0.1.0", "name: b\nversion: 0.1.0\ndependencies:\n  a:\n    git: #{git_path("a")}"
+    create_git_release "b", "0.1.0", {dependencies: {a: {git: git_path("a")}}}
 
     with_shard({dependencies: {a: "*"}}) do
       ex = expect_raises(FailedCommand) { run "shards install --no-color" }
@@ -620,7 +620,7 @@ describe "install" do
   it "install dependency with no shard.yml and show warning" do
     metadata = {dependencies: {noshardyml: "0.1.0"}}
     with_shard(metadata) do
-      stdout = run "shards install --no-color"
+      stdout = run "shards install --no-color", env: {"CRYSTAL_VERSION" => "0.34.0"}
       assert_installed "noshardyml", "0.1.0"
       stdout.should contain(%(W: Shard "noshardyml" version (0.1.0) doesn't have a shard.yml file))
     end
@@ -655,6 +655,22 @@ describe "install" do
     with_shard(metadata) do
       ex = expect_raises(FailedCommand) { run "shards install --no-color" }
       ex.stdout.should contain(%(E: No shard.yml was found for shard "noshardyml" at commit #{git_commits(:noshardyml)[1]}))
+    end
+  end
+
+  it "install version according to current crystal version" do
+    metadata = {dependencies: {incompatible: "*"}}
+    with_shard(metadata) do
+      run "shards install", env: {"CRYSTAL_VERSION" => "0.3.0"}
+      assert_installed "incompatible", "0.2.0"
+    end
+  end
+
+  it "doesn't match crystal version for major upgrade" do
+    metadata = {dependencies: {incompatible: "*"}}
+    with_shard(metadata) do
+      ex = expect_raises(FailedCommand) { run "shards install --no-color", env: {"CRYSTAL_VERSION" => "2.0.0"} }
+      refute_installed "incompatible"
     end
   end
 end
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -8,7 +8,7 @@ describe "install" do
     }
 
     with_shard(metadata) do
-      debug "shards install"
+      run "shards install"
 
       # it installed dependencies (recursively)
       assert_installed "web", "2.1.0"
--- a/src/config.cr
+++ b/src/config.cr
@@ -65,7 +65,17 @@ module Shards
   end
 
   def self.crystal_version
-    ENV["CRYSTAL_VERSION"]? || `crystal env CRYSTAL_VERSION`.chomp
+    @@crystal_version = ENV["CRYSTAL_VERSION"]? || begin
+      output = IO::Memory.new
+      error = IO::Memory.new
+      status = begin
+        Process.run("crystal", {"env", "CRYSTAL_VERSION"}, output: output, error: error)
+      rescue e
+        raise Error.new("Could not execute 'crystal': #{e.message}")
+      end
+      raise Error.new("Error executing crystal:\n#{error}") unless status.success?
+      output.to_s.strip
+    end
   end
 
   class_property? production = false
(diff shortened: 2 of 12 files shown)


src/molinillo_solver.cr
```diff
@@ -150,7 +151,20 @@ module Shards
     end
 
     def dependencies_for(specification : S) : Array(R)
-      specification.dependencies
+      return specification.dependencies if specification.name == "crystal"
+      crystal_pattern =
```
bcardiff, line 155: so neat 😻

bcardiff:
> Why would old versions be installed? Shards should always pick the newest version that's compatible.

if my_shard 0.1.0 is published today without crystal restriction,
then is published as 0.2.0 with crystal: 1.1.0 restriction,
when crystal 2.0 is out the compatible version will be odly 0.1.0.

I think is better to force a < 1.0.0 as default restriction. It will be no worse than in the next major release. Whether that mean changing the source code or just bumping the crystal restriction.

Having a default unrestricted crystal version seems as a bad practice as pointing to a shard repo without version restriction.

waj:
> Why would old versions be installed? Shards should always pick the newest version that's compatible.

Consider this scenario: a shard today published to be compatible with 0.x was published in the past without the constraint (because it didn't exist). Once Crystal 1.0 is released, if you try to install that shard, an old version will be installed, instead of just raising an error because the shard was not updated.

A desirable side effect of this is that unmaintained shards clear themselves from the ecosystem. This is the approach that Elm uses and it works really well. Making the `crystal` version optional allows (already) old versions of shards to last forever.

On the other hand, I don't see any advantage on keeping the property optional. Is it possible to write a shard that is compatible with any Crystal version in the future? Anyway, if someone want to do that is still possible by writing `crystal: >= 0` but I'd prefer to see that explicit.

straight-shoota:
This hasn't been a major problem so far, with constantly breaking Crystal versions. Why would it be a problem in the future?

I don't know any dependency managers that make such a restrictive implication about language version compatibility.

This change makes **all** shards incompatible with Crystal 1.0. Every shard would need to make a new release, even if nothing has changed (and it may have actually been compatible with Crystal 1.0 for the last X releases).

Even when we bracket the semantics of the property being absent, 60% of all shards currently use this property according to the old semantics, i.e. optional indication of minimal compatible version. All these shards would no longer install with this change to dependency resolution.

Such a huge disruption is completely unacceptable. It breaks the entire ecosystem.

The only action we can take right now is to update the specification to describe the intended semantics. But it can't have an immediate effect as long as the vast majority of shards is still on the old semantics. And even the new semantics seem to need some discussion about the meaning of absence.
The following step could be some kind of warning, possibly with a `shards validate` tool (#29).

waj:
> This change makes **all** shards incompatible with Crystal 1.0

And that's the idea. Shards should be tested and upgraded to work with Crystal 1.0. With so many deprecations and breaking changes I doubt I could trust any shard could just work once we release 1.0.

> Such a huge disruption is completely unacceptable. It breaks the entire ecosystem.

Quite the opposite. It might feel disruptive at first but in the long term it will help to keep a healthy ecosystem. Any shard owner that is still maintaining the code, once Crystal 1.0 is released just need to check for incompatibilities, change the shard.yml and tag a new version. If they were keeping the source up to date with latest changes and pay attention to deprecation notices, it should take just a couple of minutes. And it wont happen that often. Just for any major Crystal release.

> I don't know any dependency managers that make such a restrictive implication about language version compatibility.

Yes, there are examples. Dart also implemented this constraint and it seems is mandatory since 2.0 (https://dart.dev/dart-2). Elm also has very strict constraints.

straight-shoota:
This might work if every shard owner updates publishes their shards immediately. But that's totally unrealistic. Even if every individual release can be done in a few minutes or seconds, the entire process is certainly going to take a long time because shard owners won't respond immediately.
The result would be hundreds of shards that work perfectly fine on 1.0 but `shards` refuses to install them. Applications depending on such shards won't update to use Crystal 1.0.

As the user of a shard, I care that I can install and use that shard. I don't care whether it says it works. As long as it works, I want to be able to use it.
I want shards to do that for me and not make me fork the shard I'm depending on only to adjust the `crystal` property.

RX14:
I'm with @straight-shoota. Can you find an example of another language's package manager which is strict about language versions?

bcardiff:
I see that enforcing an implicit < 1.0.0 is consistent with the future default values for crystal init: `crystal: x.y.z` meaning `crystal: ~> x.y, >= x.y.z`.

The proposed implicit value < 1.0 is consistent with the current status: all shards have been targeted to 0.x version the language and std-lib so far.

Everything that is working today is doing it with a 0.x version of crystal and std-lib. It is understandable that there is a maintenance to make them available in 1.0. When discussing the upper bound constraints implications, we acknowledge that on a major release implies a revisit on published shards. Going from 0.x to 1.0 is the first of those events.

Having a crystal constraint is like having the std-lib as a dependency. How would you know which version of the std-lib you are expecting to use otherwise? When adding a shard dependency, I think it's a bad practice to leave it unconstrained.

I think it encourages a better ecosystem that packages have version constraints, for dependencies and std-lib. I see no reason to treat std-lib different than any other dependency.

I don't like to encourage an ecosystem where the dependencies are so loosly stated that you might not even know which version of the language is a package ready for. 

If we think of Crystal running in the long run, not everything will need to run in the latest version. If we allow optional crystal constraint, that won't work. 

As for other ecosystem examples:

* Elm enforces minor version since they are in 0., that is more strict that what we've done so far and the ecosystem is able to move forward.
* Haskell/cabal uses base to express this kind of dependencies and also enforce upper bounds. https://www.haskell.org/cabal/users-guide/developing-packages.html
* Dart also enforce upper bounds, yet they didn't start from the beginning and they choose <2.0.0 as the implicit value.

---

We might need to revisit how pre release version of the std-lib will be interpreted, to ensure the the ecosystem is able to catch up without too much hassle. Although that story is again, not that different from how a shard can be up to date with head version of dependencies.

jhass:
I'm fine with an implicit `< 1` constraint, if it's overridable by the library consumer.

This could either be a simple flag to `shards install` or a per dependency override to ignore version constraints on the dependency in general, including the Crystal version.

The current proposal will almost certainly cause the Python 2 - 3 schism with each major Crystal release.

RX14:
The problem is that we're gonna release 1.0, end up at the top of HN, and nobody's gonna be able to use it because none of the shards will have updated. Bad experience for everyone.

waj:
Nobody says we're going to release 1.0 overnight without giving time for shards to prepare. There's going to be 1.0 prereleases that everyone will be using to test the compatibility of their shards. It's gonna be bad experience anyway if you rush to install shards and none of them even compile because the authors didn't take the time to pay attention to deprecation notices. BTW, majority of the most used shards already have versions explicitly defined, so the "pain" would already exist. @bcardiff have been always taking the time to test many broadly used shards before each release, sending PRs in many cases to help them keep up to date with so many breaking changes.

We can enable some mechanisms to disable or circumvent dependency checking, like local overrides or force install. That also applies to dependencies between shards. But still we should provide a good path to prepare for 1.0.

I think we could make the parsing of the current Crystal version to ignore the prerelease portion. That way shards doesn't need to be re-published between 1.0-pre and 1.0. In other words, the sequence would be:
  * Crystal 1.0.0-pre1 is published
  * Shards create new releases with `crystal: 1.0.0`. Just like any shard release for 0.x because of incompatibilities.
  * Crystal 1.0.0 is released. All the shards already published for the prerelease are ready to go.

waj:
I just added another commit to relax the version comparison during prerelease as I explained in the previous comment. Before the release we could add options to disable the check on individual shards o globally. Let's see what's the best option for that.

Once 1.0.0-pre1 is released we will be able to experiment with this behaviour. As I said, most of the most famous shards will have to make the update anyway. I don't think this is going to represent ire from the community, but let's see.

src/config.cr
```diff
@@ -75,6 +75,17 @@ module Shards
       end
       raise Error.new("Error executing crystal:\n#{error}") unless status.success?
       output.to_s.strip
+    end)
+  end
+
+  def self.crystal_version(@@crystal_version : String)
```
bcardiff, line 81:
```suggestion
  def self.crystal_version=(@@crystal_version : String)
```

was this supposed to be a setter? and used somewhere?

RX14:
I'm still against this.

asterite:
I have a quedtion: if I mark my shard as compatible with 1.0 and Crystal 2.0 is released, I'd I don't update my shard then it's not eligible for install anymore?

jhass:
> I have a quedtion: if I mark my shard as compatible with 1.0 and Crystal 2.0 is released, if I don't update my shard then it's not eligible for install anymore?

Yes, that's the current proposal and the issue we (@RX14, @straight-shoota  and me) seem to have about this.

asterite:
So my experience with Elm when upgrading to 0.19.0 from 0.18.0 was that a lot of packages refused to install because they didn't say "elm: 0.19.0". But maybe those packages compile without changes? What we ended up doing is creating forks of those packages to be able to use them. The authors were maybe on vacation, maybe didn't have time to look at it anymore (even if the code still worked fine! Code doesn't degrade like fruits). It's pretty annoying, I think.

Specially with a compiled language, if you have a newer Crystal version and you try to use a package that wasn't "supposedly" updated to the newer version, you will immediately find out if the code doesn't compile. If it compiles, then there's a really big chance that it works.

Ruby has been like that for years and it's worked really well.

Also, if I request an author to upgrade their code to Crystal 2.0.0, can they go ahead, set "crystal: 2.0.0", tag, commit and push without even verifying that the code works for Crystal 2.0.0? Then the shard will be compatible according to shards, but it will fail compilation. That means the property is still a "best effort" thing.

On the other hand, if we trust authors to always have time to keep updating the `crystal` property (which is a bit unrealistic, because people take vacations, they have family and other stuff to do) (and also code doesn't rotten like a fruit if you don't update it) then this might be a good compromise. It could mean all shards marked with `crystal: x` are %99.99 likely to work with that crystal version. Then you will almost never find surprises that you install some shards but then the code doesn't compile.

I still don't know which one I prefer. Both have pros and cons.

waj:
@jhass I agree with the bypass, just that I prefer to do that on a separate PR. We have time because this wont have real impact until 1.0

@asterite Elm approach was quite more strict. Maybe too much. This only disables automatic compatibility between major versions. Many times we said Crystal 2.0 might have some breaking changes. Note that this also works the other way around: if someone still uses 1.0, there is no need to do any juggling with the versions to add new shards to the project, because the dependency manager already would be filtering the compatible ones.

Note also, that the restriction can be disabled or relaxed by a shard owner by specifying `crystal: *`, `crystal: > 0` or `crystal: >= 1, <3`. 

All this was already discussed in #365. I just changed the semantics of not having a `crystal` version from `*` to `< 1.0.0` because there are many shards that didn't have the attribute in the past but they have it now. So without this change, as soon as we release 1.0.0-pre1, the older version gets selected instead of just failing because the shards was not updated yet. I can revert this change, but I think it's a small price to pay at the beginning instead of carrying with this heritage for the eternity (yeah, I have big plans for Crystal :o). Majority of well maintained shards will do the update between 1.0.0-pre1 and 1.0.0, so nobody would notice disruption on those shards by the time of release. On the other hand, abandoned shards could be left out of sight in indexers for example if you could optionally filter by Crystal version.

src/config.cr
```diff
@@ -75,6 +75,17 @@ module Shards
...
+  end
+
+  private def self.without_prerelease(version)
+    if version =~ /^(\d+)\.(\d+)\.(\d+)([^\w]\w+)$/
+      "#{$1}.#{$2}.#{$3}"
+    else
+      version
```
Sija, line 88: Why not use `SemanticVersion` for that?

spec/integration/install_spec.cr
```diff
@@ -8,7 +8,7 @@ describe "install" do
     }
 
     with_shard(metadata) do
-      run "shards install"
+      debug "shards install"
```
bcardiff, line 11: debug leftover

```suggestion
      run "shards install"
```

PR #400, opened 2020-06-02
Bump lock version

waj:
This upgrades the `version` in `shards.lock` to `2.0`, because it effectively cannot be used by previous versions (git refs are stored as part of the version metadata).

### DIFF
--- a/src/lock.cr
+++ b/src/lock.cr
@@ -2,7 +2,13 @@ require "./ext/yaml"
 require "./dependency"
 
 module Shards
-  module Lock
+  class Lock
+    property version : String
+    property shards : Array(Dependency)
+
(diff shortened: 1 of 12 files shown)


src/lock.cr
```diff
@@ -15,9 +23,9 @@ module Shards
       pull.read_stream do
         pull.read_document do
           pull.read_mapping do
-            key, value = pull.read_scalar, pull.read_scalar
+            key, version = pull.read_scalar, pull.read_scalar
 
-            unless key == "version" && value == "1.0"
+            unless key == "version" && version.in?("1.0", "2.0")
```
Sija, line 28: I'd extract those into a `VALID_VERSIONS = {"1.0", "2.0"}` constant.

PR #401, opened 2020-06-03
Delete old .sha1 files

waj:
Files `.sha1` left from previous Shards version are now deleted when the `.shards.info` is generated the first time.

### DIFF
--- a/src/info.cr
+++ b/src/info.cr
@@ -17,6 +17,11 @@ class Shards::Info
 
   def save
     Dir.mkdir_p(@install_path)
+
+    unless File.exists?(info_path)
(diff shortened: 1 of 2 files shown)


src/info.cr
```diff
@@ -17,6 +17,11 @@ class Shards::Info
 
   def save
     Dir.mkdir_p(@install_path)
+
+    unless File.exists?(info_path)
+      Dir[File.join @install_path, "*.sha1"].each { |p| File.delete(p) }
```
straight-shoota, line 22: This is incorrect: Glob patterns must always use `/` as separator, not `\` even on Windows.

PR #411, opened 2020-06-17
Disable interactive credential prompt for git resolver

### LINKED ISSUE #316: "shards install" halts asking for github username
```
name: kemal_app
version: 0.1.0

authors:
  - IOM <>

targets:
  kemal_app:
    main: src/app.cr

crystal: 0.32.1

license: MIT

dependencies:
  kemal:
    github: kemalcr/kemal
  spec-kemal:
    github: kemalcr/spec-kemal
    branch: master
  accord:
    github: neovintage/accord
  denetmen:
    github: izniburak/denetmen
  sentry:
    github: samueleaton/sentry
    branch: master
  kemal-session:
    github: kemalcr/kemal-session
  pg:
    github: will/crystal-pg
  kemal-pg:
    github: sdogruyol/kemal-pg
  raven:
    github: Sija/raven
```

```
$ shards install
Fetching https://github.com/kemalcr/kemal.git
Fetching https://github.com/luislavena/radix.git
Fetching https://github.com/jeromegn/kilt.git
Fetching https://github.com/crystal-loot/exception_page.git
Fetching https://github.com/kemalcr/spec-kemal.git
Fetching https://github.com/neovintage/accord.git
Fetching https://github.com/izniburak/denetmen.git
Fetching https://github.com/samueleaton/sentry.git
Fetching https://github.com/kemalcr/kemal-session.git
Fetching https://github.com/sdogruyol/kemal-pg.git
Fetching https://github.com/ysbaddaden/pool.git
Fetching https://github.com/Sija/raven.git
Username for 'https://github.com': 
```
after adding:
```
  raven:
    github: Sija/raven
```

straight-shoota:
Resolves #316

Git repository requests over HTTP can result in 401 Authentication Required status code. This is typically the case when a repository is private. Some git providers also return this status code when a repository exists to avoid disclosing the existence of private repositories.
By default, `git clone` prompts for credentials when it receives a 401 status code. When `shards install` uses `git clone`, this leads to unexpected and misleading behaviour. `shards install` is not meant to be used interactively, so this prompt should be completely disabled.

This PR implements this by using the environment variable `GIT_ASKPASS` set to a value resulting in false. It can still be overriden by setting the environment variable.

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -305,7 +305,7 @@ module Shards
     end
 
     private def mirror_repository
-      run_in_current_folder "git clone --mirror --quiet -- #{Helpers::Path.escape(git_url)} #{local_path}"
+      run_in_current_folder "GIT_ASKPASS=${GIT_ASKPASS:-/usr/bin/test} git clone --mirror --quiet -- #{Helpers::Path.escape(git_url)} #{local_path}"
     rescue Error
       raise Error.new("Failed to clone #{git_url}")
     end
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -305,7 +305,7 @@ module Shards
     end
 
     private def mirror_repository
-      run_in_current_folder "GIT_ASKPASS=${GIT_ASKPASS:-/usr/bin/test} git clone --mirror --quiet -- #{Helpers::Path.escape(git_url)} #{local_path}"
+      run_in_current_folder "git clone -c core.askPass=true --mirror --quiet -- #{Helpers::Path.escape(git_url)} #{local_path}"
     rescue Error
       raise Error.new("Failed to clone #{git_url}")
     end
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -305,6 +305,13 @@ module Shards
     end
 
     private def mirror_repository
+      # The git-config option core.askPass is set to a command that is to be
+      # called when git needs to ask for credentials (for example on a 401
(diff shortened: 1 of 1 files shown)


src/resolvers/git.cr
```diff
@@ -305,7 +305,7 @@ module Shards
     end
 
     private def mirror_repository
-      run_in_current_folder "git clone --mirror --quiet -- #{Helpers::Path.escape(git_url)} #{local_path}"
+      run_in_current_folder "GIT_ASKPASS=${GIT_ASKPASS:-/usr/bin/test} git clone --mirror --quiet -- #{Helpers::Path.escape(git_url)} #{local_path}"
```
bcardiff, line 308: In OSX 

```
$ which test
/bin/test
```

jhass: Does it  need to be an absolute  path or does it do PATH lookup?  Maybe  `${GIT_ASKPASS:-false}` would be most portable.

straight-shoota:
The option doesn't need to be an absolute path. I also changed it to `true` because that's the minimal reduced value and should work in any shell (`test` probably too, but we don't really need a test here).
I also changed from using the environment variable `GIT_ASKPASS` to the git config setting `core.askPass` because that seems cleaner. Overriding is still possible by setting the environment variable because it has precedence over the config setting.

jhass:
Maybe  add a  comment  explaining this both  :)

I read the diff first and was very confused, as you need to know that the setting takes a command rather than a boolean value and that it's still overridable  :)

PR #414, opened 2020-06-23
Add convenient makefile arguments for packaging

bcardiff:
Similar to https://github.com/crystal-lang/crystal/pull/9471
Follow-up to https://github.com/crystal-lang/distribution-scripts/pull/71#discussion_r442992993

I'm only adding release, debug, static to use the same in crystal repo. But I think having FLAGS is enough.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,13 @@
 .POSIX:
 
+release ?=      ## Compile in release mode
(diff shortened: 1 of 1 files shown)


bcardiff:
Good catch @taylorthurlow. The README is indeed outdated regarding the CRFLAGS (since #344).

PR #415, opened 2020-06-23
Allow CRYSTAL env var to determine compiler

### LINKED ISSUE #406: Resolution of crystal dependency is hard tied to `crystal` executable
The implicit crystal dependency introduced in #395 calls `crystal env` to determine the crystal version it should treat as a dependency.
It assumes you're installing dependencies for use with `crystal` binary in current PATH. But that might not at all be what I intend to do.
If I want to build with a compiler that's not `crystal`, `shards` might install wrong dependency versions when it implicitly uses `crystal` as reference. For example, I usually have a couple different compiler executables in my PATH. Currently, the versions are not very different, so it's probably not going to cause many problems. But as soon as we move on after 1.0, it would be quite common to have different compiler versions available, say `crystal-1.0` and `crystal-1.1` to test against the different releases.
Using `--ignore-crystal-version` would completely disable the crystal dependency and thus doesn't reproduce the same behaviour as when using a proper version.

Currently, you would have to change `PATH` to point to the appropriate executable in order for shards to pick the correct version as dependency. That requires much effort and can easily cause issues by itself. We need a better solution.

Ideas that come to mind are to specify either the crystal version directly or the path to the compiler that should be used as configuration option to shards.
A very neat solution would use the `CRYSTAL` enviornment value. That's typically used to make the compiler executable configurable in Makefiles. So it would integrate very well with that.

bcardiff:
Fixes #406

Shards will read CRYSTAL env var, fallback to `crystal`, to determine which crystal compiler to use.

This also enables that a compiler+shards package can force to use the compiler of that same packaging if wanted.

Something that is not handled is how the `executables` + `postinstall` can deal with this override.

A dependency is forced to use in the `postinstall` a `${CRYSTAL:-"crystal"}` instead of `crystal` which is not great.
Another alternative is to encourage `shards build` instead of `crystal build` that way shards will be invoked again and can choose the appropiate crystal compiler since the environment is preserved.
But we actually have the same problem, the `postinstall` assumes that `shards` is on the path. So what? we should have a `${SHARDS:-"shards"}`? Not great.

I think that the scenario of having executable helpers in the bin directory that are compiled by dependencies on postinstall should be more declarative. That would be a way to solve it I think.

### DIFF
--- a/src/script.cr
+++ b/src/script.cr
@@ -46,7 +46,7 @@ module Shards
     end
 
     private def self.prepend_path(prefix : String, suffix : String?)
-      suffix ? "#{prefix}:#{suffix}" : prefix
+      suffix ? "#{prefix}#{Process::PATH_DELIMITER}#{suffix}" : prefix
     end
   end
 end
--- a/src/script.cr
+++ b/src/script.cr
@@ -25,7 +25,7 @@ module Shards
             Dir.delete bin_path
           end
           env_path = prepend_path(bin_path, ENV["PATH"]?)
-          Log.debug { "Creating shards script environment created at #{bin_path}. PATH=#{env_path}" }
+          Log.debug { "shards script environment created at #{bin_path}. Updated PATH=#{env_path}" }
 
           add_executable bin_path, "shards", Process.executable_path
           add_executable bin_path, "crystal", Process.find_executable(Shards.crystal_bin)
--- a/src/script.cr
+++ b/src/script.cr
@@ -6,9 +6,47 @@ module Shards
     def self.run(path, command, script_name, dependency_name)
       Dir.cd(path) do
         output = IO::Memory.new
-        status = Process.run("/bin/sh", input: IO::Memory.new(command), output: output, error: output)
+        status = Process.run("/bin/sh", env: Script.environment, input: IO::Memory.new(command), output: output, error: output)
         raise Error.new("Failed #{script_name} of #{dependency_name} on #{command}:\n#{output.to_s.rstrip}") unless status.success?
       end
+    end
+
+    @@environment : Process::Env = nil
+
+    def self.environment : Process::Env
+      @@environment ||=
+        begin
+          bin_path = File.tempname(".shards-env")
+          Dir.mkdir(bin_path)
+          at_exit do
+            Dir.new(bin_path).each_child do |e|
+              File.delete(File.join(bin_path, e))
+            end
+            Dir.delete bin_path
+          end
+          env_path = prepend_path(bin_path, ENV["PATH"]?)
+          Log.debug { "Creating shards script environment created at #{bin_path}. PATH=#{env_path}" }
+
+          add_executable bin_path, "shards", Process.executable_path
+          add_executable bin_path, "crystal", Process.find_executable(Shards.crystal_bin)
+
+          {"PATH" => env_path}
+        rescue e
+          Log.error(exception: e) { "Unable to create shards script environment" }
+
+          {} of String => String
+        end
+    end
+
+    private def self.add_executable(bin_path : String, name : String, original_path : String?)
+      if original_path
+        Log.debug { "Adding #{name}=#{original_path} to shards script environment" }
+        File.symlink(original_path, File.join(bin_path, name))
+      end
+    end
+
+    private def self.prepend_path(prefix : String, suffix : String?)
+      suffix ? "#{prefix}:#{suffix}" : prefix
     end
   end
 end
--- a/spec/integration/spec_helper.cr
+++ b/spec/integration/spec_helper.cr
@@ -90,7 +90,7 @@ private def setup_repositories
   create_git_release "transitive", "0.2.0", {
     dependencies: {version: {git: git_path(:version)}},
     scripts:      {
-      postinstall: "crystal build src/version.cr",
+      postinstall: %(${CRYSTAL:-"crystal"} build src/version.cr),
     },
   }
 
--- a/spec/integration/spec_helper.cr
+++ b/spec/integration/spec_helper.cr
@@ -90,7 +90,7 @@ private def setup_repositories
   create_git_release "transitive", "0.2.0", {
     dependencies: {version: {git: git_path(:version)}},
     scripts:      {
-      postinstall: %(${CRYSTAL:-"crystal"} build src/version.cr),
+      postinstall: "crystal build src/version.cr",
     },
   }
 
--- a/spec/support/factories.cr
+++ b/spec/support/factories.cr
@@ -121,7 +121,7 @@ end
 
 def run(command, *, env = nil)
   cmd_env = {
-    "CRYSTAL_PATH" => "#{Shards::INSTALL_DIR}:#{`crystal env CRYSTAL_PATH`.chomp}",
+    "CRYSTAL_PATH" => "#{Shards::INSTALL_DIR}:#{`#{Shards.crystal_bin} env CRYSTAL_PATH`.chomp}",
   }
   cmd_env.merge!(env) if env
   output, error = IO::Memory.new, IO::Memory.new
--- a/src/commands/build.cr
+++ b/src/commands/build.cr
@@ -31,10 +31,11 @@ module Shards
           target.main,
         ]
         options.each { |option| args << option }
-        Log.debug { "crystal #{args.join(' ')}" }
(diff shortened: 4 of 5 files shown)


jhass:
Can't we just figure out the directory `$CRYSTAL` lives in and prepend it to the `$PATH` before invoking `postinstall`? Or if we want to avoid whatever else might be in that directory to appear in `$PATH`, we could symlink `$CRYSTAL` to `/tmp/something/crystal` and push `/tmp/something` into `$PATH`.

bcardiff:
I think that prepending current shards _and_ crystal location to the path when invoking scripts from shards is acceptable.

jhass:
I just realized, the symlink approach could have another advantage: it removes the need that `$CRYSTAL` points to something called `crystal`. It is not uncommon to deal with parallel installation of multiple versions of something by renaming some of the executables, so you would have for example `/usr/bin/crystal` and `/usr/bin/crystal-1.2`.

bcardiff:
Good point. I usually have directories like `crystal-x.y/bin/crystal`. So I didn't consider it.

Should shards create & prepend the directory to the path only when needed? Identifying if the resolution to shards and crystal leads to the same executable that is running and to the `Shards.crystal_bin` location?

Those `/usr/bin/crystal-1.2` that you and @straight-shoota seems to have are those symlinks? Or there are custom made wrapper scripts?

jhass:
> Should shards create & prepend the directory to the path only when needed? Identifying if the resolution to shards and crystal leads to the same executable that is running and to the Shards.crystal_bin location?

I think it's fine to start simple and make it smart afterwards. That is I'd be totally fine if any `shards install` invocation sets up the temp directory & symlink early and only cleans it up on process termination. Then iterative optimizations upon that can be to
* lazily setup only once the first `postinstall` is hit
* skip  the setup if currently running `shards` comes first in current `$PATH` and current `$CRYSTAL` already comes first in path and both are named `shards` and `crystal` respectively.

> Those /usr/bin/crystal-1.2 that you and @straight-shoota seems to have are those symlinks? Or there are custom made wrapper scripts?

I don't actually have that anywhere right now. It's just common to see from packages of other things, especially languages. I know for certain that Archlinux, Debian and Fedora/RHEL families of distributions install `/usr/bin/ruby-2.6` or `/usr/bin/ruby2.6` if you install a ruby2.6 package while the ruby package is say 2.7. Also it's basically ubiquitous to have `/usr/bin/python`, `/usr/bin/python2` and `/usr/bin/python3`. Whether things are symlinks into `/opt/`, wrapper scripts that setup some environment variables (hello `virtualenv`, `rbenv`) or just executables with different standard paths compiled in shouldn't matter. Crystal would support all three  modes of operation right now already, we should only make sure to preserve such here by simply making no assumptions about it. Symlinking a symlink is entirely fine.

bcardiff:
Ok, will take the route of lazily setup only once the first postinstall is hit for now.

Something that might not work is if `crystal-x.y`/`shards-x.y` is an alias I think. They will not be found in path to append. `$CRYSTAL` will be forwarded but `shards` will not resolve to `shards-x.y`. This is out of my comfort zone 🙈 . It's fine if aliases are not supported, that why I wanted to check how the `crystal-x.y` was created (yet from the `/usr/bin` prefix it was obvious they are not aliases)

jhass:
well yes, aliases only exist in the current shell anyhow afaik, it won't even work to put them into `$CRYSTAL`.

src/script.cr
```diff
@@ -6,9 +6,47 @@ module Shards
...
+          bin_path = File.tempname(".shards-env")
+          Dir.mkdir(bin_path)
+          at_exit do
+            Dir.new(bin_path).each_child do |e|
+              File.delete(File.join(bin_path, e))
+            end
+            Dir.delete bin_path
+          end
+          env_path = prepend_path(bin_path, ENV["PATH"]?)
+          Log.debug { "Creating shards script environment created at #{bin_path}. PATH=#{env_path}" }
```
jhass, line 28:
```suggestion
          Log.debug { "shards script environment created at #{bin_path}. Updated PATH=#{env_path}" }
```

src/script.cr
```diff
@@ -6,9 +6,47 @@ module Shards
...
+
+    private def self.add_executable(bin_path : String, name : String, original_path : String?)
+      if original_path
+        Log.debug { "Adding #{name}=#{original_path} to shards script environment" }
+        File.symlink(original_path, File.join(bin_path, name))
+      end
+    end
+
+    private def self.prepend_path(prefix : String, suffix : String?)
+      suffix ? "#{prefix}:#{suffix}" : prefix
```
jhass, line 49:
```suggestion
      suffix ? "#{prefix}#{Process::PATH_DELIMITER}#{suffix}" : prefix
```

straight-shoota:
> Well yes, aliases only exist in the current shell

They do work in subshells when configured via shellrc.

My two cents: I don't like where this is going. It adds even more build tool features to shards. Shards shouldn't be a build tool. It's a dependency manager. There are lots of build tools out there and most of them are better than shards for this job.

bcardiff:
@straight-shoota how would you solve that the environment used is preserved along the way? Even if we assume that crystal is in the path, how we ensure the same shards is called in the postinstall?

bcardiff:
Plus if CRYSTAL env var is used to choose the crystal compiler to grab the version for the resolution, how we can ensure that that binary is used as the compiler all along?

straight-shoota:
> how we ensure the same shards is called in the postinstall?

I'd let that be the responsibility of a real build tool.

jhass:
So you would remove `postinstall`? Fact is that while people can just `crystal build` in a `postinstall` script, they will. I agree we shouldn't  overboard, but I view this as the minimum effort we can make to not make that break in hilarious and hard to debug ways.

waj:
I don't think having a `postinstall` hook makes Shards a build tool. I don't think it's just a dependency manager either. It's the tool that allows building, managing and (someday)  publishing shards. Ideally many projects would just specify `postinstall: make`.

To me the noise is coming from the relationship with the `targets` spec. That is totally expendable and it would remove the need of exposing the `shards` binary in any way to the postinstall step. But I think the idea was always give an option to not require the users to deal with makefiles, bringing a default behaviour to call the Crystal compiler. 

With the risk of turning this conversation away from the topic of this PR, maybe we could extend the `postinstall` syntax to integrate it better with `targets` and thus not requiring to specify a command that invokes `shards`. For example this could indicate that the target `foo` must be built after the shard is installed:

```yaml
postinstall:
  target: foo
```

I may even suggest that the `postinstall` could be (optionally) an array, so more than one target or command can be specified. Some might be thinking "Oh no... MORE features", but this would provide more cohesion within the configuration.

If this message receive at least one thumbs up, I'll open a RFC so we could discuss the feature there 🙂

bcardiff:
I will give the 👍  because I mention something like that in https://github.com/crystal-lang/shards/pull/415#issue-438750245

> I think that the scenario of having executable helpers in the bin directory that are compiled by dependencies on postinstall should be more declarative. That would be a way to solve it I think.

Yet, that will not solve the issue that if a makefile is used either shards or crystal might be called there.

bcardiff:
It seems that whatever solution we try to achieve here will lead to half-way implementation. Although #406 seems legit at first, shards is already bound to crystal binary in reality since post_install steps might call crystal compiler.

It was a nice excercise though :-) 

For now it seems that the best things is to close this PR and maybe #406.

jhass:
I didn't get what the  outstanding issue with this approach is.  Also let's not get the aim for a perfect solution in the way of useful iteration (that theme seems to expose in a couple of other places as well...). Once we find it, we still can move towards the perfect solution, improving status quo is not  useless in the meantime.

bcardiff:
Part of the line of thought was

#406 proposed another way to determine which crystal version is wanted. Choosing the binary push the story of how to honor that choice within the postinstall steps.

If the postinstall where more declarative, not requiring to explicitly call crystal in a shell script there will be no need to set the PATH since shards could do that per target directly.

But users might still prefer to use make or other build tools and call crystal.

Maybe in the future we can have a standardize an env var like CC to choose the crystal compiler, that would remove the need for changing the PATH with temporal directories. It will also allow compilers named crystal-x.y to be used. Shards could set a fallback to that env var.

Part of my motivations here was to pave the road for a self-containted .tar.gz where the crystal and shards binary there will call each other. This will be better handle by wrapper scripts of crystal and shards that will prepend the ./bin location relative to the wrapper scripts to PATH.

PR #416, opened 2020-06-24
Fix spec for running under 32-bit Linux

lugia-kun:
* 1900/01/01 is not a representable date on 32-bit Linux and restricted to the oldest date which is representable.

See crystal-lang/crystal#9542 for more info.

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -419,7 +419,7 @@ describe "install" do
 
     with_shard(metadata) do
       run "shards install"
-      File.touch "shard.lock", Time.utc(1900, 1, 1)
+      File.touch "shard.lock", Time.utc(1901, 12, 14)
       mtime = File.info("shard.lock").modification_time
       run "shards install"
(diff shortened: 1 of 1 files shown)


spec/integration/install_spec.cr
```diff
@@ -419,7 +419,7 @@ describe "install" do
 
     with_shard(metadata) do
       run "shards install"
-      File.touch "shard.lock", Time.utc(1900, 1, 1)
+      File.touch "shard.lock", Time.utc(1901, 12, 14)
```
asterite, line 422: I wonder why such an old time is needed here 🤔 

Can't we just use a date from last year or something like that?

jhass: Something pre 1970 makes sense I guess in case somebody  boots their system without a hardware clock and it starts from `0` :D

Sija: Good luck with doing whatever on such machine (think SSL certificates with specific timeframe) ;)

PR #418, opened 2020-06-25
Add --ignore-crystal-version related suggestion and warnings to guide user

bcardiff:


src/commands/command.cr
```diff
@@ -59,8 +59,12 @@ module Shards
     def handle_resolver_errors
       yield
     rescue e : Molinillo::ResolverError
+      if e.is_a?(Molinillo::VersionConflict) && e.conflicts.has_key?(CrystalResolver.key)
+        suggestion = ", try with --ignore-crystal-version or update incompatible shards."
```
straight-shoota, line 63: I'd reverse the suggestions.  Updating incompatible shards should be the first cause of action. If that does not work, you can resort to ignore.

asterite: I also feel like users will be "Huh, that didn't work. Okay, let's try that flag" without understanding what's going on. Eventually, people will just use that flag blindly because it works better than not using that flag. So I think there should be a longer explanation here.

oprypin: > Eventually, people will just use that flag blindly because it works better than not using that flag

Well yes, that is exactly what happens when the default is strictly worse. Not much to do with the wording. For best results I'll be recommending `alias shards='shards --ignore-crystal-version'` :)

asterite: I think `--ignore-crystal-version` should be removed.

The thing is, once we release 1.0.0 there will surely be broken shards. If we pretend they will work with 1.0.0 then users will be upset too. Run `shards`, everything installs fine. Then you run the app and, opps, `undefined method JSON.mapping`.

By making sure a shard is upgraded to be compatible with 1.0.0 that problem is solved. If the shard is not upgraded, the error message is clearer: there's no such shard compatible with 1.0.0.

bcardiff: As with local overrides, is a mean to avoid blocking the user and be forced to use manual checkouts in directories.

The `--ignore-crystal-version` was introduced a way to avoid blocking between major versions. Maybe it can go away after 1.0 if the local overrides give a more fine grained control.

asterite: The thing is that if users have a way to ignore the version checks, they will. Then `--ignore-crystal-version` is exactly the same as not having the feature at all.

And users **will** do that, because when Crystal 1.0.0 almost all shards won't be compatible with it.

Sija: You better have some good marketing goin' along on with the release, because this will put a wave through the ecosystem. Warning by default instead of blocking would be _a bit_ more gentle for all of the shard authors. Just sayin'

oprypin: Actually yeah, making it a warning would solve almost all the supposed issues with almost none of the downsides

bcardiff:
I've added a couple of warnings to help the user

1. know that --ignore-crystal-version might be useful
2. advertise clearly which shards are might be incompatible due to a mismatch with crystal version restriction
3. advertise if --ignore-crystal-version is not actually needed

src/commands/command.cr
bcardiff: We can't warn by default. If we go that route we won't be able to make it an error later. Otherwise, a future shard would not be able to install the lock file a previous one did.

PR #422, opened 2020-07-22
Shards overrides

### LINKED ISSUE #105: Local overrides for shards development
When I am developing multiple shards and they have dependencies between them, like crystal-db and crystal-sqlite I can't specify that, during development, in my local machine the dependencies should be found in another way than the usual.

I would like to use a `path:` in crystal-sqlite but leave with the `github:` dependency in shard.yml for public usage.

I would like to be able to override where the dependencies should be looked in my environment. This is similar to http://bundler.io/v1.5/git.html allows me to do.

In order to allow this behavior per project basis I propose allowing shards to look up for shard.yml overrides in shard.local.yml that should not be checked in.

```
# crystal-sqlite/shard.yml
dependencies:
  db:
    github: bcardiff/crystal-db
```

```
# crystal-sqlite/shard.local.yml
dependencies:
  db:
    path: path/to/working-copy-of/crystal-db
```

Additionally we could expand shard.local.yml when creating a lib and adding it to the default .gitignore file.

### LINKED ISSUE #299: Allow overriding dependencies
It would be helpful to be able to manually override a dependency, especially with how quickly Crystal changes.

Let's say for example that I have a dependency called `taco_shard`  and it  relies  on another dependency  called `cheese_shard`. Everything works but then a new version of Crystal makes a change that breaks `cheese_shard`. `cheese_shard` is no longer maintained so you fork it to make changes to make it work with the new Crystal. The problem is that you don't have a way to easily override that dependency without *also* forking `taco_shard`. It would be *awesome* to be able to just add something to `shard.yml` that overrides the version/resolver:

```yaml
cheese_shard:
  github: myfork/cheese_shard
  override: true # say that it is ok to override what `taco_shard` specified
```

So even if `taco_shard` says it should use `original/cheese_shard` my `shard.yml` would override that and use the specific  fork. This could also work for versions.

## Prior art

* `override` option in Elixir mix: https://hexdocs.pm/mix/Mix.Tasks.Deps.html#module-dependency-definition-options (search for override)

### LINKED ISSUE #412: RFC: Shards overrides
Related to #105, #240, and #299

## Motivation

There are stories that would benefit from forcing the resolution for a given shard.

* A developer of a framework might prefer to use local working copies for the different shards
* A user willing to use a shard version despite published constraints (it might just work and dependency `shard.yml` was not updated yet)
* A user fixing a dependency used in a framework or application 
* A user checking app against development version of a dependency

### Use local working copies

This is currently possible but requires the user to manually do symbolic links inside the `lib` folder. Yet shards provides a way to have `path:` dependencies. The user should be able to state that in the current machine a specific path can be used as the path source instead of the git one.

But we don't want to put these changes in `shard.yml` since it will pollute with something specific for the development environment.

### Use a shard version despite published constraints

Maybe a shard `foo` was updated with a fix or feature we need in the `v0.10.0` release. But another shard `bar` has a conservative constraint to `foo` of `~> 0.9.2`. If an application requires both `foo` and `bar` the latest release `v0.10.0` of `foo` is not a candidate.

This scenario currently push the user to fork `bar`. If there would be a need to actually fix `bar`, then it will be ok. But it might happen that the only need is to update the version constraint in it.

It would be desired to tweak and force the resolution of `foo` to a given version, despite the constraints of the rest of the dependencies. This only applies in the top level resolution. It is not desired for a dependency to force a version of another one.
### Fixing a dependency

Frameworks have a medium graph size of dependencies. A bug might appear at any level.

In the following scenario with conservative constraints, if we find a bug on `bar` we will be forced to fork `foo` just to change the `bar` restriction.

```
framework
  shard.yml
    dependencies:
      foo:
        github: ...
        version: ~> 0.8.0

foo
  shard.yml
    dependencies:
      bar:
        github: ...
        version ~> 0.5.0
```

If I'm developing / contributing to `framework` it becomes a burden to fork `foo` in order to fix `bar`.

Some specs, specially on framework, will expand crystal projects and call the compiler within the test suite. Currently, if I’m trying to fix `bar`, I need to deal not only with forking `foo`, but to update the template used in the test suite.

### Checking app against development version of a dependency

There is no easy way to keep a CI checking if the upcoming version of a dependency will work. Having an alternate `shard.yml` requires file renaming and also to deal with some complications mentioned in other scenarios.

At the moment, the way to accomplish this is for the shard to follow something like git-flow and use version restriction only in published releases, leaving branch master restriction on development. It's error prone and a burden that the community seems to prefer not to deal with.
## Proposal

Allow shards to use as input not only `shard.yml` file but also 

1. a `shard.override.yml` in the same directory, and
2. a user-wide `shard.override.yml` located in `~/.crystal/shards.override.yml`, XDG_CONFIG_DIRS or similar directory.

Allow shards to accept `-f <file>` option to add additional override files into consideration.

Override files will have the same format as `shard.yml`, yet only `dependencies:` will be used. The semantic would be as follow:

* An override file will never **add** a dependency
* Each time a dependency is added to the resolution graph, if an override with the same name exists, then the specs in the override are used.
* All override files are interpreted as one
* Explicit `-f <file>` overrides the local override file directory 
* Local override overrides the user-wide one

If an override was used along the way, the `shard.lock` file will contain a warning comment indicating that it was generated with input information of specific files. This will help to prevent checking `shard.lock` files with local information, or at worst, recognizing them once checked in.
## Usage

* If I am a developer of crystal-db and crystal-mysql I could have:

```
# ~/.crystal/shards.override.yml
dependencies:
  db:
    path: ~/path/to/crystal-db
```

Doing a `shards install` in crystal-mysql will create a symlink.

* If I need to deal with a restriction along the way, changing it, using a branch, or a different fork I could set that in the project’s override file.

```
# shards.override.yml
dependencies:
  foo:
    github: non-main-author/foo
```

Whether I need to share that with others I can track `shards.override.yml` or not.

* If I need to deal with a fix on a framework or shard that expands crystal projects in their specs I can use the user-wide override file.

* If I want to check in CI against upstream I can have a `shard.upstream.yml` checked in and use `shards install -f shard.upstream.yml`. Once dependencies are installed crystal spec will work. If those specs expand crystal projects, I would need first to move shard.upstream.yml to a user-wide location.
## Other considerations

I consider allowing the main shard.yml file to have overrides like

```
# shard.yml 
dependencies:
  foo:
    github: non-main-author/foo
    forced: true
  bar:
    github: bar/bar
```

But introducing that will be backward incompatible, requiring to add a shard_version marker. Because if bar depends on foo as `github: original-main-author/foo` then there will be conflicts between which fork to use.

There is no reason to have a forced dependency on a library `shard.yml`. For development the prior proposal seems enough.

The only scenario where a forced dependency should be tracked is on application `shard.yml`. In this case, dealing with addition `shard.override.yml` seems like a reasonable compromise.

If we want overrides to be in the `shard.yml` we would need an `overrides:` or `resolutions:` sibling to `dependencies:`. 

```
# shard.yml 
dependencies:
  bar:
    github: bar/bar
resolutions:
  foo:
    github: non-main-author/foo
```

This is similar to what is done in [yarn](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-how-to-use-it). It would offer a slightly cleaner solution to the scenario where a team needs to share overrides. Here, there would be no need to add a warning in the `shard.lock` file. If wanted, this extension can be introduced later.

bcardiff:
Implements most of #412 .

Reads a `shard.override.yml` (or the file specified by `SHARDS_OVERRIDE`) to override the source and restriction of a dependency. There are no changes regarding CLI options of shards.

This allows the cases described in the motivation of #412

* Use local working copies
* Use a shard version despite published constraints
* Fixing a dependency
* Checking app against development version of a dependency

---

### Usage sample

On a project that uses `github:crystal-lang/crystal-sqlite3` it will current use `crysta-db 0.8.0`.

```yaml
# file: shard.yml
name: sample
version: 0.1.0

dependencies:
  sqlite3:
    github: crystal-lang/crystal-sqlite3
    version: ~> 0.15.0

crystal: 0.35.1
```

```yaml
# file: shard.lock
version: 2.0
shards:
  db:
    git: https://github.com/crystal-lang/crystal-db.git
    version: 0.8.0

  sqlite3:
    git: https://github.com/crystal-lang/crystal-sqlite3.git
    version: 0.15.0
```

Let's suppose I need to work on a fix for `crystal-db`. I can create a `shard.override.yml` file pointing to my working copy of `crystal-db`

```yaml
# file: shard.override.yml
dependencies:
  db:
    path: /my/path/to/crystal-db
```

Execute `$ shards update` to change the installed and locked version of `crystal-db`.
(Note: if `$ shards install` is used instead of update, the use will get an error regarding 
that db has ambiguous sources due to the existing shard.lock and the shard.override.yml)

```terminal-session
$ shards update
Resolving dependencies
Fetching https://github.com/crystal-lang/crystal-sqlite3.git
Installing db (0.9.0 at /my/path/to/crystal-db)
Using sqlite3 (0.15.0)
Writing shard.lock
```

After this update you will have a symlink to the /my/path/to/crystal-db

```terminal-session
$ ls -la lib/
total 8
drwxr-xr-x   5 bcardiff  staff  160 Jul 22 11:15 .
drwxr-xr-x  15 bcardiff  staff  480 Jul 22 11:14 ..
-rw-r--r--   1 bcardiff  staff  196 Jul 22 11:15 .shards.info
lrwxr-xr-x   1 bcardiff  staff   43 Jul 22 11:15 db -> /my/path/to/crystal-db
drwxr-xr-x  12 bcardiff  staff  384 Jul 22 11:14 sqlite3
```

And the `shard.lock` has some comments to clarify if the lock was computed with overrides and will not be safe to commit.

```yaml
# file: shard.lock
# WARNING: This lockfile was generated using also shard.override.yml
version: 2.0
shards:
  db: # Overridden
    path: /my/path/to/crystal-db
    version: 0.9.0

  sqlite3:
    git: https://github.com/crystal-lang/crystal-sqlite3.git
    version: 0.15.0
```

Note that in this example we also bump `crystal-db` from 0.8 to 0.9 despite the fact that `crystal-sqlite3` requires `~> 0.8.0`. So we are overriding the requirements along the dependency graph.

You can override to a specific version, branch, fork, local path and solve ambiguos reference in case the ecosystem do not agree what is the main fork for a shard name.

Having a `shard.master.yml` will all dependencies to the master/develop branch will simplify how to check if you the project is up to date with non-released changes of dependencies.

---

Closes #412
Closes #105
Closes #299

### DIFF
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -56,7 +56,7 @@ module Shards
              else
                @spec.dependencies
              end
-      deps = deps.map { |dep| on_override(dep) || dep }
+      deps = apply_overrides(deps)
 
       base = Molinillo::DependencyGraph(Dependency, Dependency).new
       if locks = @locks
@@ -138,7 +138,6 @@ module Shards
     @specs = Hash({String, Version}, Spec).new
 
     def search_for(dependency : R) : Array(S)
-      dependency = on_override(dependency) || dependency
       resolver = dependency.resolver
       check_single_resolver_by_name resolver
 
@@ -160,6 +159,10 @@ module Shards
       @override.try(&.dependencies.find { |o| name_for(o) == name_for(dependency) })
     end
 
+    def apply_overrides(deps : Array(Dependency))
+      deps.map { |dep| on_override(dep) || dep }
+    end
+
     def name_for_explicit_dependency_source
       SPEC_FILENAME
     end
@@ -169,11 +172,11 @@ module Shards
     end
 
     def dependencies_for(specification : S) : Array(R)
-      return specification.dependencies if specification.name == "crystal"
-      return specification.dependencies if @ignore_crystal_version
+      return apply_overrides(specification.dependencies) if specification.name == "crystal"
+      return apply_overrides(specification.dependencies) if @ignore_crystal_version
 
       crystal_dependency = Dependency.new("crystal", CrystalResolver::INSTANCE, MolinilloSolver.crystal_version_req(specification))
-      specification.dependencies + [crystal_dependency]
+      apply_overrides(specification.dependencies) + [crystal_dependency]
     end
 
     def self.crystal_version_req(specification : Shards::Spec)
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -156,7 +156,7 @@ module Shards
     end
 
     def on_override(dependency : Dependency | Shards::Spec) : Dependency?
-      @override.try(&.dependencies.find { |o| name_for(o) == name_for(dependency) })
+      @override.try(&.dependencies.find { |o| o.name == dependency.name })
     end
 
     def apply_overrides(deps : Array(Dependency))
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -44,6 +44,7 @@ module Shards
         spec = dep.resolver.spec(lock_version)
 
         spec.dependencies.each do |dep|
+          dep = on_override(dep) || dep
           add_lock(base, lock_index, dep)
         end
       end
@@ -55,6 +56,7 @@ module Shards
              else
                @spec.dependencies
              end
+      deps = deps.map { |dep| on_override(dep) || dep }
 
       base = Molinillo::DependencyGraph(Dependency, Dependency).new
       if locks = @locks
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -172,11 +172,13 @@ module Shards
     end
 
     def dependencies_for(specification : S) : Array(R)
-      return apply_overrides(specification.dependencies) if specification.name == "crystal"
-      return apply_overrides(specification.dependencies) if @ignore_crystal_version
+      spec_dependencies = apply_overrides(specification.dependencies)
+
+      return spec_dependencies if specification.name == "crystal"
+      return spec_dependencies if @ignore_crystal_version
 
       crystal_dependency = Dependency.new("crystal", CrystalResolver::INSTANCE, MolinilloSolver.crystal_version_req(specification))
-      apply_overrides(specification.dependencies) + [crystal_dependency]
+      spec_dependencies + [crystal_dependency]
     end
 
     def self.crystal_version_req(specification : Shards::Spec)
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -208,6 +208,8 @@ module Shards
     private def versions_for(dependency, resolver) : Array(Version)
       check_single_resolver_by_name resolver
 
+      return resolver.versions_for(Any) if resolver.is_override
+
       matching = resolver.versions_for(dependency.requirement)
 
       if (locks = @locks) &&
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -11,7 +11,7 @@ module Shards
     include Molinillo::SpecificationProvider(Shards::Dependency, Shards::Spec)
     include Molinillo::UI
 
-    def initialize(@spec : Spec, *, prereleases = false, ignore_crystal_version = false)
+    def initialize(@spec : Spec, @override : Override? = nil, *, prereleases = false, ignore_crystal_version = false)
       @prereleases = prereleases
       @ignore_crystal_version = ignore_crystal_version
     end
@@ -136,10 +136,12 @@ module Shards
     @specs = Hash({String, Version}, Spec).new
 
     def search_for(dependency : R) : Array(S)
-      check_single_resolver_by_name dependency.resolver
+      resolver = on_override(dependency, &.resolver)
+      resolver ||= dependency.resolver
+
+      check_single_resolver_by_name resolver
 
       @search_results[{dependency.name, dependency.requirement}] ||= begin
-        resolver = dependency.resolver
         versions = Versions.sort(versions_for(dependency, resolver)).reverse
         result = versions.map do |version|
           @specs[{dependency.name, version}] ||= begin
@@ -151,6 +153,12 @@ module Shards
 
         result
       end
+    end
+
+    def on_override(dependency : Dependency)
+      o = @override.try(&.dependencies.find { |o| name_for(o) == name_for(dependency) })
+
+      yield o if o
     end
 
     def name_for_explicit_dependency_source
@@ -185,6 +193,8 @@ module Shards
     end
 
     def requirement_satisfied_by?(dependency, activated, spec)
+      on_override(dependency) { return true }
+
       unless @prereleases
         if !spec.version.has_metadata? && spec.version.prerelease? && !dependency.prerelease?
           vertex = activated.vertex_named(spec.name)
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -98,7 +98,7 @@ module Shards
         resolver = spec.resolver || raise "BUG: returned Spec has no resolver"
         version = spec.version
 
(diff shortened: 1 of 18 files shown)


jhass:
Should we really touch the regular `lock` opposed to creating a `shard.override.lock` or not writing down the overrides at all?

If I have to take care to not commit the dirty lock, why have a separate file for overrides in the first place? It could just be section in `shard.yml` then, there's no difference to partially committing one or two files.

OTOH what's wrong with committing the overrides file and the lock changes from that?  If collaborating on a project with outdated dependencies (the primary usecase for this), why should everybody have to redo the  overrides?

Sija:
This whole flow with additional file seems wrong to me. I'd rather take inspiration from Elixir (already mentioned) instead, KISS plz.

bcardiff:
@jhass in theory there is nothing wrong with committing the override and lock with overrides. That should be done if the overrides are required for an app to work. It happens that I have more present the workflow of temporal overrides to fix things.

If the story for sharing overrides, or forcing some resolution without warnings is wanted the last proposal of the proposal in #412 can be implemented later:

> The only scenario where a forced dependency should be tracked is on application `shard.yml`. In this case, dealing with addition `shard.override.yml` seems like a reasonable compromise.
> 
> If we want overrides to be in the `shard.yml` we would need an `overrides:` or `resolutions:` sibling to `dependencies:`. 
> 
> ```
> # shard.yml 
> dependencies:
>   bar:
>     github: bar/bar
> resolutions:
>   foo:
>     github: non-main-author/foo
> ```
> 
> This is similar to what is done in [yarn](https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-how-to-use-it). It would offer a slightly cleaner solution to the scenario where a team needs to share overrides. Here, there would be no need to add a warning in the `shard.lock` file. If wanted, this extension can be introduced later. 

I think that having another lock file will generate confusion. The lock can be used as the source of truth for reinstalling the dependencies, if there are two, then there is a logic need to choose.

@Sija in #412 I mention that adding a forced or override attribute similar to Elixir one is not backward compatible and we will be forced to add a version in the shard.yml. The alternative is to override as in Yarn which I mention. Having an additional files allows per-user customization similar to Bundler which I've found convenient to work with custom forks and local paths.

src/molinillo_solver.cr
```diff
@@ -153,6 +155,14 @@ module Shards
       end
     end
 
+    def on_override(dependency : Dependency | Shards::Spec) : Dependency?
+      @override.try(&.dependencies.find { |o| name_for(o) == name_for(dependency) })
```
waj, line 159: `name_for` is just a callback for Molinillo. I wouldn't use it here.
```suggestion
      @override.try(&.dependencies.find { |o| o.name == dependency.name })
```

src/molinillo_solver.cr
```diff
@@ -162,11 +172,11 @@ module Shards
     end
 
     def dependencies_for(specification : S) : Array(R)
-      return specification.dependencies if specification.name == "crystal"
-      return specification.dependencies if @ignore_crystal_version
+      return apply_overrides(specification.dependencies) if specification.name == "crystal"
```
waj, line 175: I'd extract `dependencies = apply_overrides(specification.dependencies)` to avoid repeating the same call

mloughran:
This is a very useful feature but it's currently not easy to find via google – I arrived here by following a trail of issues.

This may be related to the [shards command](https://crystal-lang.org/reference/1.4/the_shards_command/index.html) page being outdated and not linked to the [man page](https://man.archlinux.org/man/community/shards/shards.1.en)? /cc https://github.com/crystal-lang/crystal-book/issues/476

Issue #398, opened 2020-06-02
Issue with clearing cache


jwoertink:
I have a private shard that my app uses. Right now I just point it to master branch. 
When I do these steps, I seem to run in to a cache issue.
* Make update to shard
* Push master to github
* run `shards update` from project
* boot project and get error because it never actually updated the shard
* `rm -rf lib/my_shard`
* `shards install`
* boot project and now it's all good.

```
$ shards --version
Shards 0.10.0 (2020-04-06)
$ crystal -v
Crystal 0.34.0 (2020-04-07)

LLVM: 10.0.0
Default target: x86_64-apple-macosx
```

This is how I have my shard defined.

```
dependencies:
  my_shard:
    git: https://jwoertink:abc@github.com/group/my_shard.cr
    branch: master
```

jwoertink:
@waj I will try with the latest version to make sure and then report back. If it works fine, I'll close this out. Thanks!

jwoertink:
Ok, just ran in to this again, but maybe a slightly different issue. I installed my shards, then I created an override and ran my `shards update` and things were good. Next I wanted to go back, so I removed `shard.lock` `lib/` `~/.cache/crystal`, and `shard.override.yml`. Then I `shards install` and I was back to original. Finally, I wanted to get back to the override one more time. I delete all those files, add my shard override, and run `shards install` and it still had the original version even though the lock showed the override. The `lib` still contained the wrong one. To fix this, I went in to the lib and removed just that one shard, and ran `shards update` and it was all good..

Seems like some combo of me running `install` vs `update`. Even though the lock file is being updated properly, the `lib` directory isn't.

straight-shoota:
It would help a lot if you could give us instructions on how to reproduce this issue from scratch.

matthewmcgarvey:
I think I'm seeing this issue when a shard I'm installing has a dependency on the overridden shard and has a postinstall step that calls `shards build`. It doesn't see that the dependency is satisfied through an override and install the version it wants, overwriting the override install.

jgaskins:
[...]

PR #425, opened 2020-07-28
Fix: reinstall when resolver changes

waj:
Currently running an `install` after changing the resolver from `path` to `git` doesn't do anything because `GitResolver` fails to detect that the current install is not what's required.

Now the `Package` uses the `.shards.info` to check that the resolver and version matches to figure out if the dependency is installed.

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -300,6 +300,28 @@ describe "install" do
     end
   end
 
+  it "reinstall when resolver changes" do
+    metadata = {dependencies: {web: {git: git_url(:web)}}}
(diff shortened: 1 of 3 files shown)


waj:
@bcardiff I plan to (re)move some code now present within the `Resolver` to the `Package`. That logic you mention only worked for git repositories. Now it also works for path sources.

jwoertink:
Could this be related? https://github.com/crystal-lang/shards/issues/398

waj:
@jwoertink is that still happening with v0.11.1? Anyway I don't think it's related. Let's continue the conversation on the issue.

PR #426, opened 2020-07-31
Move install responsibilities from Resolver to Package

waj:
This is the first part of a refactor I'm working on to better separate responsibilities between internal entities in Shards. I split these changes in two so it can be reviewed more easily. With this PR the `Resolver` still represents a source to obtain different versions of a shard, but the installation is delegated completely to the `Package` class.

This will make more sense with the second part, where locks and installed packages (`Shards.info.installed`) will return instances of `Package` instead of `Dependency`. In other words, finally the `Dependency` will represent just the intention with some more or less precise definition of which version is desired, while a `Package` represents a specific version that must be or is already installed.

There is a small change in behaviour after this refactor, but I think in the good direction. Now the only source of truth to determine if a shard is already installed is the `.shard.info` file. So the contents of the shard could be manually changed, or the symlink created for a `path` resolver modified but Shards still would assume the dependency is already there and nothing must be done to fix it. Only in the case of a complete removal of the directory or symlink it will reinstall the dependency ignoring what was recorded in the info file.

### DIFF
--- a/src/commands/list.cr
+++ b/src/commands/list.cr
@@ -14,19 +14,21 @@ module Shards
 
       private def list(dependencies, level = 1)
         dependencies.each do |dependency|
-          resolver = dependency.resolver
-
-          # FIXME: duplicated from Check#verify
-          unless _spec = resolver.installed_spec
+          installed = Shards.info.installed[dependency.name]
+          unless installed
             Log.debug { "#{dependency.name}: not installed" }
             raise Error.new("Dependencies aren't satisfied. Install them with 'shards install'")
           end
 
+          version = installed.requirement.as(Shards::Version)
+          package = Package.new(installed.name, installed.resolver, version)
+          resolver = installed.resolver
+
           indent = "  " * level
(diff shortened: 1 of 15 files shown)


waj:
I just changed `Package#spec` so it reads the `shard.yml` from filesystem when the package is installed. This brings back functionality similar to `Resolver#installed_spec`. Otherwise, when running `spec check` or `spec list` it would "fetch" the source repositories.

src/commands/list.cr
```diff
@@ -14,19 +14,21 @@ module Shards
 
       private def list(dependencies, level = 1)
         dependencies.each do |dependency|
-          resolver = dependency.resolver
-
-          # FIXME: duplicated from Check#verify
-          unless _spec = resolver.installed_spec
+          installed = Shards.info.installed[dependency.name]
```
Sija, line 17:
```suggestion
          installed = Shards.info.installed[dependency.name]?
```

waj: Thanks @Sija good catch! I added a spec to check for that condition

PR #428, opened 2020-08-04
Use Package for locks and installed shards

waj:
This is the followup from #426. Previously, locks and installed shards were represented as `Dependency` with a `Version` requirement. But that forces to do castings and checks on many places. With this change the code is simplified, and the check for invalid dependency is only done while parsing.

### DIFF
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -53,7 +53,7 @@ module Shards
       end
     end
 
-    def as_package
+    def as_package?
       version =
         case req = @requirement
         when VersionReq then Version.new(req.to_s)
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -58,6 +58,8 @@ module Shards
         case req = @requirement
         when VersionReq then Version.new(req.to_s)
         else
+          # This conversion is used to keep compatibility
+          # with old versions (1.0) of lock files.
           versions = @resolver.versions_for(req)
           unless versions.size == 1
             return
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -11,7 +11,7 @@ module Shards
     def initialize(@name : String, @resolver : Resolver, @requirement : Requirement = Any)
     end
 
-    def self.from_yaml(pull : YAML::PullParser, *, is_lock = false)
+    def self.from_yaml(pull : YAML::PullParser)
       mapping_start = pull.location
       name = pull.read_scalar
       pull.read_mapping do
@@ -40,10 +40,6 @@ module Shards
         resolver = resolver_data[:type].find_resolver(resolver_data[:key], name, resolver_data[:source])
 
         requirement = resolver.parse_requirement(params)
-        if is_lock && requirement.is_a?(VersionReq)
-          requirement = Version.new(requirement.to_s)
-        end
-
         Dependency.new(name, resolver, requirement)
       end
     end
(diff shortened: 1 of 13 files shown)


src/dependency.cr
```diff
@@ -57,6 +53,21 @@ module Shards
       end
     end
 
+    def as_package
```
bcardiff, line 56:
```suggestion
    def as_package : Package?
```

for clarity, but as per another comment this method might not be needed.

src/lock.cr
```diff
@@ -32,13 +33,18 @@ module Shards
             case key = pull.read_scalar
             when "shards"
               pull.each_in_mapping do
-                dependencies << Dependency.from_yaml(pull, is_lock: true)
+                dep = Dependency.from_yaml(pull)
```
bcardiff, line 36: I think we should add a `Package.from_yaml`. The `Dependency#as_package` might call the resolver while trying to convert it to a package and that does not seems right to do here.

src/molinillo_solver.cr
```diff
@@ -21,27 +21,15 @@ module Shards
...
-            unless versions.size == 1
-              Log.warn { "Lock for shard \"#{dep.name}\" is invalid" }
-              return
-            end
-            lock.requirement = versions.first
-          end
-
         check_single_resolver_by_name dep.resolver
-        base.add_vertex(lock.name, lock, true)
+        base.add_vertex(lock.name, Dependency.new(lock.name, lock.resolver, lock.version), true)
```
bcardiff, line 25:
```suggestion
        base.add_vertex(lock.name, Dependency.new(lock.name, dep.resolver, lock.version), true)
```

We should use dep.resolver to allo chaning source without bumping the version when possible.
Otherwise `check_single_resolver_by_name` should be called with lock.resolver and it will get ambiguos source when the shard.yml and lock differ in source.

src/package.cr
```diff
@@ -11,6 +11,8 @@ module Shards
     def initialize(@name, @resolver, @version, @is_override = false)
     end
 
+    def_equals @name, @resolver, @version
```
bcardiff, line 14: This is not exactly the same comparision that was used in `commands/install.cr`

I think it works because the same resolver instance is returned if the same key/name/source is used. And there is no equality defined in Resolver, so the default reference equality is used.

Maybe the equality should be defined in Resolver and compare class, name and source.

waj: `Resolver#==` is already defined:
https://github.com/waj/shards/blob/rework/package/src/resolvers/resolver.cr#L24-L28

src/lock.cr
waj: The call to the resolver is done only when converting from an old `shard.lock` file. Otherwise it cannot be converted to a plain `Version`. In the future we could just deprecate lock files with version `1.0` and the `Package` could be parsed directly from the yaml.

src/dependency.cr
waj: Yeah... maybe the method should be named `as_package?` to make it even more clear

src/molinillo_solver.cr
waj: Is that a scenario not already covered by specs?

bcardiff: It was added in https://github.com/crystal-lang/shards/pull/419/files#diff-d38f7bcbd4d1fda0933136e184879150R268-R282 . With your proposed change (using `lock.resolver`) I would expect that the `assert_locked` would wrongly use `git_url(:awesome)` instead of `git_url(:forked_awesome)`

src/package.cr
bcardiff: Oh. ok then 👍

src/lock.cr
bcardiff: Ok, Let's add a comment there to keep track of that.

src/molinillo_solver.cr
bcardiff: Ok, since previously the whole lock was added as a vertex and that included the lock's resolver.

src/lock.cr
waj: Comments added

src/molinillo_solver.cr
waj: Anyway I changed the resolver as you suggested. It's the right thing, even though the resolver from the locks is not being used in the resolution process.

PR #436, opened 2020-09-19
Don't use POSIX-specific shell constructs

oprypin:
This is part of work to bring 'shards' to Windows.

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -113,7 +113,7 @@ module Shards
 
     protected def self.has_git_command?
       if @@has_git_command.nil?
-        @@has_git_command = Process.run("command -v git", shell: true).success?
+        @@has_git_command = (Process.run("git", ["--version"]).success? rescue false)
       end
       @@has_git_command
     end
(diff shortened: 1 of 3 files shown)


src/resolvers/git.cr
```diff
@@ -113,7 +113,7 @@ module Shards
 
     protected def self.has_git_command?
       if @@has_git_command.nil?
-        @@has_git_command = Process.run("command -v git", shell: true).success?
+        @@has_git_command = (Process.run("git", ["--version"]).success? rescue false)
```
Sija, line 116:
```suggestion
        @@has_git_command = Process.run("git", ["--version"]).success? rescue false
```

oprypin: It can work both ways but just consider what advantages we're weighing here (picking one or the other):
* Saving two characters.
* Not having to think what the rescue *really* applies to every time when reading the code. I literally didn't know what the priority would be and I wanted to save everyone the cognitive load and the possible bugs.

Sija: Ok, I thought it's pretty clear what's going on here (there's nothing else that 'rescue' could apply to), but it's your call.

jhass (approved):

oprypin:
@bcardiff have you had a chance to look at [my pull requests to Shards](https://github.com/crystal-lang/shards/pulls/oprypin)?

straight-shoota (approved):

PR #444, opened 2020-11-12
Touch install_path and lockfile to express dependency

### LINKED ISSUE #390: Update modification time of install dir and shards.lock
If `shards.yml` was updated with not changes to the dependency resolution, the modification time of `shards.yml` is newer than that of `shards.lock` and/or the install dir (`.crystal/shards`) even after running `shards update`.
This breaks mtime-based freshness detection (for example used by make). See #380 for details.
`shards update` should update the modification time of both `shard.lock` and `.crystal/shards` even if it didn't change anything. Ditto for `shards install`.

/cf https://github.com/crystal-lang/shards/pull/380#issuecomment-624007356 f.

straight-shoota:
This patch makes sure that modification times are always in order `lib > shard.lock > shard.yml`. This is necessary for mtime-based dependency resolution for example with make.

Resolves #390

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,6 @@ test_integration: bin/shards phony
 lib: shard.lock
 	mkdir -p lib/molinillo
 	$(SHARDS) install || (curl -L $(MOLINILLO_URL) | tar -xzf - -C lib/molinillo --strip-components=1)
-	touch lib
 
 shard.lock: shard.yml
 	$(SHARDS) update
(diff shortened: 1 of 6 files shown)


straight-shoota:
Related:  #381

straight-shoota:
Failing specs depend on #445

straight-shoota:
Depends on #453

bcardiff (approved):

PR #447, opened 2020-11-14
Expand Windows support + fix all specs

oprypin:


src/resolvers/git.cr
```diff
@@ -357,6 +362,10 @@ module Shards
 
     # Parses a URI string, with additional support for ssh+git URI schemes.
     private def parse_uri(raw_uri)
+      if (path = raw_uri.lchop?("file://"))
```
straight-shoota, line 365: Would be nice to have a comment on this as well.
Also what about `file:\\\\` URIs?

src/package.cr
```diff
@@ -101,9 +102,13 @@ module Shards
       Dir.mkdir_p(Shards.bin_path)
 
       spec.executables.each do |name|
-        Log.debug { "Install bin/#{name}" }
-        source = File.join(install_path, "bin", name)
-        destination = File.join(Shards.bin_path, name)
+        exe_name = name
+        {% if flag?(:win32) %}
+          exe_name += ".exe"
```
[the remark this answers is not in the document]
oprypin: OK. Not feeling strongly about this.

src/resolvers/git.cr
```diff
@@ -214,20 +215,24 @@ module Shards
...
       @local_path ||= begin
         uri = parse_uri(git_url)
 
-        path = uri.path.to_s[1..-1]
-        path = path.gsub('/', File::SEPARATOR) unless File::SEPARATOR == '/'
+        path = uri.path
         path += ".git" unless path.ends_with?(".git")
+        path = Path[path]
+        if (anchor = path.anchor)
+          path = Path[path.drive.to_s.rchop(":"), path.relative_to(anchor)]
```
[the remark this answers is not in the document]
oprypin: Comment done.

One thing I'm really sure of is that no, `join` should never consider putting the anchor in the middle of the path. It's doing it at the moment, though, just that nobody (except me) had given it a second thought before seeing this edge case. Respectfully, `join`'s behavior on Windows is more useful than POSIX because at least in almost all cases it will fail with a loud bang (unfortunately indirectly though).

The fact that this strange operation is explicitly chosen here doesn't mean anything for general usefulness of this. It should stay explicit.

src/resolvers/git.cr
oprypin: Comment done.

I don't know, what about them? Backslashes are not valid here, and if there are 4 slashes, then I guess 2 will remain, and maybe Windows will still do something useful with them?
https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes.3F

src/resolvers/git.cr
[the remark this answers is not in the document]
straight-shoota: Yeah, but maybe a method to transform an anchored path into a safe-to-use relative path could be a useful tool.

src/resolvers/git.cr
straight-shoota: Sry, I wrote the backslashes escaped. I just meant `file:\\` followed by an anchored path. This is at least somewhat common on windows, although not valid as a URI.
Still, maybe adding `|| (path = raw_uri.lchop?("file:\\\\"))` wouldn't be a bad idea?

straight-shoota (approved):

spec/support/factories.cr
```diff
@@ -142,11 +150,15 @@ def run(command, *, env = nil)
   }
   cmd_env.merge!(env) if env
   output, error = IO::Memory.new, IO::Memory.new
-  status = Process.run(command, shell: true, env: cmd_env, output: output, error: error)
+  {% if flag?(:win32) %}
+    # Concurrent streams are currently broken on Windows. Drop one.
```
Sija, line 154: Perhaps it's worth adding a `FIXME` ?

src/commands/prune.cr
```diff
@@ -1,19 +1,20 @@
 require "file_utils"
 require "./command"
+require "../helpers/files"
```
Sija, line 3: This file should be moved:

```suggestion
require "../helpers"
```

straight-shoota: It doesn't really matter either way.

src/info.cr
```diff
@@ -21,7 +21,11 @@ class Shards::Info
     Dir.mkdir_p(@install_path)
 
     unless File.exists?(info_path)
-      Dir[File.join @install_path, "*.sha1"].each { |p| File.delete(p) }
+      Dir.each_child(@install_path) do |name|
```
bcardiff, line 24: Why is this needed? I thought glob were already supported in windows.

oprypin: No it doesn't work.

https://github.com/crystal-lang/crystal/blob/5999ae29beacf4cfd54e232ca83c1a46b79f26a5/src/dir/glob.cr#L6

oprypin: Also it's a change for the better because now file paths with `*` aren't a hazard

bcardiff: Ok, but the culprit for this change is `File.join` then, not `Dir#[]`. I'm ok with the change, but I wanted to know the reason for it.

bcardiff (approved):

src/info.cr
straight-shoota: `each_child` is actually better :+1:

In general: glob patterns should just be written as string interpolation, not use `File.join` (you could use `Path.posix` but what's the point?).

src/commands/prune.cr
Sija: IMO it does from the good practices standpoint - i.e. naming files according to their namespace path `Foo::Bar::Baz` -> `foo/bar/baz.cr`

oprypin: Yes, that does actually seem better. Will include this change. No idea

straight-shoota: @Sija In general, that's true. But for this use case not so much IMO. A helpers namespace is intended to contain a mixture of different things (eventually). Otherwise you would use a more specifc namespace. When there are lots of helpers you may not want to place them in a single file at `./helpers` but separted by concerns in `./helpers/*`. Currently, we don't need that. But maybe in the future. It can be adapter later when necessary, but it wouldn't hurt to leave it as it was.
The change from `helpers/files` to `helper` IMO didn't help to make the PR any better, yet it requires another round of reviews (quick ones, obviously) when it could have just been merged as it was. 🤷‍♂️

straight-shoota (approved):

straight-shoota:
Thanks @oprypin

PR #450, opened 2020-12-03
Improve git reliability by retrying on failures

fudanchii:
Occasionally, though it's always happen with my connection, either `git clone --mirror` or `git fetch --all` via http will tend to fail with closed connection.

Doing retry directly on the git command seem to fix the situation. In addition, bundle also doing this with their [git_proxy](https://github.com/rubygems/rubygems/blob/f7f8f7c968e31ebcadf1cc7b8243ef023920a65f/bundler/lib/bundler/source/git/git_proxy.rb#L157-L161)

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -312,15 +312,27 @@ module Shards
       # be used interactively.
       # This configuration can be overriden by defining the environment
       # variable `GIT_ASKPASS`.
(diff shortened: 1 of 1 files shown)


bcardiff (approved):
LGTM

straight-shoota (approved):

straight-shoota:
Thanks @fudanchii

PR #453, opened 2020-12-11
Write lockfile even when there are no dependencies

### LINKED ISSUE #445: Lockfile is not overridden when dependencies are removed
When a lockfile exists with locked dependencies and all dependencies are removed from `shard.yml`, the next run of `shards install`/`shards update` does not update the lockfile which leads to incorrect information.

This test case describes this:

```cr
  it "updated lockfile on completely removed dependencies" do
    metadata = NamedTuple.new
    lock = {web: "1.0.0"}

    with_shard(metadata, lock) do
      run "shards install" # or update ...

      refute_installed "web"
      refute_locked "web"
    end
  end
```

The reason is one condition for `generate_lockfile?` is packages not being empty. That just seems incorrect.

The question is, what should happen instead?
It seems the current assumption is that no dependencies means no shard.lock at all. I don't think that's a good idea because it can lead to running `shards install` all the time when a `Makefile` depends on `shards.lock`. IMO both `shards install` and `shards update` should always write `shard.lock` file (except for production mode). If there are no dependencies, `shard.lock` would just be empty. It still represents that dependency resolution has happened.
bundler also writes a `Gemfile.lock` with no dependencies.

Related: #381

straight-shoota:
With this change `shards update` and `shards install` (except with `--production` option) *always* write `shard.lock` (or touch if there are not changes) even if there are no dependencies.

Resolves #445
Unlocks #444

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -458,6 +458,7 @@ describe "install" do
     }
 
     with_shard(metadata) do
+      File.exists?(File.join(application_path, "shard.lock")).should be_false
       run "shards install --production"
 
       # it installed dependencies (recursively)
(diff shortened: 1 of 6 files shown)


bcardiff (approved):
So much love inspired by Makefiles :-)

PR #459, opened 2021-01-03
Mercurial command server

f-fr:
This is a modification of the HgResolver from some other pull request to make use of the hg command server to speed up hg commands.

The mercurial command server is feature of hg to overcome the (relatively long) startup time of the Python interpreter. Instead of starting a new interpreter for each single hg command only a single process is started. All subsequent hg commands are then passed to this running process.

The command server is started by `HgResolver` the first time a command is run. The same command server is also used for all test cases.

This overall reduces the running time of all hg tests to a few seconds.

### DIFF
--- a/src/resolvers/hg.cr
+++ b/src/resolvers/hg.cr
@@ -110,6 +110,7 @@ module Shards
   class HgResolver < Resolver
     @@has_hg_command : Bool?
     @@hg_version : String?
+    @@hg_process : Process | Bool | Nil
 
     @origin_url : String?
 
@@ -134,7 +135,7 @@ module Shards
     end
 
     protected def self.hg_version
-      @@hg_version ||= `hg --version`[/\(version\s+([^)]*)\)/, 1]
+      @@hg_version ||= hg("--version")[/\(version\s+([^)]*)\)/, 1]
     end
 
     def read_spec(version : Version) : String?
@@ -142,7 +143,7 @@ module Shards
       ref = hg_ref(version)
 
       if file_exists?(ref, SPEC_FILENAME)
-        capture("hg cat -r #{Process.quote(ref.to_hg_ref)} #{Process.quote(SPEC_FILENAME)}")
+        capture_hg("cat", "-r", ref.to_hg_ref, SPEC_FILENAME)
       else
         Log.debug { "Missing \"#{SPEC_FILENAME}\" for #{name.inspect} at #{ref}" }
         nil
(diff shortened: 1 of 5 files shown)


src/resolvers/hg.cr
```diff
@@ -0,0 +1,571 @@
...
+        version.has_metadata?
+      else
+        # TODO: check branch and tags
+        true
+      end
+    end
+
+    protected def versions_from_tags
+      capture_hg("tags", "--template", "{tag}\\n")
+        .split('\n')
```
Sija, line 206:
```suggestion
        .lines
```

src/resolvers/hg.cr
```diff
@@ -0,0 +1,571 @@
...
+        case key
+        when "branch"
+          return HgBranchRef.new value
+        when "bookmark"
+          return HgBookmarkRef.new value
+        when "tag"
+          return HgTagRef.new value
+        when "commit"
+          return HgCommitRef.new value
+        else
```
Sija, line 271:
```suggestion
```

src/resolvers/hg.cr
```diff
@@ -0,0 +1,571 @@
...
+
+    private def mirror_repository
+      path = local_path
+      FileUtils.rm_r(path) if File.exists?(path)
+      Dir.mkdir_p(path)
+
+      source = hg_url
+      # Remove a "file://" from the beginning, otherwise the path might be invalid
+      # on Windows.
+      source = source[7..] if source.starts_with?("file://")
```
Sija, line 333:
```suggestion
      source = source.lchop("file://")
```

src/resolvers/hg.cr
```diff
@@ -0,0 +1,571 @@
...
+      hg_retry(err: "Failed to update #{hg_url}") do
+        run_hg "pull"
+      end
+    end
+
+    private def hg_retry(err = "Failed to update repository")
+      retries = 0
+      loop do
+        yield
+        break
```
Sija, line 354:
```suggestion
        return yield
```

src/resolvers/hg.cr
```diff
@@ -0,0 +1,571 @@
...
+      @origin_url = nil
+    end
+
+    private def cloned_repository?
+      Dir.exists?(local_path)
+    end
+
+    private def valid_repository?
+      File.each_line(File.join(local_path, ".hg", "dirstate")) do |line|
+        return true if line =~ /mirror\s*=\s*true/
```
Sija, line 374:
```suggestion
        return true if line.matches?(/mirror\s*=\s*true/)
```

Sija (commented):
Couple of idiomatic suggestions.

PR #461, opened 2021-01-12
Fix README referencing CRFLAGS

straight-shoota:
`CRFLAGS` was dropped in favour of generic `CRYSTAL_OPTS` in #344

/cf https://github.com/crystal-lang/shards/pull/414#issuecomment-758347018

Maybe we should add a `release=1` flag to the makefile for this use case to add some abstraction?


Blacksmoke16:
> Maybe we should add a `release=1` flag to the makefile for this use case to add some abstraction?

Wasn't this already done as part of https://github.com/crystal-lang/shards/pull/414?

straight-shoota:
Oops, I just looked at #344 🙈

bcardiff (approved):

Issue #413, opened 2020-06-22
Due to the default of `crystal: <1.0.0`, can't install shards for Crystal nightlies


oprypin:
In the wild:

https://github.com/oprypin/crystal-chipmunk/runs/796810944?check_suite_focus=true#step:16:12

My shard became not installable as a dependency ever since crystal-1.0.0-dev.

I think that (despite @jhass [hinting](https://gitter.im/crystal-lang/crystal?at=5ef11502a813c72dcf037be0) that this might be intentional), from the point of view of any user, this looks like a bug in Shards. Some things come to mind:

* Is it that `>= 0.25.1` and `< 1.0.0` are conflicting?
* Wait, there's never any `< 1.0.0` specified..?
* Why did it stop working?

---

Repro:

```shell
$ cat shard.yml
```
```yaml
name: main
version: 0.1.0

dependencies:
  sub:
    path: sub
```
```shell
$ cat sub/shard.yml
```
```yaml
name: sub
version: 0.1.0
```
```shell
$ PATH=~/repos/crystal/.build:/usr/bin shards install
```
```
Resolving dependencies
Unable to satisfy the following requirements:

- `crystal (< 1.0.0)` required by `sub 0.1.0`
Failed to resolve dependencies
```
```shell
$ shards --version
```
```
Shards 0.11.1 (2020-06-21)
```
```shell
$ ~/repos/crystal/.build/crystal env
```
```
CRYSTAL_CACHE_DIR=/tmp/cache-blaxpirit/crystal
CRYSTAL_PATH=lib
CRYSTAL_VERSION=1.0.0-dev
CRYSTAL_LIBRARY_PATH=''
CRYSTAL_OPTS=''
```

bcardiff:
If there is no crystal property the semantic is `crystal: < 1.0.0`

`shards install --ignore-crystal-version` can be used as a workaround until a more fine grained alternative of forcing the resolution is added.

Maybe shards should also read a `SHARDS_OPTS` env variable so this can be widely configured 🤔

oprypin:
> If there is no crystal property the semantic is crystal: `< 1.0.0`

Yea I get it, but nobody asked for that to be that way??

bcardiff:
It is a choice that make things work similar to what will happen from 1.0 to 2.0 eventually. Major releases will require some work from the ecosystem to migrate, or for the user to use some workaround.

[...]

PR #463, opened 2021-01-13
Bump supported Crystal version to ">= 0.35.0" in shard.yml

Sija:


straight-shoota (approved):

bcardiff:
I don't agree with >= version restrictions. Compile with --ignore-crystal-version until 1.0 is out and ecosystem is updated.

straight-shoota:
But it *is* compatible with 0.35.0 and 1.0. If not open ended, it could at least be something like `crystal: >= 0.35.1, < 2.0`.

I am strictly against forcing to use `--ignore-crystal-version` everywhere as per https://github.com/crystal-lang/shards/issues/413#issuecomment-727633776. It that's supposed to be the solution, it would be much, much better for everyone to just not enforce Crystal versions at all.

Sija:
@bcardiff As far as I understand, shards _are_ and **will be** compatible with crystal `>= 0.35.0`, right?  _I don't agree with_ doesn't sound like an argument but your private opinion, no offence - and it seems a bit far away from welcoming community standards and meritocratic approach, which I hope you agree with as beneficiary.

bcardiff:
@Sija I believe I have expressed my concerns in posts and issues already why not having and upper limit in the stated crystal version is a practice I won't encourage.

@straight-shoota proposal is better in this regard. Yet we need first to release 1.0-pre1 and to everybody to honor that no more breacking changes are wanted in 1.x to make the statement of for `< 2.0`. Even small breaking changes that are motivated by valid cleanups and improvements will break that statement.

Of course, there are ways of disregard whatever version restriction is stated. You need to put whatever you want in lib and the compiler lookup path will use it. But I prefer for version restrictions to mean something, even if they are more conservative.

straight-shoota:
I don't think you have expressed your concerns enough. At least I don't follow them. And there was no response to the latest counter arguments in #413.

Around major releases with breaking changes, a version restriction is always suject to being incorrect, wether it's overspecific or underspecific. Being conservative IMO has mostly ideal value, but practically it's relatively irrelevant whether the stated version compatibility is factual. We're only talking about a development version here. It just makes using it harder for everyone. I don't think we need to release 1.0-pre1 before stating that shards is compatible with 1.0. At this point it should be a given that the next shards release is *definitely* compatible with 1.0.

PR #464, opened 2021-01-14
Add macOS nightlies to CI

Sija:
Refactored some parts as well like `include`-ing windows in the `matrix` instead of having it there from start - especially apt as windows support is still in-the-works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,19 +4,19 @@ on:
   push:
   pull_request:
   schedule:
-    - cron: '0 6 * * 1'  # Every monday 6 AM
+    - cron: "0 6 * * 1" # Every monday 6 AM
 
 jobs:
-  build:
+  test:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
-        crystal: [null]
+        os: [ubuntu-latest, macos-latest]
+        crystal: [latest, nightly]
         include:
-          - os: ubuntu-latest
+          - os: windows-latest
             crystal: nightly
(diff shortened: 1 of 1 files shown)


straight-shoota (approved):

Sija:
This could be merged...

Sija:
@bcardiff Could you please review this?

bcardiff:
Is there a reason to exclude windows from using crystal latest?

Sija:
@bcardiff yes, it doesn't work. for more detailed explanation please ask @oprypin.

oprypin:
And the reason for that is https://github.com/crystal-lang/distribution-scripts/issues/64#issuecomment-778775239

oprypin:
@Sija Why do you have to change the whole file's formatting to adhere to your liking?

Sija:
@oprypin _Whole file_? Don't exaggerate, adding some spaces for readability and changing name of one job is not _whole file_. What's wrong with those changes anyway?

asterite:
@Sija They are irrelevant to the change you want to make.

If you want to change someone else's style, I think that should be discussed and done separately.

Code is written by people if you go and change that when it's not needed, while also adding additional changes, the ones that wrote the original code might think "What's your problem with my style? BAM. I don't like your PR."

Not that that will effectively happened. But at least in my case when I contribute to a project I tend to avoid changing style things if they have nothing to do with the PR's goal.

asterite:
They also make reviewing the PR harder! "Why was this changed? Is this relevant? I don't understand."

bcardiff (approved):
I'm fine the changes. Let's keep it calm. But I agree that changes that satisfies author's taste should be avoided in general.

PR #471, opened 2021-01-28
feat: treat github sources as case insensitive

### LINKED ISSUE #470: github ambiguous sources
I'm getting some ambiguous sources errors when updating shards.
I have solved the problem temporarily by using shard.override however I think there is a better way, at least for github

The issue is that two separate shards are depending on `retriable`

```yaml
dependencies:
  retriable:
    github: Sija/retriable.cr
```

and

```yaml
dependencies:
  retriable:
    github: sija/retriable.cr
```

note the difference in case on `sija`

However github is case insensitive when if comes to URLs so I propose that for github dependencies the URLs are downcased to avoid errors like this:

```
Error shard name (retriable) has ambiguous sources: 'git: https://github.com/Sija/retriable.cr.git' and 'git: https://github.com/sija/retriable.cr.git'.
```

which would greatly improve the experience of using shards

stakach:
fixes #470 #275

### DIFF
--- a/spec/unit/git_resolver_spec.cr
+++ b/spec/unit/git_resolver_spec.cr
@@ -30,20 +30,21 @@ module Shards
 
     it "normalizes github bitbucket gitlab sources" do
       # deal with case insensitive paths
-      GitResolver.normalize_key_source("github", "repo/path").should eq({"git", "https://github.com/repo/path"})
-      GitResolver.normalize_key_source("github", "rEpo/pAth").should eq({"git", "https://github.com/repo/path"})
-      GitResolver.normalize_key_source("github", "REPO/PATH").should eq({"git", "https://github.com/repo/path"})
-      # GitResolver.normalize_key_source("bitbucket", "repo/path").should eq({"git", "https://bitbucket.com/repo/path"})
-      GitResolver.normalize_key_source("bitbucket", "rEpo/pAth").should eq({"git", "https://bitbucket.com/repo/path"})
-      # GitResolver.normalize_key_source("bitbucket", "REPO/PATH").should eq({"git", "https://bitbucket.com/repo/path"})
-      # GitResolver.normalize_key_source("gitlab", "repo/path").should eq({"git", "https://gitlab.com/repo/path"})
-      GitResolver.normalize_key_source("gitlab", "rEpo/pAth").should eq({"git", "https://gitlab.com/repo/path"})
-      # GitResolver.normalize_key_source("gitlab", "REPO/PATH").should eq({"git", "https://gitlab.com/repo/path"})
+      GitResolver.normalize_key_source("github", "repo/path").should eq({"git", "https://github.com/repo/path.git"})
+      GitResolver.normalize_key_source("github", "rEpo/pAth").should eq({"git", "https://github.com/repo/path.git"})
+      GitResolver.normalize_key_source("github", "REPO/PATH").should eq({"git", "https://github.com/repo/path.git"})
+      GitResolver.normalize_key_source("bitbucket", "repo/path").should eq({"git", "https://bitbucket.com/repo/path.git"})
+      GitResolver.normalize_key_source("bitbucket", "rEpo/pAth").should eq({"git", "https://bitbucket.com/repo/path.git"})
+      GitResolver.normalize_key_source("bitbucket", "REPO/PATH").should eq({"git", "https://bitbucket.com/repo/path.git"})
+      GitResolver.normalize_key_source("gitlab", "repo/path").should eq({"git", "https://gitlab.com/repo/path.git"})
+      GitResolver.normalize_key_source("gitlab", "rEpo/pAth").should eq({"git", "https://gitlab.com/repo/path.git"})
+      GitResolver.normalize_key_source("gitlab", "REPO/PATH").should eq({"git", "https://gitlab.com/repo/path.git"})
 
       # also normalise full git paths
-      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Github.com/Repo/Path.git?Shallow=true").should eq "https://User:Pass@github.com/repo/path.git?Shallow=true"
-      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Bitbucket.com/Repo/Path.Git?Shallow=true"). should eq "https://User:Pass@bitbucket.com/repo/path.git?Shallow=true"
-      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Gitlab.com/Repo/Path?Shallow=true").should eq "https://User:Pass@gitlab.com/repo/path.git?Shallow=true"
+      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Github.com/Repo/Path.git?Shallow=true")[1].should eq "https://User:Pass@github.com/repo/path.git?Shallow=true"
+      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Bitbucket.com/Repo/Path.Git?Shallow=true")[1].should eq "https://User:Pass@bitbucket.com/repo/path.git?Shallow=true"
+      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Gitlab.com/Repo/Path?Shallow=true")[1].should eq "https://User:Pass@gitlab.com/repo/path.git?Shallow=true"
+      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@www.Github.com/Repo/Path?Shallow=true")[1].should eq "https://User:Pass@github.com/repo/path.git?Shallow=true"
 
       # don't normalise other domains
       GitResolver.normalize_key_source("git", "HTTPs://mygitserver.com/Repo.git").should eq({"git", "HTTPs://mygitserver.com/Repo.git"})
--- a/spec/unit/git_resolver_spec.cr
+++ b/spec/unit/git_resolver_spec.cr
@@ -30,19 +30,20 @@ module Shards
 
     it "normalizes github bitbucket gitlab sources" do
       # deal with case insensitive paths
-      paths = {"repo/path", "rEpo/pAth", "REPO/PATH"}
-      keys = {"github", "bitbucket", "gitlab"}
-      results = {"https://github.com/repo/path.git", "https://bitbucket.com/repo/path.git", "https://gitlab.com/repo/path.git"}
-      keys.each_with_index do |key, index|
-        paths.each { |path| GitResolver.normalize_key_source(key, path).should eq({"git", results[index]}) }
-      end
+      GitResolver.normalize_key_source("github", "repo/path").should eq({"git", "https://github.com/repo/path"})
+      GitResolver.normalize_key_source("github", "rEpo/pAth").should eq({"git", "https://github.com/repo/path"})
+      GitResolver.normalize_key_source("github", "REPO/PATH").should eq({"git", "https://github.com/repo/path"})
+      # GitResolver.normalize_key_source("bitbucket", "repo/path").should eq({"git", "https://bitbucket.com/repo/path"})
+      GitResolver.normalize_key_source("bitbucket", "rEpo/pAth").should eq({"git", "https://bitbucket.com/repo/path"})
+      # GitResolver.normalize_key_source("bitbucket", "REPO/PATH").should eq({"git", "https://bitbucket.com/repo/path"})
+      # GitResolver.normalize_key_source("gitlab", "repo/path").should eq({"git", "https://gitlab.com/repo/path"})
+      GitResolver.normalize_key_source("gitlab", "rEpo/pAth").should eq({"git", "https://gitlab.com/repo/path"})
+      # GitResolver.normalize_key_source("gitlab", "REPO/PATH").should eq({"git", "https://gitlab.com/repo/path"})
 
       # also normalise full git paths
-      paths = {"HTTPS://User:Pass@Github.com/Repo/Path.git?Shallow=true", "HTTPS://User:Pass@Bitbucket.com/Repo/Path.Git?Shallow=true", "HTTPS://User:Pass@Gitlab.com/Repo/Path?Shallow=true"}
-      results = {"https://User:Pass@github.com/repo/path.git?Shallow=true", "https://User:Pass@bitbucket.com/repo/path.git?Shallow=true", "https://User:Pass@gitlab.com/repo/path.git?Shallow=true"}
-      paths.each_with_index do |path, index|
-        GitResolver.normalize_key_source("git", path).should eq({"git", results[index]})
-      end
+      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Github.com/Repo/Path.git?Shallow=true").should eq "https://User:Pass@github.com/repo/path.git?Shallow=true"
+      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Bitbucket.com/Repo/Path.Git?Shallow=true"). should eq "https://User:Pass@bitbucket.com/repo/path.git?Shallow=true"
+      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Gitlab.com/Repo/Path?Shallow=true").should eq "https://User:Pass@gitlab.com/repo/path.git?Shallow=true"
 
       # don't normalise other domains
       GitResolver.normalize_key_source("git", "HTTPs://mygitserver.com/Repo.git").should eq({"git", "HTTPs://mygitserver.com/Repo.git"})
--- a/spec/unit/git_resolver_spec.cr
+++ b/spec/unit/git_resolver_spec.cr
@@ -28,14 +28,24 @@ module Shards
       create_git_tag "library", "99.9.9"
     end
 
-    it "normalizes github bitbucket gitlab sources", focus: true do
-      # to deal with case insensitive paths
+    it "normalizes github bitbucket gitlab sources" do
+      # deal with case insensitive paths
       paths = {"repo/path", "rEpo/pAth", "REPO/PATH"}
       keys = {"github", "bitbucket", "gitlab"}
       results = {"https://github.com/repo/path.git", "https://bitbucket.com/repo/path.git", "https://gitlab.com/repo/path.git"}
       keys.each_with_index do |key, index|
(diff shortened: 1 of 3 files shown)


src/resolvers/git.cr
```diff
@@ -104,7 +104,9 @@ module Shards
       case key
       when "git"
         {"git", source}
-      when "github", "bitbucket", "gitlab"
+      when "github"
+        {"git", "https://#{key}.com/#{source.downcase}.git"}
+      when "bitbucket", "gitlab"
```
straight-shoota, line 109: Project URLs on bitbucket and gitlab are case insensitive as well, so might apply this change to all of them.

stakach: yeah I wasn't actually 100% on that, but just tested and seems to be the case!

bcardiff:
And I noticed the PR after mi comment in the issue

---

Would it be enough to normalize when `github:` is used, or should it also be lowercased when `git: https://...` is used? 

If so, we should not transform user/password to avoid breakinig `https://<token>@github.com/owner/repo.git` or `https://<token>:x-oauth-basic@github.com/owner/repo.git` private repos.

bcardiff:
Could you add some tests `git_resolver_spec.cr`? 🙈

stakach:
@bcardiff added code to deal with generic `git: urls` when they match github, bitbucket or gitlab

bcardiff (approved):

spec/unit/git_resolver_spec.cr
```diff
@@ -28,6 +28,26 @@ module Shards
...
+      keys.each_with_index do |key, index|
+        paths.each { |path| GitResolver.normalize_key_source(key, path).should eq({"git", results[index]}) }
+      end
+
+      # also normalise full git paths
+      paths = {"HTTPS://User:Pass@Github.com/Repo/Path.git?Shallow=true", "HTTPS://User:Pass@Bitbucket.com/Repo/Path.Git?Shallow=true", "HTTPS://User:Pass@Gitlab.com/Repo/Path?Shallow=true"}
+      results = {"https://User:Pass@github.com/repo/path.git?Shallow=true", "https://User:Pass@bitbucket.com/repo/path.git?Shallow=true", "https://User:Pass@gitlab.com/repo/path.git?Shallow=true"}
+      paths.each_with_index do |path, index|
+        GitResolver.normalize_key_source("git", path).should eq({"git", results[index]})
+      end
```
straight-shoota, line 45: Such programmatical expanded code makes the examples unnecessarily hard to understand. This is overengineering. If you just use plain expecations, it much more clearly expresses what's going on.
```suggestion
      GitResolver.normalize_key_source("github", "repo/path").should eq({"git", "https://github.com/repo/path"})
      GitResolver.normalize_key_source("github", "rEpo/pAth").should eq({"git", "https://github.com/repo/path"})
      GitResolver.normalize_key_source("github", "REPO/PATH").should eq({"git", "https://github.com/repo/path"})
      GitResolver.normalize_key_source("bitbucket", "repo/path").should eq({"git", "https://bitbucket.com/repo/path"})
      # GitResolver.normalize_key_source("bitbucket", "rEpo/pAth").should eq({"git", "https://bitbucket.com/repo/path"})
      # GitResolver.normalize_key_source("bitbucket", "REPO/PATH").should eq({"git", "https://bitbucket.com/repo/path"})
      GitResolver.normalize_key_source("gitlab", "repo/path").should eq({"git", "https://gitlab.com/repo/path"})
      # GitResolver.normalize_key_source("gitlab", "rEpo/pAth").should eq({"git", "https://gitlab.com/repo/path"})
      # GitResolver.normalize_key_source("gitlab", "REPO/PATH").should eq({"git", "https://gitlab.com/repo/path"})

      # also normalise full git paths
      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Github.com/Repo/Path.git?Shallow=true").should eq "https://User:Pass@github.com/repo/path.git?Shallow=true"
      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Bitbucket.com/Repo/Path.Git?Shallow=true"). should eq "https://User:Pass@bitbucket.com/repo/path.git?Shallow=true"
      GitResolver.normalize_key_source("git", "HTTPS://User:Pass@Gitlab.com/Repo/Path?Shallow=true").should eq "https://User:Pass@gitlab.com/repo/path.git?Shallow=true"
```

We probably don't need full validation for every combination between path style and host. As long as the behaviour is identical beween hosts, skipping a couple of the duplicated tests is fine (I've included them in the diff but commented out; they can be removed).

This is not a requirement, but I'd really prefer this.

src/resolvers/git.cr
```diff
@@ -103,9 +103,19 @@ module Shards
     def self.normalize_key_source(key : String, source : String) : {String, String}
       case key
       when "git"
-        {"git", source}
+        uri = URI.parse(source)
+        downcased_host = uri.host.try &.downcase
+        if downcased_host && downcased_host.in?("github.com", "bitbucket.com", "gitlab.com")
```
straight-shoota, line 108: `www.` subdomains should also be included here.

src/resolvers/git.cr
```diff
@@ -103,9 +103,19 @@ module Shards
...
+          uri.scheme = "https"
```
straight-shoota, line 109: I'm not sure changing schemes is a good idea.

There could be a reason you need to access via http protocol instead of https. I guess it doesn't matter much though because in practice those are just redirects.

Sija: Yeah, but if user depends on invalid `http` based path it should raise instead of silently change it to the working version.

src/resolvers/git.cr
stakach: done

src/resolvers/git.cr
bcardiff: Let's keep the protocol so we don't prevent the user to use `http://` for whatever reason and play it safe for now in that regard.

The initial issue is the owner/repo name, so it's still an improvement.

asterite:
Just my two cents, but maybe it would have been better to fix this issue just for github. The original PR was a one line change.

src/resolvers/git.cr
stakach: I've limited the protocol changes so people can use `ssh` for private repos

stakach:
@asterite I would have agreed with you before implementing @bcardiff suggestion
it feels more complete / robust this way

src/resolvers/git.cr
stakach: remembering that these services are known quantities and you can't actually clone over HTTP

straight-shoota (approved):

bcardiff (approved):

src/resolvers/git.cr
```diff
@@ -103,9 +103,20 @@ module Shards
     def self.normalize_key_source(key : String, source : String) : {String, String}
       case key
       when "git"
-        {"git", source}
+        uri = URI.parse(source)
+        downcased_host = uri.host.try &.downcase
+        scheme = uri.scheme.try &.downcase
+        if scheme && scheme.in?("git", "http", "https") && downcased_host && downcased_host.in?("www.github.com", "github.com", "www.bitbucket.com", "bitbucket.com", "www.gitlab.com", "gitlab.com")
```
Sija, line 109: These domain names would be happier as a private constant instead.

Sija, line 109:
```suggestion
        if scheme.in?("git", "http", "https") && downcased_host && downcased_host.in?("www.github.com", "github.com", "www.bitbucket.com", "bitbucket.com", "www.gitlab.com", "gitlab.com")
```

src/resolvers/git.cr
Sija: > I've limited the protocol changes so people can use `ssh` for private repos

```suggestion
```

It's still there.

Sija (changes requested):

src/resolvers/git.cr
```diff
@@ -100,12 +100,25 @@ module Shards
       "git"
     end
 
+    private KEY_HOSTS = {"www.github.com", "github.com", "www.bitbucket.com", "bitbucket.com", "www.gitlab.com", "gitlab.com"}
```
Sija, line 103:
```suggestion
    private CASE_INSENSITIVE_HOSTS = {"www.github.com", "github.com", "www.bitbucket.com", "bitbucket.com", "www.gitlab.com", "gitlab.com"}
```

src/resolvers/git.cr
```diff
@@ -100,12 +100,25 @@ module Shards
...
+
     def self.normalize_key_source(key : String, source : String) : {String, String}
       case key
       when "git"
-        {"git", source}
+        uri = URI.parse(source)
+        downcased_host = uri.host.try &.downcase
+        scheme = uri.scheme.try &.downcase
+        if scheme.in?("git", "http", "https") && downcased_host && downcased_host.in?(KEY_HOSTS)
```
Sija, line 111:
```suggestion
        if scheme.in?("git", "http", "https") && downcased_host.in?(CASE_INSENSITIVE_HOSTS)
```

src/resolvers/git.cr
```diff
@@ -100,12 +100,25 @@ module Shards
...
+          uri.scheme = "https"
+          downcased_path = uri.path.downcase
+          uri.path = downcased_path.ends_with?(".git") ? downcased_path : "#{downcased_path}.git"
```
Sija, line 114: Are you sure it works with all providers (and not only GitHub)?

src/resolvers/git.cr
Sija: This will make impossible to use `git` protocol, I'm not sure why would you like to forbid that?

src/resolvers/git.cr
```diff
@@ -100,12 +100,25 @@ module Shards
...
+          uri.host = downcased_host.gsub("www.", "")
```
Sija, line 115: ditto, again I don't think we should transform the path in this way, as it will hide the errors on the user side. This PR is concerned about resolving repositories in a case insensitive manner, not "fixing" the repo paths for users.

Sija (commented):
I believe the changes here should be scoped only to the 3 resolvers: `github`, `gitlab` and `bitbucket`. All of the logic for `git` resolver is a "magic" in charge of _fixing user mistakes_ - which IMO shouldn't be the case at all - user is free to choose the resolver, and with `git` it will simply have to be correct, as that's the nature of generic `git` - or for that matter `http` - paths.

src/resolvers/git.cr
stakach: I prefer key hosts, as we are matching the hosts against the key sources

src/resolvers/git.cr
stakach: yes, when you use `bitbucket: org/repo` it adds the .git

src/resolvers/git.cr
stakach: github, gitlab and bitbucket all default their clone URLs without www and redirect to to this version of the URL and enforce HTTPS transfer.

So by doing this not only are we preventing potential clashes, we're also improving performance.

image

image

stakach:
@Sija we're only normalising github, bitbucket and gitlab paths, these services and their behaviour are well known and the path is being normalised to the one they recommend when you click the clone button on their webpage.

Not only will this improve performance, as a HTTP to HTTPS redirect will occur if you are cloning from the service
it also improves security, preventing man in the middle in case someone entered HTTP instead of HTTPS (thinking of the recent SolarWinds attack where a build service was hijacked)

src/resolvers/git.cr
Sija: That's correct, but I don't get why want you handle this within generic `git` resolver?

Sija:
> [...] it also improves security, preventing man in the middle in case someone entered HTTP instead of HTTPS (thinking of the recent SolarWinds attack where a build service was hijacked)

1. We're talking about the `git` resolver, not the known provider specific ones.
2. No it won't, this kind of _mistakes_ are made and will be made even more if you "silently" start fixing the insecure URLs. I'd rather display big fat warning instead so the user is made aware about the situation.

stakach:
I personally think it's fine as we're only fixing up well known services.
That said, I don't particularly care either way as I assume most people will always use the specific service resolver, at least based on what I've seen in the community

Sija:
@stakach I'm also confused about the SolarWinds remarks - what does _that_ have to do with any of this?

stakach:
https://www.zdnet.com/article/third-malware-strain-discovered-in-solarwinds-supply-chain-attack/

For example, lets say you have a popular app that might be a target of hackers
* I add a HTTP github URL in my shard.yml for an application I'm building
* a hacker breaks into my build server and notices that they can man-in-the-middle the build
* they edit the servers hosts file so they can inject whatever code they want without me knowing, as long as the builds succeed

was using the solarwinds example as a reference to show that this kind of attack does happen, getting around code signing

Sija:
> * I add a HTTP github URL in my shard.yml for an application I'm building

How would you do it, if the copy-pasta from every website you mentioned leads uses `https`? Would you remove the `s` by hand? How else would you end up with `http` protocol?

src/resolvers/git.cr
straight-shoota: Yes, all three providers work like that.

straight-shoota:
I think treating `git` URLs with well-known hosts the same way as the respective named resolver is good. That automatically removes any potential issues with non-matching sources. If some shard adds a dependency with `git:` and a github URL we don't want that to conflict with some other shard referencing the same dependdency with `github:` plus path. They should just be equivalent. That means if github.com resolves different request URLs to the same repository, shards should normalize them. I don't think there's a downside to that.

shardbox.org normalizes resolvers that way, too: https://github.com/shardbox/shardbox-core/blob/19fb582433bbd914ba31295c76cf411d322c2a71/src/repo/ref.cr#L22 (there's no downcasing the path, but it still works because the DB field is case-insensitive; should probably normalize that explicitly, though).

To address security concerns with unencrypted connections, we could consider to make `http` protocol an error for services known to only serve on HTTPS. But that's really a different topic.

src/resolvers/git.cr
straight-shoota: Maybe `KNOWN_PROVIDERS` would be a better alternative?

src/resolvers/git.cr
straight-shoota, line 115: This is more efficient:
```suggestion
          uri.host = downcased_host.lstrip("www.")
```

asterite: It should actually be `lchop`. I have no idea why `lstrip` exists, it looks pretty dangerous! We should remove that immediately from the language. Same with `String#delete(string)`.

asterite: (says the guy who seems to have implemented this... ouch!)

straight-shoota: Argh, yeah I've tripped over this before 😞

src/resolvers/git.cr
```diff
@@ -112,7 +112,7 @@ module Shards
           uri.scheme = "https"
           downcased_path = uri.path.downcase
           uri.path = downcased_path.ends_with?(".git") ? downcased_path : "#{downcased_path}.git"
-          uri.host = downcased_host.gsub("www.", "")
+          uri.host = downcased_host.lstrip("www.")
```
asterite, line 115: Note that this is wrong, it should be `lchop`

stakach: thanks! updated

src/resolvers/git.cr
```diff
@@ -100,12 +100,26 @@ module Shards
...
+    private KNOWN_PROVIDERS = {"www.github.com", "github.com", "www.bitbucket.com", "bitbucket.com", "www.gitlab.com", "gitlab.com"}
+
     def self.normalize_key_source(key : String, source : String) : {String, String}
       case key
       when "git"
-        {"git", source}
+        uri = URI.parse(source)
+        downcased_host = uri.host.try &.downcase
+        scheme = uri.scheme.try &.downcase
+        if scheme.in?("git", "http", "https") && downcased_host && downcased_host.in?(KNOWN_PROVIDERS)
```
Sija, line 111:
```suggestion
        if scheme.in?("git", "http", "https") && downcased_host.in?(KNOWN_PROVIDERS)
```

stakach:
@Sija 

> What's the difference here between http and https? Why wouldn't the attacker use a https endpoint with Let's Encrypt certificate for instance?

Let's Encrypt will only generate a certificate for a domain if you can prove you control it, they just make the proving process super simple.
Because of this proof requirement HTTPS prevents man in the middle attacks and is the main reason the [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security#description) header exists and is present on all the major repository hosting platforms

src/resolvers/git.cr
stakach: I want to ensure `downcased_host` is not nil here as otherwise I would have to replace `downcased_host.lchop("www.")` with `downcased_host.try &.lchop("www.")`

bcardiff:
I think I would leave out `git://` from the normalization. There are some private repo that use `git: git@github.owner/repo.git` sources to use the SSH identity directly. Using https will force them to do workarounds like https://stackoverflow.com/a/64587959/30948 which seems more indirect than using the SSH identity.

I guess that using github: or git: with http[s] is the most common thing. @straight-shoota can you confirm from shardbox that? If we leave out the `git://` from the normalization the case sensitivity issues on those pattern will still be solved and it will not break existing workflows with private shards.

Other than that the PR looks good to me. I think it's fine to rewrite http to https for these known hosts.

@stakach don't you have private shards that would be affected by this change for example?

cc: @bararchy

stakach:
@bcardiff github and I believe the `git` protocol itself doesn't support authentication at all.
So no private repos will work using `git`

The accepted answer in that stackoverflow post mentions:

> git:// URLs are read only, and it looks like private repos do not allow this form of access.

to use ssh with authentication you need to specify the URL in the form of:
`git@github.com` or `ssh://git@github.com/` both of which will continue to work

straight-shoota:
> I guess that using github: or git: with http[s] is the most common thing. @straight-shoota can you confirm from shardbox that?

Yes. All dependencies known to shardbox with `git:` source actually use https scheme. Not even a single http URL.

straight-shoota (approved):

bcardiff (approved):

Issue #468, opened 2021-01-24
Postinstall on Windows


oprypin:
Shards usually specify `postinstall: make`. But on Windows that line means that `make.exe` must exist. Usually it is not globally available, of course, and even when it is, it's some quirky alternative implementation of `make`.

What I've been doing is adding a `make.cmd` file to my Windows-supporting libs and telling people to run that directly. Unfortunately, even though running just "`make`" in a CMD shell would invoke that `make.cmd` file, `StartProcess` (the current underlying implementation of `system()` in Crystal) doesn't do that (unconditionally expects `.exe` extension), so anyone installing a shard that needs `postinstall: make` on POSIX will always get an error on Windows.

It's not even possible to skip the `postinstall` step, or stop `shards` from deleting the lib files altogether on such a failure.

*Anyway*... Even though I've been [arguing that `system()` should not pick any shell on Windows](https://github.com/crystal-lang/crystal/issues/9030), specifically for Shards postinstall it would probably be useful to wrap the supplied command into `cmd /c`. *Maybe* I'm saying this just because it happens to make my `make.cmd` hack work. But in any case, I think this is strictly an upgrade from the current state.

Let me know if there are any concerns, or I'll send a PR doing that. Or, well, any other ideas are welcome. Seems hard without allowing OS-specific command lines.

bcardiff:
I'm not sure what make sense because I rarely use postinstall. When I do I use crystal alone instead of make.

What I know I would like is to have a `shards [install|update] --skip-postinstall` option and a `shards postinstall [--dry-run] [shard names]` . 

Although it does not offer any input in a more portable postinstall support, it does offer a workaround and a transition.

oprypin:
This is for libraries that have some part implemented in C. I may also not use make at the top level but something is needed to compile code after installation.

straight-shoota:
@bcardiff These options are great :+1:

But the postinstall hook should nevertheless work on any system. So running it with `cmd /c` seems like a good idea, not only if you're using `make`. This should also work with any script file in the repository where `script` is the posix variant and `script.cmd` for windows. Only difficulty is that posix scripts usually need to be invoked by an explicitly relative path (`./script`) but `cmd /c ./script` fails. Transforming paths to backslash separators should fix this, though: `cmd /c .\script` works.

oprypin:
> `cmd /c ./script` fails

Oh, that sucks. Hmmm. And the slash transform doesn't seem like a safe idea, we can't just replace everything.

[...]

PR #475, opened 2021-02-19
Add --skip-postinstall cli option to install and update

bcardiff:
Partially implements the suggestions of https://github.com/crystal-lang/shards/issues/468#issuecomment-767064550

The postinstall instructions will be logged with a trailing `(skipped)`.

Usually the postinstall will generate target binaries that are then linked/copied by `Package#install_executables`.

With this PR the target binaries will not exist and wont be linked/copied, which is fine.

To complete the idea https://github.com/crystal-lang/shards/issues/468#issuecomment-767064550 and decouple the commands for Windows environments in a comfortable way I am not sure if it's better to a) make the `shards postinstall --dry-run [shard names]` perform postinstall and install_executables or if b) postinstall and install-executables should be different commands. 

Either way that can be handled in a separate PR. This should be enough to download dependencies with postinstall in Windows.

### DIFF
--- a/spec/integration/update_spec.cr
+++ b/spec/integration/update_spec.cr
@@ -235,6 +235,15 @@ describe "update" do
     end
   end
 
+  it "can skipe postinstall with transitive dependencies" do
+    with_shard({dependencies: {transitive: "*"}}, {transitive: "0.1.0"}) do
+      output = run "shards update --no-color --skip-postinstall"
+      binary = install_path("transitive", Shards::Helpers.exe("version"))
+      File.exists?(binary).should be_false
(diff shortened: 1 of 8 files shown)


spec/integration/update_spec.cr
```diff
@@ -235,6 +235,15 @@ describe "update" do
     end
   end
 
+  it "can skipe postinstall with transitive dependencies" do
```
straight-shoota, line 238:
```suggestion
  it "skips postinstall with transitive dependencies" do
```

straight-shoota (approved):

PR #485, opened 2021-03-20
Run the main shard's `postinstall` script as well

### LINKED ISSUE #484: `postinstall` or other hook needed to build c library
I have a shard with no dependencies that builds a c library through `postinstall`.  Unfortunately `postinstall` never runs from `shards install` when using the shard directly.

This is causing minor pain in 2 areas:
1. Other developers can't clone a project and start working without special instructions.
 ```
  git clone project
  shards install
  crystal spec # FAILS: c library missing
```
2. Ci needs special steps or the library is never built.

Solution: an additional hook that always runs or run `postinstall` after installing other dependencies in the same order as the shard being listed as a dependency.

If neither solution is acceptable then:
* Why should there be an extra step when `postinstall` is automatic for deps?
* Why should `postinstall` be special when listed as a dep but not for the shard itself?
* It's bad UX and inconsistent.

In #305 @ysbaddaden indicates 1 hook should do everything and #319 `postinstall` is the appropriate hook for building non crystal dependencies.  If so then the missing `postinstall` call is probably an oversight.

### PR DESCRIPTION
**Pull request by Blacksmoke16**
Resolves #484.

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -718,18 +718,18 @@ describe "install" do
   end
 
   it "runs postinstall of the main shard" do
-    with_shard({postinstall: "make"}) do
+    with_shard({postinstall: "touch made.txt"}) do
       output = run "shards install --no-color"
       File.exists?(File.join(application_path, "made.txt")).should be_true
-      output.should contain("Postinstall of test: make")
+      output.should contain("Postinstall of test: touch made.txt")
     end
   end
 
   it "can skip postinstall of the main shard" do
-    with_shard({postinstall: "make"}) do
+    with_shard({postinstall: "touch made.txt"}) do
       output = run "shards install --no-color --skip-postinstall"
       File.exists?(File.join(application_path, "made.txt")).should be_false
-      output.should contain("Postinstall of test: make (skipped)")
+      output.should contain("Postinstall of test: touch made.txt (skipped)")
     end
   end
 
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -717,6 +717,22 @@ describe "install" do
     end
   end
 
+  it "runs postinstall of the main shard" do
+    with_shard({postinstall: "make"}) do
+      output = run "shards install --no-color"
+      File.exists?(File.join(application_path, "made.txt")).should be_true
+      output.should contain("Postinstall of test: make")
+    end
+  end
+
+  it "can skip postinstall of the main shard" do
+    with_shard({postinstall: "make"}) do
+      output = run "shards install --no-color --skip-postinstall"
+      File.exists?(File.join(application_path, "made.txt")).should be_false
+      output.should contain("Postinstall of test: make (skipped)")
+    end
+  end
+
   it "fails with circular dependencies" do
     create_git_repository "a"
     create_git_release "a", "0.1.0", {dependencies: {b: {git: git_path("b")}}}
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: spec/integration/install_spec.cr [hunk 1]
```diff
@@ -717,6 +717,22 @@ describe "install" do
     end
   end
 
+  it "runs postinstall of the main shard" do
+    with_shard({postinstall: "make"}) do
```
straight-shoota, line 721: Can `touch` just run directly? I don't think we need make for this.

```suggestion
    with_shard({postinstall: "touch made.txt"}) do
```

straight-shoota (approved):

straight-shoota (changes requested):
Can you please edit `shards.adoc` as well an remove "of dependencies" in the description of `--skip-postinstall`?

And the checked-in manpages need to be rebuilt after that.

straight-shoota (dismissed):

asterite (approved):

Blacksmoke16:
Oof, guess this needs a merge.  I can do that tonight.

beta-ziliani:
Hi @Blacksmoke16 we are releasing a new Shards version, and we discussed within the Core team which PRs to merge. This particular PR sparked discussions (that I will transfer to the issue), so we decided to postpone it until we solved those. Perhaps we will keep is as is, but we decided not to rush it for this release. Sorry for the late notice, and for the extra work that you put because of this.

src/script.cr
```diff
@@ -3,7 +3,16 @@ module Shards
     class Error < Error
     end
 
-    def self.run(path, command, script_name, dependency_name)
+    def self.run(path : String, command : String, script_name : String, dependency_name : String, skip : Bool) : Nil
+      if !skip
+        Log.info { "#{script_name.capitalize} of #{dependency_name}: #{command}" }
+        self.run(path, command, script_name, dependency_name)
```
Sija, line 9: Redundant `self`

```suggestion
        run(path, command, script_name, dependency_name)
```

src/spec.cr
```diff
@@ -213,6 +213,12 @@ module Shards
       @scripts ||= {} of String => String
     end
 
+    def postinstall(spec_path : String) : Nil
+      if command = self.scripts["postinstall"]?
+        Script.run spec_path, command, "postinstall", self.name, Shards.skip_postinstall?
+      end
```
Sija, line 219: ditto

```suggestion
      if command = scripts["postinstall"]?
        Script.run spec_path, command, "postinstall", name, Shards.skip_postinstall?
      end
```

Sija (commented):
Style-related suggestions

src/script.cr
Blacksmoke16: Redundant maybe, but it improves the readability of the code by removing the ambiguity of if it's a variable or method call.  Either way based on the discussion in #484 this PR is prob going to be closed anyway so it doesn't really matter.

Sija: There's no need for it, since you can't _call_ a variable, nor there's a variable named _run_ in the context mentioned.

Blacksmoke16:
@beta-ziliani Going back thru my older PRs. Did those extra discussions ever make it to this PR?

beta-ziliani:
Sorry for the super long delay about this one. We discussed it with the team, and there's a general agreement that postinstall is not really recommended. Therefore, continuing to develop it is going kind of in the opposite direction of what we'd like to see happening here.

We're happy to discuss this further if you want this is a mistake.

Blacksmoke16:
@beta-ziliani Nope it's fine. I personally don't have a use for this feature at the moment; so happy to just close for now.

PR #487, opened 2021-03-23
Versioning examples consider crystal 2.x changes in `shard.yml.adoc`

kachick:
I'm a newbie in crystal-lang and the tools around them. 😅 
So this PR might be just a question...

As far as I know, Crystal made many changes the specs so far.
So planning `2.0.0` will contain other many changes?
Then the versioning including `< 2.0.0` might be robust?
I saw `shards` specifying as it in own shard.yml https://github.com/crystal-lang/shards/blob/addc26a3f22fee9fccb01cbb3e8878bef02d4295/shard.yml#L15 👀 

I know `examples` does not mean `recommended`. But it often spread into wide, I think.
How about to write this in example?

### DIFF
--- a/docs/shard.yml.adoc
+++ b/docs/shard.yml.adoc
@@ -101,7 +101,7 @@ NOTE: If this property is not included it is treated as _< 1.0.0_.
 +
 [source,yaml]
 ----
-crystal: ">= 0.35.0"
+crystal: ">= 0.35.0, < 2.0.0"
 ----
 
 *dependencies*::
@@ -355,7 +355,7 @@ with some dependencies:
 ----
 name: shards
 version: 1.2.3
-crystal: '>= 0.35.0'
+crystal: '>= 0.35.0, < 2.0.0'
 
 authors:
 - Julien Portalier <julien@example.com>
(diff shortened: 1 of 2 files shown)


straight-shoota:
Well I guess that's where different philosphies collide.

I wouldn't place an upper restriction by default because it's likely that a shard's code will just work with 2.0 as well as it did with 1.0. Yes, 2.0 will bring some breaking changes, but the goal is still to keep that as minimal as possible. Crystal 2.0 won't be a completely different language and break every Crystal code out there.
Looking backwards, some of my shards written for ancient versions like 0.26.0 or so still work with 1.0. There have been a lot of breaking changes since then to the language and stdlib. But that doesn't matter if a shard doesn't use the broken features.

Currently we see a lot of issues with shard dependencies not installing for 1.0 just because they don't state to be compatibile. If you install with `--ignore-crystal-version`, everything works. But that shouldn't be the solution every time a new major Crystal release happens and shard maintainers don't get to update their shard dependencies quickly.

kachick:
Agreed with your comments! Thanks for the polite answer 🙏

PR #488, opened 2021-03-26
Improve manpage format

straight-shoota:
* Use title case for headlines (automatically transformed to uppercase in manpage transform)
* Uniformly use bold text for code highlights (backticks do not work)

These improvements affect both the manpage format and enable a decent HTML rendering (see https://github.com/crystal-lang/crystal-book/issues/476).

### DIFF
--- a/docs/shard.yml.adoc
+++ b/docs/shard.yml.adoc
@@ -4,28 +4,28 @@
 :man manual: File Formats
 :man source: shards {shards_version}
 
-== NAME
-shard.yml - metadata for projects managed by shards(1)
+== Name
+shard.yml - metadata for projects managed by *shards*(1)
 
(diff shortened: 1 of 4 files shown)


oprypin:
Regarding "backticks do not work" - I see that that is not represented in any way when viewing the manpage on the terminal. OK.

I don't think it's good to change all underlines to bold, because underlines seem to have a very specific meaning - some kind of literal sample -- actually very close to what backticks would've been used with. And the underline looks very good on man pages.

straight-shoota:
I know, it's not an optimal result. If it was feasible, I would go for any improvement. But I think that's the best we can get for a reasonable effort. IMO the manpage representation looks reasonable and consistenst.

PR #490, opened 2021-03-30
Let `shards build` error if no targets defined

straight-shoota:
If no `targets` are defined in `shard.yml`, the command `shards build` should fail with an error message instead of silently succeeding without actually building anything.

/cc https://forum.crystal-lang.org/t/trouble-with-building-shard/3122

### DIFF
--- a/spec/integration/build_spec.cr
+++ b/spec/integration/build_spec.cr
@@ -71,4 +71,19 @@ describe "build" do
       File.exists?(bin_path("app")).should be_false
     end
   end
+
+  it "errors when no targets defined" do
+    File.write File.join(application_path, "shard.yml"), <<-YAML
+      name: build
+      version: 0.1.0
+    YAML
+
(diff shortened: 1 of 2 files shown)


asterite:
What do you think of showing an example in the error message? It could check if src/name.cr exists and provide that as an example.

straight-shoota:
Not sure that's a good idea. I'd usually recommend *not* to use the shard name as target file to allow easily requiring it as a dependency without starting the app.

asterite:
Regardless, what do you think of showing how it's done in the error?

straight-shoota:
I don't think there is yet any error message with detailed instructions in shards. Typically it just says s.th. like "run `shards update` first". But it's more complex with `shard.yml` configuration.
Maybe we could reference `man shard.yml`. But it's bad UX if the manpage is missing.

asterite (approved):

PR #494, opened 2021-04-09
Add documentation for `shard.override.yml`

straight-shoota:
Follow-up to #422. Documents the override behaviour.


asterite (approved):
Looks good!

PR #515, opened 2021-06-30
[Makefile] Fix shard.lock recipe

straight-shoota:
The `shard.lock` recipe fails if `SHARDS=false` (i.e. building from scratch without a previously installed `shards` binary), we must skip `$SHARDS update` in that case.

see https://github.com/crystal-lang/crystal/issues/10861


beta-ziliani (approved):
Thanks

PR #517, opened 2021-07-01
Fix pass no-color and verbose flags to crystal build

### LINKED ISSUE #508: "--no-color" is broken when "shards build" fails
## Bug Report

There is no difference if "--no-color"  is present or absent in args during "shards build" failure:
image

BUT there is difference when compilation is successful:

image

```shell
❯ crystal -v
Crystal 1.0.0 (2021-03-22)

LLVM: 11.1.0
Default target: x86_64-apple-macosx
```
(ordinal "brew install crystal" on macOS BigSur 11.4, iMac 2020)

straight-shoota:
Resolves #508

### DIFF
--- a/src/commands/build.cr
+++ b/src/commands/build.cr
@@ -34,7 +34,7 @@ module Shards
           "-o", File.join(Shards.bin_path, target.name),
           target.main,
         ]
-        if Shards.colors? == false
+        unless Shards.colors?
           args << "--no-color"
         end
         if Shards::Log.level <= ::Log::Severity::Debug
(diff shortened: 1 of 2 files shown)


src/commands/build.cr
```diff
@@ -34,6 +34,12 @@ module Shards
           "-o", File.join(Shards.bin_path, target.name),
           target.main,
         ]
+        if Shards.colors? == false
```
Sija, line 37: Why not use `unless`?

```suggestion
        unless Shards.colors?
```

oprypin (approved):

PR #531, opened 2021-11-25
Add subdirectory support for git dependencies

matthewmcgarvey:
Provides a way to load a shard that is a subdirectory in a git repo. Allows for the possibility of "monorepos" with multiple shards. I know this has been discussed a lot, though I can't find a related issue at the moment, but I don't believe I've seen anyone discuss this way of solving it. This follows a pattern used in other package managers, but the one I was most recently referenced was [nim's package manager, nimble](https://github.com/nim-lang/nimble#package-urls). To reference a subdirectory of a github repo, you add a query param to the url with a key of `subdir` and a value of the directory to use (`?subdir=alpha`).

## Example

```yaml
alpha:
  git: https://github.com/matthewmcgarvey/cronorepo.git?subdir=alpha
  branch: main
```

## Technical

The way this works is by cloning the larger monorepo as normal, but then checking out the desired commit/branch/tag into a temp directory. From there, the subdirectory is copied over to the install path. I believe the temp directory is required because I did not find a way to checkout the ref into the install path directly with only the subdirectory and it be flattened as expected.

This `subdir` query param causes the clone to fail (I believe) so I also added code to remove it before cloning.

## Other Considerations

This code is not clean, and I didn't intend for it to be. There doesn't seem to be consensus on whether or not things like this even _should_ be done so I don't want to spend a bunch of time making the code perfect if it's going to be rejected on a foundational level.

This code _only_ supports git urls. This is for many of the same reasons as above. If it's acceptable and merged, I would like to move on to considering what it might look like to include this in the github shorter version instead of the full url but I don't want to get ahead of myself.

## Considered Alternative

I considered a different way of doing it that seems cleaner with our yaml files which was to have something like:

```yaml
alpha:
  git: https://github.com/matthewmcgarvey/cronorepo.git
  branch: main
  subdir: alpha
```

But we currently only handle one `Requirement` and it's not passed to the `Resolver` anyways. I think a good refactor in the future would be to simply pass the `Dependency` into the resolver instead of pieces but I'm sure there's tradeoffs.

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -231,6 +231,11 @@ module Shards
     def subdir
       uri = parse_uri(git_url)
       subdir = uri.query_params.fetch("subdir", "")
+      if subdir.blank? && git_url.includes?("?")
+        params = git_url.split("?").last
+        subdir = URI::Params.parse(params).fetch("subdir", "")
+      end
+      subdir
     end
 
     def local_path
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -144,26 +144,30 @@ module Shards
       update_local_cache
       ref = git_ref(version)
 
-      if file_exists?(ref, SPEC_FILENAME)
-        capture("git show #{Process.quote("#{ref.to_git_ref}:#{SPEC_FILENAME}")}")
+      if file_exists?(ref, spec_filename)
+        capture("git show #{Process.quote("#{ref.to_git_ref}:#{spec_filename}")}")
       else
-        Log.debug { "Missing \"#{SPEC_FILENAME}\" for #{name.inspect} at #{ref}" }
+        Log.debug { "Missing \"#{spec_filename}\" for #{name.inspect} at #{ref}" }
         nil
       end
+    end
+
+    def spec_filename
+      File.join(subdir, SPEC_FILENAME)
     end
 
     private def spec_at_ref(ref : GitRef, commit) : Spec
       update_local_cache
 
-      unless file_exists?(ref, SPEC_FILENAME)
-        raise Error.new "No #{SPEC_FILENAME} was found for shard #{name.inspect} at commit #{commit}"
+      unless file_exists?(ref, spec_filename)
+        raise Error.new "No #{spec_filename} was found for shard #{name.inspect} at commit #{commit}"
       end
 
-      spec_yaml = capture("git show #{Process.quote("#{ref.to_git_ref}:#{SPEC_FILENAME}")}")
+      spec_yaml = capture("git show #{Process.quote("#{ref.to_git_ref}:#{spec_filename}")}")
       begin
         Spec.from_yaml(spec_yaml)
       rescue error : Error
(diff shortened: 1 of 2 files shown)


src/resolvers/git.cr
```diff
@@ -144,26 +144,30 @@ module Shards
...
+        raise Error.new "No #{spec_filename} was found for shard #{name.inspect} at commit #{commit}"
       end
 
-      spec_yaml = capture("git show #{Process.quote("#{ref.to_git_ref}:#{SPEC_FILENAME}")}")
+      spec_yaml = capture("git show #{Process.quote("#{ref.to_git_ref}:#{spec_filename}")}")
       begin
         Spec.from_yaml(spec_yaml)
       rescue error : Error
-        raise Error.new "Invalid #{SPEC_FILENAME} for shard #{name.inspect} at commit #{commit}: #{error.message}"
+        raise Error.new "Invalid #{spec_filename} for shard #{name.inspect} at commit #{commit}: #{error.message}"
```
matthewmcgarvey, line 170: I went ahead and changed all of these uses of the SPEC_FILENAME constant as I was working without considering the context but looking at it now, it seems like in these messages it would be better without the subdirectory name in it. 🤔

src/resolvers/git.cr
```diff
@@ -214,13 +218,26 @@ module Shards
...
     end
 
     def commit_sha1_at(ref : GitRef)
       capture("git log -n 1 --pretty=%H #{Process.quote(ref.to_git_ref)}").strip
     end
 
+    def subdir
+      uri = parse_uri(git_url)
+      subdir = uri.query_params.fetch("subdir", "")
+      if subdir.blank? && git_url.includes?("?")
```
matthewmcgarvey, line 234: This was specifically for the specs where it uses the file-based git repos. `URI` doesn't recognize the query params used there so some manual intervention is required.

src/resolvers/git.cr
```diff
@@ -316,8 +333,9 @@ module Shards
       # be used interactively.
       # This configuration can be overriden by defining the environment
       # variable `GIT_ASKPASS`.
-      git_retry(err: "Failed to clone #{git_url}") do
-        run_in_current_folder "git clone -c core.askPass=true --mirror --quiet -- #{Process.quote(git_url)} #{Process.quote(local_path)}"
+      url = git_url.split("?").first
```
matthewmcgarvey, line 336: This might need to be fixed if there _ARE_ query params that can be used in a clone. I think I've seen `shallow` used in the specs

straight-shoota:
This has previously been proposed and rejected in #238. I don't see how this PR does anything significantly differently or that any of the arguments has changed. The 1 shard = 1 repo doctrine is still standing strong.

I'm going to close this PR for now because we don't even need to look at any implementation until there's at least a reasonable consensus for adopting this. In order to reach that, I suggest to prepare a more elaborate RFC which summarizes the previous discussions, explains and disputes the arguments and finally makes a case why the previous decision should be revised.

PR #533, opened 2021-11-29
Add make `build recipe`

straight-shoota:
`make build` is commonly established as a recipe for building. I tried to run `make clean build` (which I do a lot in other repositories) and it didn't work because `build` is missing. This patch adds it.


beta-ziliani (approved):
LGTM

PR #534, opened 2021-11-30
Honour `CRYSTAL` env var

### LINKED ISSUE #406: Resolution of crystal dependency is hard tied to `crystal` executable

straight-shoota:
This is a reissue of #415 reduced to the original scope. It makes `shards` honour the `CRYSTAL` environment variable when calling the `crystal` program to determine the Crystal version and for `shards build`.

Resolves #406

### DIFF
--- a/spec/integration/spec_helper.cr
+++ b/spec/integration/spec_helper.cr
@@ -106,7 +106,7 @@ private def setup_repositories
   create_git_release "transitive", "0.2.0", {
     dependencies: {version: {git: git_url(:version)}},
     scripts:      {
-      postinstall: %(${CRYSTAL:-crystal} build src/version.cr),
+      postinstall: %(#{{{ flag?(:win32) ? "crystal" : "${CRYSTAL:-crystal}" }}} build src/version.cr),
     },
   }
 
--- a/spec/integration/spec_helper.cr
+++ b/spec/integration/spec_helper.cr
@@ -106,7 +106,7 @@ private def setup_repositories
   create_git_release "transitive", "0.2.0", {
     dependencies: {version: {git: git_url(:version)}},
     scripts:      {
-      postinstall: "crystal build src/version.cr",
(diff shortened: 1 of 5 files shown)


oprypin:
Specs fail on Windows

straight-shoota:
Yeah, that was #532

oprypin:
Seems not ¯\\\_(ツ)_/¯

oprypin:
Seems not ¯\\\_(ツ)_/¯

straight-shoota:
Yeah, now we get the actual failures on windows. Not the unrelated ones.

Sija:
Windows doesn't like `${CRYSTAL:-crystal}` within the command.

oprypin (approved):

PR #536, opened 2021-12-15
No longer depend of external git user config

luislavena:
Some Git-related tests required external configuration of email and name. This forced tweaks in CI and other environments to have those set globally before running tests.

This change makes this self-contained: Git repositories created part of the test suite are going to automatically set a default email and name.

Note: this might look silly, but removes the need to set that up when running the specs in a controlled container environment (and not a development one) and has no impact on existing CI or others running it.

Thank you.
❤️ ❤️ ❤️

### DIFF
--- a/spec/support/factories.cr
+++ b/spec/support/factories.cr
@@ -18,6 +18,9 @@ def create_git_repository(project, *versions)
     run "git init #{Process.quote(project)}"
     Dir.cd(Process.quote(project)) do
       run "git checkout --orphan master"
+
+      run "git config user.email author@example.com"
+      run "git config user.name Author"
     end
   end
 
@@ -34,6 +37,11 @@ end
 def create_fork_git_repository(project, upstream)
   Dir.cd(tmp_path) do
     run "git clone #{Process.quote(git_url(upstream))} #{Process.quote(project)}"
+
+    Dir.cd(Process.quote(project)) do
+      run "git config user.email fork@example.com"
+      run "git config user.name Fork"
+    end
   end
 end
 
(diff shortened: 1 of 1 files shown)


spec/support/factories.cr
```diff
@@ -34,6 +37,11 @@ end
 def create_fork_git_repository(project, upstream)
   Dir.cd(tmp_path) do
     run "git clone #{Process.quote(git_url(upstream))} #{Process.quote(project)}"
+
+    Dir.cd(Process.quote(project)) do
```
oprypin, line 41: `Process.quote` is invalid here. Oh oops, there's actually another invalid instance of it already checked in. Maybe let's just remove both instances here?
```suggestion
    Dir.cd(project) do
```

luislavena: Thank you @oprypin for pointing it out, looking a bit deeper into the code found that it should be `Dir.cd(git_path(project))` instead, to be consistent with the other usage across the test codebase.

Cheers.
❤️ ❤️ ❤️

straight-shoota (approved):

spec/support/factories.cr
oprypin: It's actually not necessary because the `git_path` part is already covered by `Dir.cd(tmp_path)` above, but fine either way.

oprypin (approved):

spec/support/factories.cr
luislavena: Yeah, thought that, but to be consistent on how is being used in several places: 

https://github.com/crystal-lang/shards/blob/master/spec/support/factories.cr#L40-L54

Thought will be good to be consistent 😊

PR #538, opened 2021-12-31
re-enabled ~ support in path resolver

### LINKED ISSUE #308: Doesn't support ~ in path

masukomi:
Resolves #308

[as per suggestion](https://github.com/crystal-lang/shards/issues/308#issuecomment-569764134) of @ysbaddaden 

⚠️ WARNING: I just did what was suggested. It seems to compile, but I did not actually unit test this because _most_ of the unit tests fail for me without any changes on my part.

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -914,7 +914,7 @@ describe "install" do
   end
 
   it "expands path and shows in debug info if missing" do
-    metadata = {dependencies: {nonexistent: { path: "~/nonexistent-path" }}}
+    metadata = {dependencies: {nonexistent: {path: "~/nonexistent-path"}}}
     with_shard(metadata) do
       ex = expect_raises(FailedCommand) { run "shards install --no-color -v" }
       ex.stdout.should contain(%(E: Failed no such path: #{Path.home.join("nonexistent-path")}))
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -913,6 +913,14 @@ describe "install" do
     end
   end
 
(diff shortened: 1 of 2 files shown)


straight-shoota:
I added a spec and fixed the implementation `File.expand_path` requires tilde expansion explicitly enabled (`home: true`).

straight-shoota:
Note: Besides enabling support for tilde expansion, this also improves the error message for nonexistent paths. When the path doesn't exist, you currently get the error `Missing "shard.yml" for "nonexistent"`. With this change it will be `Failed no such path: non/existent/path`.

straight-shoota (approved):

beta-ziliani (approved):
LGTM

PR #539, opened 2022-01-06
Add `--jobs` (parallel git fetch)

### LINKED ISSUE #239: Run git fetches and post installs in parallel
if possible run git fetches and post installs in parallel especially for running tests in a ci this would be helpful as the dependency fetching and building can take some time

### LINKED ISSUE #371: Non-blocking dependencies fetching
Shards fetches each dependency - one at a time. This is a slow process, because each operation blocks the other one. After the dependencies resolution, each download is totally independent of the others.
The speed up gain can be considerable bigger the list is.

On the good side, installing the libraries is fast. Parallelizing it wont be much beneficial, especially since the operations are mainly IO bound (filesystem copy-pasting).

m-o-e:
Resolves #371
Resolves #239

This PR parallelizes git fetching which gives me a ~40% speed-up
for `shards install` on average.

```
$ git clone https://github.com/amberframework/amber.git
$ cd amber

$ time shards --jobs=1
18.579s total

$ time shards --jobs=8
10.143s total
```

* Only the fetching is parallelized, post-install still runs in serial.
   I've experimented with parallelizing that too but the speed gains
   were negligible and I'm not sure if it's always safe.

* It's [only implemented for git](https://github.com/crystal-lang/shards/commit/886fdbd9a0e14104d201c234922f967bc5d593e9) so far (no hg)

* **No test-coverage, yet**!
  Although parallelism is enabled during test runs, the existing specs don't seem to actually exercise it.

### DIFF
--- a/docs/shards.adoc
+++ b/docs/shards.adoc
@@ -74,7 +74,7 @@ doesn't generate a conflict, thus generating a new _shard.lock_ file.
 --production:: same as _--frozen_ and _--without-development_
 --skip-postinstall:: Does not run postinstall of dependencies.
 --skip-executables:: Does not install executables.
---jobs:: Number of git fetches to perform in parallel (default: 8).
+--jobs:: Number of repository downloads to perform in parallel (default: 8). Currently only for git.
 --
 
 *list* [--tree]::
--- a/docs/shards.adoc
+++ b/docs/shards.adoc
@@ -58,7 +58,7 @@ Exit status:
 *init*::
 Initializes a default _shard.yml_ in the current folder.
 
-*install* [--frozen] [--without-development] [--production] [--skip-postinstall] [--skip-executables]::
+*install* [--frozen] [--without-development] [--production] [--skip-postinstall] [--skip-executables] [--parallel-fetch=N]::
 Resolves and installs dependencies into the _lib_ folder. If not already
 present, generates a _shard.lock_ file from resolved dependencies, locking
 version numbers or Git commits.
@@ -74,6 +74,7 @@ doesn't generate a conflict, thus generating a new _shard.lock_ file.
 --production:: same as _--frozen_ and _--without-development_
 --skip-postinstall:: Does not run postinstall of dependencies.
 --skip-executables:: Does not install executables.
+--parallel-fetch:: Number of git fetches to perform in parallel (default: 8).
 --
 
 *list* [--tree]::
--- a/docs/shards.adoc
+++ b/docs/shards.adoc
@@ -58,7 +58,7 @@ Exit status:
 *init*::
 Initializes a default _shard.yml_ in the current folder.
 
-*install* [--frozen] [--without-development] [--production] [--skip-postinstall] [--skip-executables] [--parallel-fetch=N]::
+*install* [--frozen] [--without-development] [--production] [--skip-postinstall] [--skip-executables] [--jobs=N]::
 Resolves and installs dependencies into the _lib_ folder. If not already
 present, generates a _shard.lock_ file from resolved dependencies, locking
 version numbers or Git commits.
@@ -74,7 +74,7 @@ doesn't generate a conflict, thus generating a new _shard.lock_ file.
 --production:: same as _--frozen_ and _--without-development_
 --skip-postinstall:: Does not run postinstall of dependencies.
 --skip-executables:: Does not install executables.
---parallel-fetch:: Number of git fetches to perform in parallel (default: 8).
+--jobs:: Number of git fetches to perform in parallel (default: 8).
 --
 
 *list* [--tree]::
--- a/src/config.cr
+++ b/src/config.cr
@@ -108,6 +108,5 @@ module Shards
   class_property? skip_postinstall = false
   class_property? skip_executables = false
 
-  # Defaults to 1 (disabled) when running in `crystal spec`
-  class_property parallel_fetch : Int32 = {{ @top_level.has_constant?("Spec") ? 1 : 8 }}
+  class_property parallel_fetch : Int32 = 8
 end
--- a/src/config.cr
+++ b/src/config.cr
@@ -9,10 +9,11 @@ module Shards
   DEFAULT_COMMAND = "install"
   DEFAULT_VERSION = "0"
 
-  VERSION_REFERENCE     = /^v?\d+[-.][-.a-zA-Z\d]+$/
-  VERSION_TAG           = /^v(\d+[-.][-.a-zA-Z\d]+)$/
-  VERSION_AT_GIT_COMMIT = /^(\d+[-.][-.a-zA-Z\d]+)\+git\.commit\.([0-9a-f]+)$/
-  VERSION_AT_HG_COMMIT  = /^(\d+[-.][-.a-zA-Z\d]+)\+hg\.commit\.([0-9a-f]+)$/
+  VERSION_REFERENCE        = /^v?\d+[-.][-.a-zA-Z\d]+$/
+  VERSION_TAG              = /^v(\d+[-.][-.a-zA-Z\d]+)$/
+  VERSION_AT_GIT_COMMIT    = /^(\d+[-.][-.a-zA-Z\d]+)\+git\.commit\.([0-9a-f]+)$/
+  VERSION_AT_HG_COMMIT     = /^(\d+[-.][-.a-zA-Z\d]+)\+hg\.commit\.([0-9a-f]+)$/
+  VERSION_AT_FOSSIL_COMMIT = /^(\d+[-.][-.a-zA-Z\d]+)\+fossil\.commit\.([0-9a-f]+)$/
 
   def self.cache_path
     @@cache_path ||= find_or_create_cache_path
--- a/src/config.cr
+++ b/src/config.cr
@@ -107,4 +107,7 @@ module Shards
   class_property? local = false
   class_property? skip_postinstall = false
   class_property? skip_executables = false
+
+  # Defaults to 1 (disabled) when running in `crystal spec`
+  class_property parallel_fetch : Int32 = {{ @top_level.has_constant?("Spec") ? 1 : 8 }}
 end
--- a/src/config.cr
+++ b/src/config.cr
@@ -108,5 +108,5 @@ module Shards
   class_property? skip_postinstall = false
   class_property? skip_executables = false
 
-  class_property parallel_fetch : Int32 = 8
+  class_property jobs : Int32 = 8
 end
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -51,19 +51,19 @@ module Shards
       return unless Shards.jobs > 1
 
       count = 0
-      active = 0
+      active = Atomic.new(0)
       ch = Channel(Exception?).new(deps.size + 1)
       deps.each do |dep|
         count += 1
-        active += 1
-        while active > Shards.jobs
+        active.add(1)
+        while active.get > Shards.jobs
           sleep 0.1
         end
         spawn do
           begin
             dep.resolver.update_local_cache if dep.resolver.is_a? GitResolver
             ch.send(nil)
-            active -= 1
+            active.sub(1)
           rescue ex : Exception
             ch.send(ex)
           end
--- a/src/molinillo_solver.cr
+++ b/src/molinillo_solver.cr
@@ -61,7 +61,7 @@ module Shards
         end
         spawn do
           begin
-            dep.resolver.update_local_cache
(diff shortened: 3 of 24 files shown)


src/resolvers/git.cr
```diff
@@ -425,7 +425,11 @@ module Shards
 
       output = capture ? IO::Memory.new : Process::Redirect::Close
       error = IO::Memory.new
-      status = Process.run(command, shell: true, output: output, error: error)
+      if path
+        status = Process.run(command, shell: true, output: output, error: error, chdir: path)
+      else
+        status = Process.run(command, shell: true, output: output, error: error)
+      end
```
straight-shoota, line 432: You should be able to directly pass a nilable value for `chdir` and `Process.run` whil change the directory when present.
```suggestion
      status = Process.run(command, shell: true, output: output, error: error, chdir: path)
```

straight-shoota:
It would be great to have specs for parallel fetch.

src/config.cr
```diff
@@ -107,4 +107,7 @@ module Shards
   class_property? local = false
   class_property? skip_postinstall = false
   class_property? skip_executables = false
+
+  # Defaults to 1 (disabled) when running in `crystal spec`
+  class_property parallel_fetch : Int32 = {{ @top_level.has_constant?("Spec") ? 1 : 8 }}
```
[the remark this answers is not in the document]
m-o-e: The idea was to avoid potential hangs / undefined behaviour when tests run in parallel.

But your comment below made me realize that was misguided for a whole
bunch of reasons (no least: afaik Spec doesn't even have concurrency [[yet?](https://github.com/crystal-lang/crystal/issues/8289)]). 

**tdlr:** I've removed that special behaviour, it now defaults to 8 unconditionally.

m-o-e:
> It would be great to have specs for parallel fetch.

Agree.

Now that it's enabled in spec it should get plenty coverage.
Do you think that is sufficient or should there be an isolated test?
(not sure how to create a meaningful one tbh, ideas welcome 🤔)

src/molinillo_solver.cr
```diff
@@ -47,7 +47,38 @@ module Shards
...
+            active -= 1
+          rescue ex : Exception
+            ch.send(ex)
+          end
+        end
+      end
+
+      count.times do
+        obj = ch.receive
+        raise obj if obj.is_a? Exception
```
Sija, line 75: This will raise only 1st `Exception`.
How about combining all found ones into a separate exception class and raising that instead?

m-o-e: Doable but would feel like overkill to me?
When one of the fetches raises (usually due to a network error) then we have to abort anyway.

Not opposed to changing it, but skeptical it's worth the addtl code complexity.

straight-shoota: I think it's okay to leave it like that. Showing all exceptions at once is useful when they have *different* causes (for example, misspelled names of two dependencies), but not when they have the same cause (for example, not network connection).

It's probably more likely that a global error affects multiple tracks than two independent errors showing up at the same time. The latter can happen of course, but then you just need to fix them sequentially. I think that's better than repeating messages for the same error.
This is in line with the current behaviour where sequential resolving fails at the first error as well.

We can consider further improvements, for example collecting all errors and then deciding based on the type of errors, which ones should be printed. But that's for a follow-up and IMO not a high priority.

m-o-e:
After using this PR-version for a while I realized it would only parallelize
when a `shard.lock` exists, but not for the initial run after cloning a repo
without that file.

This is now fixed.

(had missed it in my previous testing because I'd muck
with `shard.yml`, delete `lib`, but never the `shard.lock`...)

src/molinillo_solver.cr
```diff
@@ -47,7 +47,36 @@ module Shards
...
+      end
+    end
+
     private def add_lock(base, lock_index, deps : Array(Dependency))
+      prefetch_local_caches(deps) if Shards.parallel_fetch > 1
```
Sija, line 78: Using early return within `prefetch_local_caches` instead of in the callsite would DRY things up a bit.

```cr
return unless Shards.parallel_fetch > 1
```

m-o-e: Good point, fixed in https://github.com/crystal-lang/shards/pull/539/commits/f3a30698ccb71c0f796d4931413f35c518e3eeb0

docs/shards.adoc
```diff
@@ -74,6 +74,7 @@ doesn't generate a conflict, thus generating a new _shard.lock_ file.
 --production:: same as _--frozen_ and _--without-development_
 --skip-postinstall:: Does not run postinstall of dependencies.
 --skip-executables:: Does not install executables.
+--parallel-fetch:: Number of git fetches to perform in parallel (default: 8).
```
Sija, line 77: IIUC this works for every resolver (implementing `update_local_cache`), not just _git_?
If so, I'd rename it to `parallel-jobs` or sth like that.

m-o-e: I thought `fetch` would make the most sense for all resolvers where this applies
since it's about the data-fetching and nothing else - each VCS might call that step
something different but ultimately it's a "fetch".

It's only implemented for git atm, that's why I mentioned that one in the description.

We could change it but to me `parallel-jobs` feels less to the point than `parallel-fetch`? 🤔

Sija: > It's only implemented for git atm, that's why I mentioned that one in the description.

I don't follow. You're calling `Resolver#update_local_cache` for every resolver regardless, so how come you state it's implemented only for git?

m-o-e: Heck... you are right.

I got confused because my first iteration was git-specific,
I totally forgot that it now indeed applies to all of them. 🙈 

Ouch, that means this PR probably [breaks hg](https://github.com/crystal-lang/shards/blob/f3a30698ccb71c0f796d4931413f35c518e3eeb0/src/resolvers/hg.cr#L434).
The hg-side will need a change similar to [this](https://github.com/crystal-lang/shards/pull/539/files#diff-da83517c6d5d940ffa16fa8f176bbe55ac9b5ef3abafe8336d35b48a6be171a1L414).

Sigh. This makes things more complicated.

I can port the change from git.cr to hg.cr - but I haven't wrapped my head
around the spec-suite enough to see how to make a test to validate it.

It looks like the current [integration test](https://github.com/crystal-lang/shards/blob/635d36af83b3adfbfbc8af88e5f00a3c5af5d585/spec/integration/install_spec.cr) only uses git URLs (no `hg_url()` in there).
That may be the reason why it didn't catch the breakage - or maybe
it just doesn't exercise the parallelism at all.

(I thought it does because it uses multiple deps in [some places](https://github.com/crystal-lang/shards/blob/635d36af83b3adfbfbc8af88e5f00a3c5af5d585/spec/integration/install_spec.cr#L344) -
but the fact that it didn't break for hg... confuses me 🤔)

m-o-e: Hm. The best I could offer for now would be to _actually_ exclude hg from
parallelism (like I thought it already was), by doing something like:

```diff
          begin
-           dep.resolver.update_local_cache
+           dep.resolver.update_local_cache if dep.resolver.is_a? GitResolver
            ch.send(nil)
            active -= 1
          rescue ex : Exception
            ch.send(ex)
          end
        end
```

But I feel like if the spec's didn't catch the hg issue then they
might also not sufficiently cover the git-side. So even though it works
splendidly for me under real usage, spec-coverage should probably
be addressed before merging this.

If someone more familiar with the spec-suite and/or hg
has an idea on how to best do that, please don't hesitate. 😬

m-o-e: > _actually_ exclude hg from parallelism

Fyi, this is now added here: https://github.com/crystal-lang/shards/pull/539/commits/886fdbd9a0e14104d201c234922f967bc5d593e9

m-o-e: Agree, but that will need test-coverage first.
For git I have some confidence that it works fine since I've been using it daily for a while, but
I've never encountered a hg-shard in the wild, so don't even have an URL to test that side manually.

Not sure when I'll have time to continue on this, so excluding Hg for now.

straight-shoota: That's totally fine. Concurrent fetches are just an optimization without functional differences. There's no need to have this implemented for all resolvers immediately.

src/resolvers/resolver.cr
```diff
@@ -87,6 +87,7 @@ module Shards
     abstract def read_spec(version : Version) : String?
     abstract def install_sources(version : Version, install_dir : String)
     abstract def report_version(version : Version) : String
+    abstract def update_local_cache
```
straight-shoota, line 90: I'm not sure if this makes sense as an abstract method when the concept doesn't even apply to all resolvers (`Path`, `Crystal` don't have a local cache). Instead, we could avoid empty implementations and guard the call to this method by `responds_to?`. Alternatively, we could add an empty implementation here and override it only in ancestors that have a meaningful implementation.

m-o-e: Hmm. True.

Changed to an empty def in https://github.com/crystal-lang/shards/pull/539/commits/7f45434b3e5f20dea105b01af7a086d122bcc3ac

(I think that's the more discoverable variant of the two, but no strong opinion here,
just let me know if you'd prefer `responds_to?` instead)

m-o-e:
Good idea @carlhoerberg!

Renamed in https://github.com/crystal-lang/shards/pull/539/commits/bac393b6cdcc19af02b84fe66027fdffad8421cb

Fwiw, I left the help-screen as `Number of git fetches to perform in parallel (default: 8)`
to keep it clear what it actually does. If anyone would prefer a different text just let me know. :)

docs/shards.adoc
m-o-e: Resolving this convo as I believe all points are addressed.

* Parallelism is restricted to only apply to GitResolver (because hg not implemented, yet)
* Flag has been renamed to `--jobs` to keep it generic (for consistency with `bundler`
  and for when we add hg in the future)
* Help-text still mentions git explicitly - but I believe that makes sense as to not
   confuse users into thinking it would apply to all resolvers.
   
Please re-open or reply on main thread if any of these need further discussion.

straight-shoota: I'd prefer to keep the wording more general, but mention that it only works for git currently. For example:

```
--jobs:: Number of repository downloads to perform in parallel (default: 8). Currently only functions for git resolver.
```

m-o-e: Ok, updated in https://github.com/crystal-lang/shards/pull/539/commits/670cfb749d6f1890074a72891d25ae71679c07d6

Dropped "function" and "resolver" to make it shorter.

straight-shoota (approved):

src/molinillo_solver.cr
```diff
@@ -47,7 +47,38 @@ module Shards
...
+        count += 1
+        active += 1
+        while active > Shards.jobs
+          sleep 0.1
+        end
+        spawn do
+          begin
+            dep.resolver.update_local_cache if dep.resolver.is_a? GitResolver
+            ch.send(nil)
+            active -= 1
```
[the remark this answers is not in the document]
beta-ziliani: Ah yes! Atomic is what we need here. Do you want to do the change? Let me know otherwise and we'll take care.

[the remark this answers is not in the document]
straight-shoota: I think this is fine. Shards is currently not intended to be built with multithreading. And I wouldn't even see a good reason to that.
And with any luck we should eventually get a tool in stdlib for abstracing such simple conccurrency workflows (cf. https://github.com/crystal-lang/crystal/issues/6468). There should really be no need to implement this from scratch every time you want to run a couple tasks concurrently.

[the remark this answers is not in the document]
m-o-e: Well, the change seems easy enough, so I've [added it here](https://github.com/crystal-lang/shards/commit/aa65be131e36fe4e839d6f3c2bdcd0accdebb0a7).
Better safe than sorry.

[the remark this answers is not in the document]
m-o-e: And in fact, another fix [right here](https://github.com/crystal-lang/shards/commit/60c778836893351ed5cd40ee1267c7f8b5ed0572).

I think this latter one could actually have struck even without MT.

src/molinillo_solver.cr
```diff
@@ -51,21 +51,22 @@ module Shards
...
           sleep 0.1
         end
         spawn do
           begin
             dep.resolver.update_local_cache if dep.resolver.is_a? GitResolver
             ch.send(nil)
-            active -= 1
           rescue ex : Exception
             ch.send(ex)
+          ensure
```
beta-ziliani, line 68: Nice catch!

beta-ziliani (approved):

Issue #535, opened 2021-12-03
Warnings are being swallowed (unless `--error-on-warnings` is on)


beta-ziliani:
Basically, the problematic line is: https://github.com/crystal-lang/shards/blob/1739c0fb96cf16e405bcdebf91357540d87243b3/src/commands/build.cr#L47

It is writing the error on a local memory object that is discarded unless there is an unsuccessful build.

Example:
 
```shell
➜  d git:(main) ✗ cat shard.yml
name: d
version: 1
targets:
  d:
    main: d.cr
➜  d git:(main) ✗ cat d.cr
@[Deprecated]
def a
end

a
➜  d git:(main) ✗ shards build
Dependencies are satisfied
Building: d
➜  d git:(main) ✗ crystal build d.cr
In d.cr:5:1

 5 | a
     ^
Warning: Deprecated top-level a.

A total of 1 warnings were found.
```

straight-shoota:
Yeah, I think passing STDERR through like STDOUT makes sense. `shards build` is supposed to be just a thin wrapper around `crystal build`.

beta-ziliani:
not as stderr? I was planning on forwarding it to stderr

straight-shoota:
Yeah, pass STDERR through to STDERR of course. I just compared it to STDOUT because there we already do that.

luislavena:
@beta-ziliani this means the raise exception should change?

https://github.com/crystal-lang/shards/blob/1739c0fb96cf16e405bcdebf91357540d87243b3/src/commands/build.cr#L48

Now STDERR is already part of the output so there is no `error` to use for that message.

Is that what you mean?

If that is the case, something like the following is what you would be expecting the output to be?

```
$ shards build
Dependencies are satisfied
Building: d
In d.cr:5:1

 5 | a
     ^
Warning: Deprecated top-level a.

A total of 1 warnings were found.
```

And when it errors, the last line will be `Error target d failed to compile.`

Thank you
❤️ ❤️ ❤️

beta-ziliani:
Sorry for the delay, @luislavena , yes, that's what I mean

luislavena:
No worries @beta-ziliani, thanks for responding!

I will be happy to send a PR with this change later this week 😊 

Cheers.
❤️ ❤️ ❤️ 

PS: Happy New Year! 🎉

beta-ziliani:
I'm finishing on one right now  :-) But I'll be happy to have your review. Happy new year for you too! 🎉

[...]

PR #540, opened 2022-01-10
Feat: output STDERR from the building process

beta-ziliani:
Solves #535 by outputting the STDERR of the `crystal` build to the STDERR of the `shards` process _only if the build is successful_. This allows to notice deprecation warnings and the like.

In order to test it, I had to concatenate the STDOUT and STDERR, which isn't as clean as possible but works.

### DIFF
--- a/spec/support/factories.cr
+++ b/spec/support/factories.cr
@@ -254,10 +254,12 @@ def run(command, *, env = nil, return_error = false)
   {% end %}
   status = Process.run(command, shell: true, env: cmd_env, output: output, error: error || Process::Redirect::Close)
 
+  output = output.to_s.gsub("\r\n", "\n")
+  error = error.to_s.gsub("\r\n", "\n")
+
   if status.success?
-    output = output.to_s.gsub("\r\n", "\n")
-    output + error.to_s
+    output + error
   else
-    raise FailedCommand.new("command failed: #{command}", output.to_s.gsub("\r\n", "\n"), error.to_s.gsub("\r\n", "\n"))
+    raise FailedCommand.new("command failed: #{command}", output, error)
   end
 end
--- a/spec/support/factories.cr
+++ b/spec/support/factories.cr
@@ -242,7 +242,7 @@ def tmp_path
   Shards::Specs.tmp_path
 end
 
-def run(command, *, env = nil, return_error = false)
+def run(command, *, env = nil)
   cmd_env = {
     "CRYSTAL_PATH" => Shards::Specs.crystal_path,
   }
--- a/spec/support/factories.cr
+++ b/spec/support/factories.cr
@@ -234,7 +234,7 @@ def tmp_path
   Shards::Specs.tmp_path
 end
 
-def run(command, *, env = nil)
+def run(command, *, env = nil, return_error = false)
   cmd_env = {
     "CRYSTAL_PATH" => Shards::Specs.crystal_path,
   }
@@ -247,7 +247,8 @@ def run(command, *, env = nil)
   status = Process.run(command, shell: true, env: cmd_env, output: output, error: error || Process::Redirect::Close)
 
   if status.success?
-    output.to_s.gsub("\r\n", "\n")
+    output = output.to_s.gsub("\r\n", "\n")
+    output + error.to_s
   else
     raise FailedCommand.new("command failed: #{command}", output.to_s.gsub("\r\n", "\n"), error.to_s.gsub("\r\n", "\n"))
   end
--- a/src/commands/build.cr
+++ b/src/commands/build.cr
@@ -45,10 +45,10 @@ module Shards
 
         error = IO::Memory.new
         status = Process.run(Shards.crystal_bin, args: args, output: Process::Redirect::Inherit, error: error)
-        unless status.success?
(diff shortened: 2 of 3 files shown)


spec/support/factories.cr
```diff
@@ -255,7 +255,8 @@ def run(command, *, env = nil)
   status = Process.run(command, shell: true, env: cmd_env, output: output, error: error || Process::Redirect::Close)
 
   if status.success?
-    output.to_s.gsub("\r\n", "\n")
+    output = output.to_s.gsub("\r\n", "\n")
+    output + error.to_s
```
straight-shoota, line 259: We should do the line feed normalization for error output as well. This can also be pulled out because it's the same in both conditional branches.

src/commands/build.cr
```diff
@@ -44,9 +44,12 @@ module Shards
         Log.debug { "#{Shards.crystal_bin} #{args.join(' ')}" }
 
         error = IO::Memory.new
-
         status = Process.run(Shards.crystal_bin, args: args, output: Process::Redirect::Inherit, error: error)
-        raise Error.new("Error target #{target.name} failed to compile:\n#{error}") unless status.success?
+        unless status.success?
+          raise Error.new("Error target #{target.name} failed to compile:\n#{error}")
+        else
+          STDERR << error if error.size > 0
```
straight-shoota, line 51: I'd suggest to use `puts` for the odd chance that `error` misses a trailing newline.
Size comparison can be replaced by `empty?`.
```suggestion
          STDERR.puts error unless error.empty?
```

src/commands/build.cr
```diff
@@ -44,9 +44,12 @@ module Shards
         Log.debug { "#{Shards.crystal_bin} #{args.join(' ')}" }
 
         error = IO::Memory.new
-
         status = Process.run(Shards.crystal_bin, args: args, output: Process::Redirect::Inherit, error: error)
-        raise Error.new("Error target #{target.name} failed to compile:\n#{error}") unless status.success?
+        unless status.success?
```
straight-shoota, line 48: Minor nitbit: I prefer to never use `unless` with `else` and instead swap both branches. The success case is also much more common, another reason to have it first.

However, I'd even suggest to leave this line as before as an early exit via raise and just continue in the main body of the method with the success flow.

luislavena:
Hello @beta-ziliani, was wondering why not simply inherit error too? That why, you only need to check for success of process status and raise the single error (shown at the bottom) after all the error has been automatically output to you, no need to check if there were errors collected.

I think would be simpler code, but maybe I'm missing what you're trying to achieve as DX.

Cheers.

straight-shoota:
Currently, in case of failure the captured STDERR is part of the error message.
If we just pass STDERR through, the output from `crystal build` would be printed independent of the error and not be part of the error message.
I think it's better to keep it as part of the error message for consistency. That requires capturing and reprinting, but it's not a big deal.

luislavena:
So in the case of warning:

```
❯ shards build
Dependencies are satisfied
Building: d
In d.cr:9:1

 9 | a
     ^
Warning: Deprecated top-level a.

A total of 1 warnings were found.
```

In the case of errors:

```
❯ shards build
Dependencies are satisfied
Building: d
Error target d failed to compile:
Showing last frame. Use --error-trace for full trace.

In d.cr:3:3

 3 | foo
     ^--
Error: undefined local variable or method 'foo' for top-level
```

Thought since _error_ is being displayed multiple times, the message about the failed target got lost in the middle, instead of something like:

```
❯ shards build
Dependencies are satisfied
Building: d
Showing last frame. Use --error-trace for full trace.

In d.cr:3:3

 3 | foo
     ^--
Error: undefined local variable or method 'foo' for top-level
Target d failed to compile
```

But what you propose makes sense, to make a _section_ in the output with the error. 👍🏽 

Cheers.

straight-shoota:
We could certainly change the order in the error message! That's not the point.

I'm concerned about maintaining that the message of the exception contains all the error information.

luislavena:
> I'm concerned about maintaining that the message of the exception contains all the error information.

Not sure I follow your concern on that, from what I understand is that Crystal outputs the error message entirely to STDERR, so the error is captured with the current approach.

```
$ crystal build d.cr --error-trace 
In d.cr:9:1

 9 | a
     ^
Warning: Deprecated top-level a.

A total of 1 warnings were found.

$ crystal build d.cr --error-trace 2>/dev/null

$ echo $?
0
```

Same for errors:

```
$ crystal build d.cr --error-trace 
In d.cr:10:1

 10 | foo
      ^--
Error: undefined local variable or method 'foo' for top-level

$ crystal build d.cr --error-trace 2>/dev/null

$ echo $?
1
```

Thank you.

spec/support/factories.cr
```diff
@@ -242,7 +242,7 @@ def tmp_path
   Shards::Specs.tmp_path
 end
 
-def run(command, *, env = nil)
+def run(command, *, env = nil, return_error = false)
```
straight-shoota, line 245: This added parameter doesn't seem to be used enywhere.
```suggestion
def run(command, *, env = nil)
```

straight-shoota (approved):

PR #545, opened 2022-02-23
Allow building multiple build targets in parallel

carlhoerberg:
Pass -j X or --jobs=X to start X build process in parallel to speed up
building applications with multiple targets (with the cost of more
memory).

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -27,6 +27,7 @@ module Shards
   def self.run
     OptionParser.parse(cli_options) do |opts|
       path = Dir.current
+      jobs = 1
 
       opts.on("--no-color", "Disable colored output.") { self.colors = false }
       opts.on("--version", "Print the `shards` version.") { puts self.version_string; exit }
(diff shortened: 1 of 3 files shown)


Blacksmoke16:
Might be better to contribute to the discussion/progress of https://github.com/crystal-lang/shards/pull/539 instead of introducing a new approach?

asterite:
I was thinking about that... but I think this is about building things in parallel, while the other one is fetching things in parallel... which of course will be very confusing if we have two flags to control these.

Blacksmoke16:
Oh sorry, I didn't notice this was for _building_. But yea, that's a good point. Maybe the flag could be consolidated/generalized to support both cases?

straight-shoota:
I don't like this because I don't want `shards` to become more of a build tool. It should focus on managing dependencies, and IMO does not need advanced build features.

And there are many alternatives. If you use an *actual* build system (which I would recommend for any project), it probably already has an option for parallel building (`make -j` for example).
This is also easy to do with the `parallel` command: `parallel shards build ::: target1 target2`.

m-o-e:
> might cause memory issues as each build process uses quite a lot of memory
> Hence you might want different defaults for --jobs and --parallel-fetch

Agree. If both get added they definitely need to have two separate flags.

In **bundler** _--jobs_ means: `The maximum number of parallel download and install jobs.` ([man-page](https://bundler.io/man/bundle-install.1.html))
In **make** _--jobs_ means: `Specifies the number of jobs (commands) to run simultaneously.`  ([man-page](https://linux.die.net/man/1/make))

Our own _--jobs_ flags can only be compatible with one of them, not both at the same time.
(as the two tend to have different optimal values, esp. on low-memory machines)

Since **shards** is conceptually closer to bundler than make, I would vote for
adopting the bundler meaning of _--jobs_ (as already done in #539).

If this one should also be added then I'd suggest to call the flag here _--parallel-build_ (or similar).

> I don't want shards to become more of a build tool.

I'd be inclined to agree with this as well. The default for this will probably always have to be `1`.
Meaning anyone who wants to use it will have to change their invocation.

And at that point they could as well replace it with something involving `parallel`.

I suppose the argument for having this functionality baked in could be
to make it more discoverable/easier, tho. 🤔

hugopl:
This is that kind of feature that at first looks very nice, but then you notice that we are crossing the borders between a dependency manager and a building tool... years later you notice that you just wrote a bad version of `make`.

PR #546, opened 2022-03-05
Add run command

### LINKED ISSUE #157: Run command

luislavena:
Introduces `shards run`, similar to `shards build`, that builds and runs the specified target, providing a unified interface to developers.

With this, a developer can directly use `shards run <target>` in a single step without having first to compile (build) and then execute it with `./bin/target`. No need for a script or wrapper.

This is inspired by `crystal build` and `crystal run` user experience (with the difference that `crystal run` creates a temporary binary).

This is very helpful for automatic rebuild scenarios, where a process is monitoring your code and can rebuild and relaunch your application.

When invoking `shards run`, it will first build the target (with the supplied options) and then invoke it, passing the extra options:

```console
$ shards run myapp --stats -- --args
```

Is the same as:

```console
$ shards build myapp --stats

$ ./bin/myapp --args
```

If the target exits with error, the exit code is passed back.

Resolves #157
Relates #298

Co-authored-by: Elliot DeNolf <65888+denolfe@users.noreply.github.com>

### DIFF
--- a/src/commands/run.cr
+++ b/src/commands/run.cr
@@ -1,0 +1,41 @@
+require "./command"
+
+module Shards
+  module Commands
+    class Run < Command
+      def run(targets, options, run_options)
+        if spec.targets.empty?
+          raise Error.new("Targets not defined in #{SPEC_FILENAME}")
+        end
+
+        # when more than one target was specified
+        if targets.size > 1
+          raise Error.new("Error please specify only one target. If you meant to pass arguments you may use 'shards run target -- args'")
+        end
+
+        # when no target was specified
+        if targets.empty?
+          if spec.targets.size > 1
+            raise Error.new("Error please specify the target with 'shards run target'")
+          else
+            name = spec.targets.first.name
+          end
+        else
+          name = targets.first
+        end
+
+        if target = spec.targets.find { |t| t.name == name }
+          Commands::Build.run(path, [target.name], options)
+
+          Log.info { "Executing: #{target.name} #{run_options.join(' ')}" }
+          status = Process.run(File.join(Shards.bin_path, target.name), args: run_options, output: Process::Redirect::Inherit, error: Process::Redirect::Inherit)
+          unless status.success?
+            exit status.exit_code
+          end
+        else
+          raise Error.new("Error target #{name} was not found in #{SPEC_FILENAME}")
+        end
+      end
+    end
+  end
+end
--- a/spec/integration/run_spec.cr
+++ b/spec/integration/run_spec.cr
@@ -1,0 +1,129 @@
+require "./spec_helper"
+
+def bin_path(name)
+  File.join(application_path, "bin", name)
+end
+
+describe "run" do
+  before_each do
+    Dir.mkdir(File.join(application_path, "src"))
+    File.write(File.join(application_path, "src", "cli.cr"), "puts __FILE__")
+  end
+
+  after_each do
+    File.delete File.join(application_path, "shard.yml")
+  end
+
+  it "fails when no targets defined" do
+    File.write File.join(application_path, "shard.yml"), <<-YAML
+      name: build
+      version: 0.1.0
+      YAML
+
+    Dir.cd(application_path) do
+      ex = expect_raises(FailedCommand) do
+        run "shards run --no-color"
+      end
+      ex.stdout.should contain("Targets not defined in shard.yml")
+    end
+  end
+
+  it "fails when passing multiple targets" do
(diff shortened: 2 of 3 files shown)


src/commands/run.cr
```diff
@@ -0,0 +1,41 @@
...
+        # when no target was specified
+        if targets.empty?
+          if spec.targets.size > 1
+            raise Error.new("Error please specify the target with 'shards run target'")
+          else
+            name = spec.targets.first.name
+          end
+        else
+          name = targets.first
+        end
```
luislavena, line 25: I'm too rusty to came with a non-convoluted version of these nested conditions 😬 

Better versions are welcome! 😊

src/commands/run.cr
```diff
@@ -0,0 +1,41 @@
...
+
+        if target = spec.targets.find { |t| t.name == name }
```
Sija, line 27:
```suggestion
        if target = spec.targets.find(&.name == name)
```

src/commands/run.cr
```diff
@@ -0,0 +1,41 @@
...
+          Commands::Build.run(path, [target.name], options)
+
+          Log.info { "Executing: #{target.name} #{run_options.join(' ')}" }
+          status = Process.run(File.join(Shards.bin_path, target.name), args: run_options, output: Process::Redirect::Inherit, error: Process::Redirect::Inherit)
```
Sija, line 31:
```suggestion
          status = Process.run(File.join(Shards.bin_path, target.name), args: run_options, output: :inherit, error: :inherit)
```

luislavena: Thought about this, however decided to follow the same pattern/style used in build command:

https://github.com/crystal-lang/shards/blob/master/src/commands/build.cr#L47

Could change if really needed.

luislavena:
One note, if approved, I would like to squash all the changes into a single commit.

Thank you.

src/commands/run.cr
straight-shoota: The implementation looks fine.

But I'm not sure the resulting behaviour is what we should have. At least the discussion in #157 seemed pretty indecisive on that. It might not be that important *what* we pick, but this alternative is at least not my favourite 🤷

luislavena: > but this alternative is at least not my favourite 

Can you rephrase or explain this? Not sure I follow. The scenarios I made were:

1. When no target was indicated and spec contains only one, assume is that target, runs.
2. When no target was indicated and spec contains more than one, errors out

Looking at the comments in #157 is confusing me:

> Yes. No default, unless there is only 1 target.
https://github.com/crystal-lang/shards/issues/157#issuecomment-539667298

> I'm not sure whether the default behaviour is a good idea. shards build defaults to building all. shards run defaults to running one, but only if there is one defined seems off.
https://github.com/crystal-lang/shards/issues/157#issuecomment-540136983

> Always running the first target if no name specified sounds good. That's consistent behaviour.
https://github.com/crystal-lang/shards/issues/157#issuecomment-541431000

> I lean more and more on always specifying the target to run: consistent, no configuration, no documentation, no repeated boring message (that boils down to: you should always specify the target), no surprises (you know what will happen).
https://github.com/crystal-lang/shards/issues/157#issuecomment-541569088

(skipping the _shards isn't a build tool_ dilema)

I'm ok with making the target mandatory, that will definitely simply the above logic, but I thought of `shards run` similar to `cargo run` and the [default target selection](https://doc.rust-lang.org/cargo/commands/cargo-run.html#target-selection)

Either way, again, we can change it, that is why I sent this PR 😀

straight-shoota: Yeah, we just need to come to a decision in #157. There are contradicting views on that.
Having a PR ready certainly helps to push this forward ;)

src/commands/run.cr
beta-ziliani: IMO the original reads better.

src/commands/run.cr
beta-ziliani: Either way is good

beta-ziliani (approved):
To me this PR does what I expect. If we want to change the semantics of an unspecified target with multiple targets configured, we can see how to proceed later.

Many thanks for your attention to detail @luislavena <3

straight-shoota (approved):

PR #552, opened 2022-04-22
Fix clear ENV in integration spec

straight-shoota:
Clear environment when running integration spec. If the environment has `CRYSTAL_OPTS=--error-on-warnings` it would get passed through to the compiler and the spec fails (see https://github.com/crystal-lang/test-ecosystem/runs/6116920991).

This is a minimal change to fix the immediate issue. It could be considered that we apply this to more uses of `run`. Maybe inheriting the environment should even be opt-in.

### DIFF
--- a/spec/support/factories.cr
+++ b/spec/support/factories.cr
@@ -356,6 +356,9 @@ def run(command, *, env = nil, clear_env = false)
   cmd_env = {
     "CRYSTAL_PATH" => Shards::Specs.crystal_path,
   }
+  if clear_env
+    cmd_env["CRYSTAL_OPTS"] = ""
+  end
   cmd_env.merge!(env) if env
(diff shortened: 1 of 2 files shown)


spec/support/factories.cr
```diff
@@ -352,10 +352,13 @@ def tmp_path
   Shards::Specs.tmp_path
 end
 
-def run(command, *, env = nil)
+def run(command, *, env = nil, clear_env = false)
```
Sija, line 355: `clear_env` is misleading since it's not what it does.

```suggestion
def run(command, *, env = nil, clear_crystal_opts = false)
```

straight-shoota: I'd like the name to reflect the intention, not the actual implementation. We might need to clear other variables as well at some point.

asterite (approved):

PR #557, opened 2022-08-30
prune - Don't raise an exception if install_path doesn't exist

mjoerussell:
I noticed that if you run `shards prune` in a project with the following two conditions:

1. `shards.lock` exists
2. The install directory (`lib/`) does not exist

Then the command raises an exception. This might not be a super common scenario, but in the context of CI it can be frustrating for this command to throw and then interrupt the whole job rather than just not doing anything.

### DIFF
--- a/src/commands/prune.cr
+++ b/src/commands/prune.cr
@@ -8,6 +8,11 @@ module Shards
       def run
         return unless lockfile?
 
+        unless Dir.exists?(Shards.install_path)
+          Log.info { "Pruned nothing, because #{File.basename(Shards.install_path)}/ does not exist" }
+          return
+        end
+
         Dir.each_child(Shards.install_path) do |name|
           path = File.join(Shards.install_path, name)
           next unless File.directory?(path)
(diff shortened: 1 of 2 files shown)


src/commands/prune.cr
```diff
@@ -8,6 +8,11 @@ module Shards
       def run
         return unless lockfile?
 
+        unless Dir.exists?(Shards.install_path)
+          Log.info { "Pruned nothing, because #{File.basename(Shards.install_path)}/ does not exist" }
```
straight-shoota, line 12: I don't think this should be logged. At least not as `info` severity, which gets printed to STDOUT by default.
If the directory doesn't exist, the command simply is a no-op. No need to be chatty about that.
The same happens when there are no directories in `install_path`.

We can make it a `debug` level log entry:
```suggestion
          Log.debug { "Pruned nothing, because #{File.basename(Shards.install_path)}/ does not exist" }
```
But I'm not sure whether even that is much useful. Removing it entirely is probably fine.

straight-shoota (approved):

beta-ziliani (approved):
🚀

PR #561, opened 2022-10-05
Inherit the standard input descriptor

### LINKED ISSUE #550: Standard input file descriptor immediately closed when running a program using `shards run`
I noticed that standard input descriptor is immediately closed when running `shards run`. This is evident by the following code:

https://github.com/crystal-lang/shards/blob/85b30b5755b5aad8e5b4bd94cc94610c6a617bea/src/commands/run.cr#L31

Only standard output and standard error are inherited from the parent process.

Is this a mistake or by design? I'd be happy to address the problem in case it's the former. 🙂

hovsater:
When running `shards run`, you should be able to work with standard input.

Fix #550

### DIFF
--- a/spec/integration/run_spec.cr
+++ b/spec/integration/run_spec.cr
@@ -148,4 +148,25 @@ describe "run" do
       output.should contain("args: foo,bar,baz")
     end
   end
+
+  it "works well with stdin" do
+    File.write File.join(application_path, "src", "stdin.cr"), <<-CR
+      print "input: ", STDIN.gets.inspect
+      CR
+
+    File.write File.join(application_path, "shard.yml"), <<-YAML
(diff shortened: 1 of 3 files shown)


straight-shoota (approved):
Awesome first contribution.

Thanks for picking this up =)

asterite (approved):

Blacksmoke16:
In one of my projects I'm using `stty` to get the size of the terminal. When running via `shards run` I noticed it prints `stty: 'standard input': Inappropriate ioctl for device` to standard error, while building and running it does not.

My understanding is it should be using the same IO, so they should be equivalent? I don't think it really hurts anything so could just silence `STDERR` and not worry about it too.

PR #573, opened 2023-01-19
Fix swallowing original error message in `git_retry`

### LINKED ISSUE #525: Wrong error message if git is unvailable
If the `git` command is unavailable I get: 

```
Locked version 0.1.1 for PKG was not found in git: https://gitlab.com/X/PKG.git.
```

Instead of expected: 
```
Error missing git command line tool. Please install Git first!
```

I guess https://github.com/crystal-lang/shards/blob/02055f9588f2b80c7dfd290b5a29051744ef0c00/src/resolvers/git.cr#L421 is swallowed here: https://github.com/crystal-lang/shards/blob/02055f9588f2b80c7dfd290b5a29051744ef0c00/src/molinillo_solver.cr#L33

straight-shoota:
The equivalent method in the other resolvers already prints the original error message, just `git` didn't do that.

Fixes #525

### DIFF
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -1266,4 +1266,12 @@ describe "install" do
       end
     end
   end
+
+  it "fails when git is missing" do
+    metadata = {dependencies: {web: "*"}}
+    with_shard(metadata) do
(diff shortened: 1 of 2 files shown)


beta-ziliani (approved):

beta-ziliani:
ah, needs a re-format it seems

PR #577, opened 2023-01-26
refactor: ameba linting

devnote-dev:
This PR adds the [ameba](https://github.com/crystal-ameba/ameba) code analysis/linting tool to Shards to help catch some code structure issues and improve performance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,8 @@
 /.crystal
 /.shards
 /lib
-/bin/shards
-/bin/shards.dwarf
+/bin
+!/bin.keep
 /spec/.repositories
 /spec/.shards
 /spec/unit/.lib
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 /.shards
 /lib
 /bin
-!/bin.keep
+!/bin/.keep
 /spec/.repositories
 /spec/.shards
 /spec/unit/.lib
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -533,6 +533,7 @@ describe "install" do
         override = {dependencies: {
           awesome: {git: git_url(:forked_awesome), branch: "feature/super"},
         }}
+        git_commits(:forked_awesome).first?.should_not be_nil
 
         with_shard(metadata, lock, override) do
           ex = expect_raises(FailedCommand) { run "shards install --no-color --#{flag}" }
@@ -550,6 +551,7 @@ describe "install" do
         override = {dependencies: {
           awesome: {git: git_url(:forked_awesome), branch: "feature/super"},
         }}
+        git_commits(:forked_awesome).first?.should_not be_nil
 
         with_shard(metadata, lock, override) do
           ex = expect_raises(FailedCommand) { run "shards install --no-color --#{flag}" }
--- a/spec/integration/install_spec.cr
+++ b/spec/integration/install_spec.cr
@@ -533,7 +533,6 @@ describe "install" do
         override = {dependencies: {
           awesome: {git: git_url(:forked_awesome), branch: "feature/super"},
         }}
-        expected_commit = git_commits(:forked_awesome).first
 
         with_shard(metadata, lock, override) do
           ex = expect_raises(FailedCommand) { run "shards install --no-color --#{flag}" }
@@ -551,7 +550,6 @@ describe "install" do
         override = {dependencies: {
           awesome: {git: git_url(:forked_awesome), branch: "feature/super"},
         }}
-        expected_commit = git_commits(:forked_awesome).first
 
         with_shard(metadata, lock, override) do
           ex = expect_raises(FailedCommand) { run "shards install --no-color --#{flag}" }
--- a/spec/support/factories.cr
+++ b/spec/support/factories.cr
@@ -321,7 +321,7 @@ def fossil_url(project)
 end
 
 def fossil_path(project)
-  File.join(tmp_path, "#{project}")
+  File.join(tmp_path, project.to_s)
 end
 
 def rel_path(project)
--- a/spec/support/factories.cr
+++ b/spec/support/factories.cr
@@ -306,13 +306,13 @@ def fossil_commits(project, rev = "trunk")
   # using an old Fossil version.  See the #commit_sha1_at method in
   # src/resolvers/fossil.cr for info.
   Dir.cd(fossil_path(project)) do
-    retStr = run("fossil timeline #{Process.quote(rev)} -t ci -W 0").strip.split('\n')
-    retLines = retStr.flat_map do |line|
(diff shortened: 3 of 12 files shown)


.gitignore
```diff
@@ -1,8 +1,8 @@
 /.crystal
 /.shards
 /lib
-/bin/shards
-/bin/shards.dwarf
+/bin
+!/bin.keep
```
j8r, line 5:
```suggestion
!/bin/.keep
```

src/target.cr
```diff
@@ -9,12 +9,7 @@ module Shards
       main = nil
 
       pull.each_in_mapping do
-        case key = pull.read_scalar
-        when "main"
-          main = pull.read_scalar
-        else
-          # ignore unknown dependency mapping for future extensions
-        end
```
straight-shoota, line 17: This is a very questionable improvement.
In the original code the order of `read_scalar` calls is clearly aligned with the execution order. With this change it's confusing because the key is read first, but it's in a trailing condition.
So this should at least use a normal conditional, not a trailing one.

But I actually like the `case` statement. I guess it's not too important to keep it, but I don't see a real reason for changing it either.
*If* we change it, the comment should be kept, though.

straight-shoota (commented):
I'm honestly surprised there are just a couple of changes. Most of them look good.

src/target.cr
```diff
@@ -9,12 +9,7 @@ module Shards
...
+        main = pull.read_scalar if pull.read_scalar == "main"
```
straight-shoota, line 12: This is a very questionable improvement.
In the original code the order of `read_scalar` calls is clearly aligned with the execution order. With this change it's confusing because the key is read first, but it's in a trailing condition.
So this should at least use a normal conditional, not a trailing one.

But I actually like the `case` statement. I guess it's not too important to keep it, but I don't see a real reason for changing it either.
*If* we change it, the comment should be kept, though.

devnote-dev: Ameba didn't like the unused `key` variable so I just switched it to something more compact, but I agree that the `case` statement gave it better readability. Perhaps a simple `if` statement instead? It feels like a bit of a waste to use `case` for one condition.

spec/integration/install_spec.cr
```diff
@@ -533,7 +533,6 @@ describe "install" do
         override = {dependencies: {
           awesome: {git: git_url(:forked_awesome), branch: "feature/super"},
         }}
-        expected_commit = git_commits(:forked_awesome).first
```
asterite, line 536: It might be the case that this was checking that there was at least one commit.

devnote-dev: That seems to be the case, but the output was never used and the specs still ran without it.

asterite: `first` will raise an exception if the array is empty, making the spec fail. I think we should keep this expectation (not in this form, but in another form)

straight-shoota: `git_commits(:forked_awesome).first?.should_not be_nil`

spec/support/factories.cr
```diff
@@ -321,7 +321,7 @@ def fossil_url(project)
 end
 
 def fossil_path(project)
-  File.join(tmp_path, "#{project.to_s}")
+  File.join(tmp_path, "#{project}")
```
HertzDevil, line 324:
```suggestion
  File.join(tmp_path, project.to_s)
```

src/commands/check.cr
```diff
@@ -15,7 +15,7 @@ module Shards
       end
 
       private def has_dependencies?
-        spec.dependencies.any? || (Shards.with_development? && spec.development_dependencies.any?)
+        !spec.dependencies.empty? || (Shards.with_development? && !spec.development_dependencies.empty?)
```
jkthorne, line 18: why use `!` and `.empty?` this seems to be less clear.

devnote-dev: According to Ameba, `!{...}.empty?` is faster than `.any?`, and you can kinda gauge this based on the method implementations:
* [`any?`](https://github.com/crystal-lang/crystal/blob/master/src/enumerable.cr#L111)
* [`empty?`](https://github.com/crystal-lang/crystal/blob/master/src/enumerable.cr#L1496)

Sija: It's more about the correctness than speed. Since `any?` checks for the first truthy element, it could return `true` even if the enumerable is not empty (given it contains `nil` or `false` values).

straight-shoota: `any?` vs `!empty?` only makes a difference when items can be falsey. That's not the case here. `spec.depdencies` is of type `Array(Dependency)`, so `any?` has the exact same semantics as `!empty?`.

Now there is still a performance difference, but the implementation of `any?` just needs to be optimized for this case.

IMO an important aspect is code readability and there `any?` clearly wins over `!empty?`.

straight-shoota: I just checked, in release mode `any?` produces exactly identical code as `empty?` when the item type cannot be falsey (i.e. does not include `Nil`, `Bool` or `Pointer`).
So I don't think this change is justified.

devnote-dev: Would you be able to provide this code in an issue for Ameba? I think it would be useful to revisit the rule that enforces `empty?` over `any?`. Otherwise, I'm happy to revert this and disable the rule.

Sija: > IMO an important aspect is code readability and there `any?` clearly wins over `!empty?`.

I disagree. `any?` has a different semantics - checking for truthy elements. Checking whether the collection is empty should be done with call to `empty?` (duh).

Sija: There's a lot already written on the subject, for instance: https://thoughtbot.com/blog/any-empty

straight-shoota: I've created a forum post to continue this general discussion: https://forum.crystal-lang.org/t/collections-any-vs-empty/5303

As for this particular change, I'd suggest to drop it and keep the status quo with `any?`. Both versions are semantically equivalent and the current one is more readable, so I see no gain from changing it.

devnote-dev: Reverted in e7d8a727e0f45ae468a27ef23f0901418c7b5558.

src/target.cr
devnote-dev: Reverted in 4525063e95295a6081fc790529b8e41740168fb8.

src/commands/check.cr
Sija: > Both versions are semantically equivalent and the current one is more readable

I'm sorry, but please do not rewire words. They're not _semantically_ equivalent - _is (not) empty_ is semantically different thing than: _is there any element (that ... - `any?` takes a block after all)_.

And there's nothing more readable in having to understand the semantics of `any?` to answer this simple question - _empty or not_.

devnote-dev:
No longer interested in pursuing this.

PR #594, opened 2023-07-19
ci: Ensure manpages are generated with no diff

straight-shoota:
Manpages in `man/` are generated from adoc files in `docs/`. Both source and target format are checked into git.
Often it's forgotten to update the manpages when the adoc sources are updated (latest example: #587).

This change ensures that the manpages are up to date in CI by rebuilding them and checking if they produce a diff.


beta-ziliani:
Can you explain the issue?

straight-shoota:
Yeah, added it tot he OP. Sorry, I was deep in a rabbit hole 😆

PR #602, opened 2023-12-06
[CI] Pin `Analog-inc/asciidoctor-action` version

straight-shoota:


beta-ziliani:
Looks good, do we have a way to track versions? (Renovate or something?). The errors in CI are expected?

straight-shoota:
We would have renovate if just somebody approved #564 🤷

The failure was kinda expected because that was the reason to pin this. I still needed it to run to verify the exact version of asciidoctor used in the action.

HertzDevil (approved):

PR #615, opened 2024-02-16
Add `Makefile.win`

HertzDevil:
The docs-related targets are dropped, because manfiles do not make sense on Windows. Everything else should work, in particular setting up the Git commit hash and the build date for `bin\shards.exe --version`. Setting `%SHARDS%` to `false` also works as long as both `curl` and `tar` are available ([Windows 10 should come with them already](https://learn.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows)).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,10 +60,10 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Build
-        run: make
+        run: make -f ${{ matrix.os == 'windows-2022' && 'Makefile.win' || 'Makefile' }}
 
       - name: Run specs
-        run: make test
+        run: make -f ${{ matrix.os == 'windows-2022' && 'Makefile.win' || 'Makefile' }} test
 
       - name: Check formatting
         run: crystal tool format --check src spec
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,10 +60,10 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Build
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -60,10 +60,10 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Build
-        run: make
+        run: make -f ${{ runner.os == 'Windows' && 'Makefile.win' || 'Makefile' }}
```
Sija, line 63: I'd suggest using ternary operator as it's more straightforward.

```suggestion
        run: make -f ${{ runner.os == 'Windows' ? 'Makefile.win' : 'Makefile' }}
```

.github/workflows/ci.yml
```diff
@@ -60,10 +60,10 @@ jobs:
...
 
       - name: Run specs
-        run: make test
+        run: make -f ${{ runner.os == 'Windows' && 'Makefile.win' || 'Makefile' }} test
```
Sija, line 66: ditto

```suggestion
        run: make -f ${{ runner.os == 'Windows' ? 'Makefile.win' : 'Makefile' }} test
```

.github/workflows/ci.yml
HertzDevil: [They are not available](https://github.com/orgs/community/discussions/75928)

Sija: Wow, GHA weirdness make my head spin sometimes 🤯 Thanks for the clarification.

straight-shoota (approved):
Sigh, I don't like this duplication at all. 😢  But I guess there's no other way (without throwing everything over).

Issue #574, opened 2023-01-19
Refactor resolver code


straight-shoota:
The three resolvers git, hg and fossil share a lot of very similar code. The latter ones are both based on the git implementation.
Just copying the code was nice for getting started. But that much duplication carries a technical debt. It's hard to keep track of things, especially when there are slight differences between the implementations, some of them seemingly arbitrary.

Two recently fixed bugs #569 and #573 only existed in the git resolver. The other two used basically the same code, but with the bugs already fixed. The fixes were never brought back to the git resolver, though.

I think quite a bit of shared functionality can be abstracted and put in a general utility module.
A major common aspect is the CLI, running processes with error handling etc.

PR #626, opened 2024-04-19
Fix: make 'v' prefix optional

### LINKED ISSUE #521: UX: Shards requires tags to start with "v"
I recently went to restrict a dependency to a specific version:

```
dependencies:
  taskbox:
    git: git@git.sr.ht:~z64/taskbox
    version: 0.2.0
```
and received:
```
$ shards -v install
Resolving dependencies
git ls-remote --get-url origin
Fetching git@git.sr.ht:~z64/taskbox
git fetch --all --quiet
git tag --list --column=never

- `taskbox (0.2.0)` required by `shard.yml`
```
so I did some checking:
```
taskbox $ git tag --list --column=never
0.1.0
0.1.1
0.2.0

taskbox $ cat shard.yml
name: taskbox
version: 0.2.0
```
Seeing as it wasn't working even when everything shown to me says "0.2.0" so far, I checked the source:
```
$ rg 'VERSION_TAG' src/
src/resolvers/git.cr
209:        .compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }

src/config.cr
13:  VERSION_TAG           = /^v(\d+[-.][-.a-zA-Z\d]+)$/
```
and it would appear there is a hard requirement that tags start with "v".

---

The docs have this to say about `git`:

> git
> A Git repository URL (string).
> The URL may be [any protocol](https://git-scm.com/docs/git-clone#_git_urls) supported by Git, which includes SSH, GIT and HTTPS.
> The Git repository will be cloned, the list of versions (and associated shard.yml) will be extracted from Git tags (e.g., v1.2.3).
> One of the other attributes (version, tag, branch or commit) is required. When missing, Shards will install the HEAD refs.
> Example: git: git://git.example.org/crystal-library.git

If I am not missing something, if the "e.g., v1.2.3" was supposed to convey this requirement, I think it would be nice to reword this section to make it more clear - or even better, an improved error message. 

Personally, I'm not sure that this requirement is necessary in the first place, but I don't really mind as long as the above UX could be smoothed over.

devnote-dev:
Closes #521.

### DIFF
--- a/src/config.cr
+++ b/src/config.cr
@@ -10,7 +10,7 @@ module Shards
   DEFAULT_VERSION = "0"
 
   VERSION_REFERENCE        = /^v?\d+[-.][-.a-zA-Z\d]+$/
-  VERSION_TAG              = /^v(\d+[-.][-.a-zA-Z\d]+)$/
+  VERSION_TAG              = /^v?(\d+[-.][-.a-zA-Z\d]+)$/
   VERSION_AT_GIT_COMMIT    = /^(\d+[-.][-.a-zA-Z\d]+)\+git\.commit\.([0-9a-f]+)$/
   VERSION_AT_HG_COMMIT     = /^(\d+[-.][-.a-zA-Z\d]+)\+hg\.commit\.([0-9a-f]+)$/
   VERSION_AT_FOSSIL_COMMIT = /^(\d+[-.][-.a-zA-Z\d]+)\+fossil\.commit\.([0-9a-f]+)$/


ysbaddaden:
Thanks!

An immediate issue is that there are no tests to ensure that the resolvers do support the new kind of version tags (install might be broken), or that they can handle a mix of them from different shards (probably not an issue).

I suppose edge cases should also be handled: what if both tags exist (with the current patch they will both be listed)? won't Molinillo break with two version tags? what if they point to different commits? should that be an error? should one form take priority (to add to the confusion).

Another issue is that sites such as https://shardbox.org and the ecosystem will now have to support both kinds, instead of assuming a convention.

devnote-dev:
> An immediate issue is that there are no tests to ensure that the resolvers do support the new kind of version tags (install might be broken), or that they can handle a mix of them from different shards (probably not an issue).

@ysbaddaden I'm not sure I understand what you're asking for here — the prefix was a hard requirement prior to this meaning resolvers had to support it, all this PR does is remove said requirement which wouldn't affect anything.

> I suppose edge cases should also be handled: what if both tags exist (with the current patch they will both be listed)? won't Molinillo break with two version tags? what if they point to different commits? should that be an error? should one form take priority (to add to the confusion).

This isn't changing any core functionality of the resolver. If both `v0.2.0` and `0.2.0` exist, the resolver will pick the version that is specified. Version mismatches (i.e. meaning to pick one over the other) ultimately comes down to a user error which is beyond the scope of Shards.

> Another issue is that sites such as https://shardbox.org/ and the ecosystem will now have to support both kinds, instead of assuming a convention.

Not sure how that plays into this? If Shardbox also faced this issue then it should have been something addressed on their end.

> So far there has not been any talk about these details because https://github.com/crystal-lang/shards/issues/521 hasn't come to any conclusion whether this is actually wanted.

@straight-shoota I opted for this solution because it's easier to implement (literally a one line change) and aligns with other package management tools. Changing the error message to be more user-friendly was the original suggestion but Shards' codebase is annoyingly ambiguous regarding typings and its order of operations (highlighted in #574).

ysbaddaden:
> I'm not sure I understand what you're asking for

The test suite verifies that version tags starting with a `v` still work; they don't test that version tags _without_ a `v` are working. We need a set of tests to prove that they work.

> This isn't changing any core functionality of the resolver

Having both `v0.2.0` and `0.2.0` means that we'll have the `0.2.0` version twice from different tags. In the lucky case they're duplicates and it's fine (though we should at least de-duplicate), in the unlucky case they're different commits with different `shard.yml` and we must have a rule to handle this edge case (e.g. skip both tags and _warn_ about it).

devnote-dev:
> The test suite verifies that version tags starting with a `v` still work; they don't test that version tags _without_ a `v` are working. We need a set of tests to prove that they work.

As far as I can see the resolver specs don't test for "v" so I'm not sure what to change there.

> in the unlucky case they're different commits with different `shard.yml` and we must have a rule to handle this edge case (e.g. skip both tags and _warn_ about it).

I'm not sure where exactly this would go and I'm not having any issues reproducing it locally.

straight-shoota:
I'm closing this because it's an incomplete, trivial solution. There's no point in keeping this PR open and spreading the discussion across multiple threads.

Changing this behaviour is decisively *not* a one-line change. It involves other parts of the code base, the spec suite and other participants in the shards ecosystem at large.

We should continue the discussion in https://github.com/crystal-lang/shards/issues/521 about whether this change actually makes sense, and if so, figure out *how* to properly implement it.

PR #629, opened 2024-04-25
Add devenv configuration

straight-shoota:
Configuration for a https://devenv.sh environment with all necessary tools to build and test shards.
I'm using this locally and it makes sense to share it in the repo so others can make use of it as well.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,1 @@
+/nix/store/2bvs1hdl54w2x8wjbh2icvv6wqn69dzj-pre-commit-config.json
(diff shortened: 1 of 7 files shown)


straight-shoota:
Potentially, this can be useful for setting up an environment in CI as well.
However our current setup works just fine across multiple platforms (including Windows which Nix does not support), so it's fine for the moment.

.pre-commit-config.yaml
```diff
@@ -0,0 +1 @@
+/nix/store/2bvs1hdl54w2x8wjbh2icvv6wqn69dzj-pre-commit-config.json
```
Sija, line 1: Is this file needed?

straight-shoota: Uh yeah, apparently not. Good catch!
Dunno if this has changed since I originally ran `devenv init` (that was a while ago). Nowadays it puts this file in `.gitignore`. So I'll just rebuild from scratch to get things setup to the latest configuration.

beta-ziliani (approved):

PR #631, opened 2024-05-20
Forward unmodified ARGV to subcommand

### LINKED ISSUE #600: Certain CLI options are not passed to shards custom subcommands
Hello folks,

When creating an external shards subcommand (using `shards-xxx` executable):

https://github.com/crystal-lang/shards/blob/master/src/cli.cr#L101-L105

Noticed it does not pass to the invoked executable the already processed commands like `--production`, `--without-development`, `--verbose` or `--quiet`, as these were already parsed by OptionParser.

You can confirm this with the following sample executable script:

```sh
#!/usr/bin/env sh

echo "args: $@"
```

Name it `shards-dummy`, make it executable and place it in `$PATH` and when invoked, notice certain options where already processed:

```console
$ shards dummy --production --release --static --verbose
args: --release --static
```

This differs from the behavior of `build` or `run`, which passes all `-` starting options directly to it. Additionally, the impact of `--production` or `--without-development` is not validated (something that is done by build and run by doing a check or install).

Since there hasn't been lot of discussion on this subject, I wanted to check first what is the expected behavior:

1. Should all the args be passed verbatim to the custom shards subcommand? Including `--production` or `--without-development` ?
2. Should these special shards options be processed first before delegating? (perform a check or install before invoking the subcommand)
3. Should at least the logging options be passed to the subcommand? Eg. `--verbose` and `--quiet`

Last but not least, this also impacts `--help`, as any custom subcommand cannot provide their own help information as this is short-circuit [here](https://github.com/crystal-lang/shards/blob/master/src/cli.cr#L56), but perhaps that is a separate issue.

I wanted to confirm what are the expectations before proposing a PR with changes.

Thank you in advance for your responses!
❤️ ❤️ ❤️

luislavena:
Hello, this PR aims to forward all arguments supplied to `shards` to the subcommand, if found.

It does this to allow correct parsing of all ARGV supplied, including `SHARDS_OPTS` that might be appended, avoids altering the original `ARGV` when combining it.

Introduces a naive integration test for subcommand to validate the change works correctly.

Last but not least, allows also forwarding `--help` by avoiding it to short-circuit and return immediately by
setting a flag for it and evaluating at the end of the processing of unknown options.

This is only done for the CLI invocation and is not part of Shards module (as the help and usage options are only available in this context).

I'm not sure about the implementation as this was the first integration test for subcommands, so would appreciate some notes on additional changes.

Thank you.

Fixes #600

### DIFF
--- a/spec/integration/subcommand_spec.cr
+++ b/spec/integration/subcommand_spec.cr
@@ -1,0 +1,21 @@
+require "./spec_helper"
+
+describe "subcommand" do
+  it "forwards all arguments to subcommand" do
+    create_shard("dummy", "0.1.0")
+    create_executable "dummy", "bin/shards-dummy", %(print "args: "; print ARGV)
+
+    with_path(git_path("dummy/bin")) do
+      output = run("shards dummy --no-color --verbose --unknown other argument")
+      output.should contain(%(args: ["--no-color", "--verbose", "--unknown", "other", "argument"]))
+    end
+  end
+end
+
+private def with_path(path)
+  old_path = ENV["PATH"]
+  ENV["PATH"] = "#{File.expand_path(path)}#{Process::PATH_DELIMITER}#{ENV["PATH"]}"
+  yield
+ensure
+  ENV["PATH"] = old_path
+end
--- a/spec/integration/subcommand_spec.cr
+++ b/spec/integration/subcommand_spec.cr
@@ -10,6 +10,16 @@ describe "subcommand" do
       output.should contain(%(args: ["--no-color", "--verbose", "--unknown", "other", "argument"]))
     end
   end
+
+  it "correctly forwards '--help' option to subcommand" do
+    create_shard("dummy", "0.1.0")
+    create_executable "dummy", "bin/shards-dummy", %(print "args: "; print ARGV)
+
+    with_path(git_path("dummy/bin")) do
+      output = run("shards dummy --help")
+      output.should contain(%(args: ["--help"]))
+    end
+  end
 end
 
 private def with_path(path)
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -101,7 +101,7 @@ module Shards
         else
           program_name = "shards-#{args[0]}"
           if program_path = Process.find_executable(program_name)
-            run_shards_subcommand(program_path, args)
+            run_shards_subcommand(program_path, cli_options)
           else
             display_help_and_exit(opts)
           end
@@ -119,7 +119,7 @@ module Shards
     {% else %}
       shards_opts = ENV.fetch("SHARDS_OPTS", "").split
     {% end %}
-    ARGV.concat(shards_opts)
+    ARGV.dup.concat(shards_opts)
   end
 
   def self.run_shards_subcommand(process_name, args)
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -2,6 +2,8 @@ require "option_parser"
 require "./commands/*"
 
 module Shards
+  class_property? display_help : Bool = false
+
   def self.display_help_and_exit(opts)
     puts <<-HELP
       shards [<options>...] [<command>]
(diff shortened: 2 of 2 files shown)


ysbaddaden:
I like how Git, for example, distinguishes between general and command arguments: those _before_ the command name are general git arguments, while those after the command name will all be forwarded to the actual command.

Applied to Shards:
- `shards --help dummy` would let `shards` handle `--help`
- `shards dummy --help` would call `shards-dummy --help`

luislavena:
@ysbaddaden I like that too, even it mentions `git help -a` to list all the available subcommands since the default `--help` lists only some of the core (porcelain) ones (like init, merge, commit, branch, etc).

Off-topic: I found OptionParser quite limited on that, but from my personal point of view, most of the other shards out there are too overwhelming (aka: _everything but the kitchen sink_).

Unsure if I should back off from the changes in commit [5a6dc6c](https://github.com/crystal-lang/shards/pull/631/commits/5a6dc6cb28973f22758532a76e4585cb13969cfc) and leave it for Shards main program to respond to `--help`, that way might avoid confusion.

Open to suggestions!

Thank you for looking into it!
❤️ ❤️ ❤️

src/cli.cr
```diff
@@ -119,7 +119,7 @@ module Shards
     {% else %}
       shards_opts = ENV.fetch("SHARDS_OPTS", "").split
     {% end %}
-    ARGV.concat(shards_opts)
+    ARGV.dup.concat(shards_opts)
```
straight-shoota, line 128: **question:** I'm wondering why is this `dup` necessary?

luislavena: `cli_options` modifies `ARGV` by concatenating `SHARDS_OPTS` possible values to it. Then OptionParser consumes that by eliminating the processed arguments.

This means that multiple calls to `cli_options` will return different arguments before and after OptionParser. This breaks attempting to pass arguments to the subcommand.

Duping ARGV ensures `cli_options` returns a copy of the parameters and keeps ARGV clean from the concatenation with `SHARDS_OPTS`.

spec/integration/subcommand_spec.cr
```diff
@@ -0,0 +1,21 @@
+require "./spec_helper"
+
+describe "subcommand" do
+  it "forwards all arguments to subcommand" do
+    create_shard("dummy", "0.1.0")
+    create_executable "dummy", "bin/shards-dummy", %(print "args: "; print ARGV)
```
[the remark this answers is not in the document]
luislavena: I will check if there is a helper for that, but this single compilation is way cheaper than attempting to run all the Git and Mercurial operations.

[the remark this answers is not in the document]
luislavena: @straight-shoota looking at how that is done in other parts of the codebase, seems that building the executable is the go-to for Windows and using sh script could be the option for posix, see this:

https://github.com/crystal-lang/shards/blob/master/spec/integration/spec_helper.cr#L71-L78

I just pushed some changes for this.

src/cli.cr
straight-shoota: Makes sense. It's a bit convoluted I guess, but it's what we have to work with 🤷

spec/integration/subcommand_spec.cr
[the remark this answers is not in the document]
straight-shoota: Ah I reckon the sub-command lookup doesn't work with shell scripts on Windows. We actually need a binary executable.
I'll create an issue for that.

src/cli.cr
luislavena: Indeed, but I didn't want to refactor the entire thing right now 😅

src/cli.cr
```diff
@@ -2,6 +2,8 @@ require "option_parser"
 require "./commands/*"
 
 module Shards
+  class_property? display_help : Bool = false
```
straight-shoota, line 5: **question:** What's the purpose of using a class variable?
This property is entirely scoped to the `.run` method, so it looks like it could be a local variable there?

straight-shoota:
I don't have any further comments at this point.
You could wait for feedback from others, if you want.

src/cli.cr
```diff
@@ -53,7 +55,7 @@ module Shards
       opts.on("--ignore-crystal-version", "Has no effect. Kept for compatibility, to be removed in the future.") { }
       opts.on("-v", "--verbose", "Increase the log verbosity, printing all debug statements.") { self.set_debug_log_level }
       opts.on("-q", "--quiet", "Decrease the log verbosity, printing only warnings and errors.") { self.set_warning_log_level }
-      opts.on("-h", "--help", "Print usage synopsis.") { self.display_help_and_exit(opts) }
+      opts.on("-h", "--help", "Print usage synopsis.") { display_help = true }
```
Sija, line 58: Why's this change needed at all?

luislavena: This is explained in the commit itself: 5a6dc6cb28973f22758532a76e4585cb13969cfc

The block for `#on` will short circuit and the argument `--help` is never forwarded to the subcommand.

This is necessary for the subcommand to be able to have their own `--help` option.

Sija: I see, thank you for the explanation.

straight-shoota (approved):

ysbaddaden (approved):

PR #633, opened 2024-05-31
Add debug information about tags

beta-ziliani:
This PR introduces two minor changes as a palliative to #521 :
 * When using `--versbose`, shards lists the tags it found in the repository and those that match pattern `vX.Y`.
 * When failing, it suggests using `--verbose`.
 
 I wasn't sure if `vX.Y` is descriptive enough, but the bare matching string is even less readable...
 
 <details>
<summary>Output on a failing run:</summary>
 

 ```
 ➜  shard-test git:(main) ✗ shards          
Resolving dependencies
Fetching https://github.com/spider-gazelle/priority-queue.git

- `priority-queue (0.2.0)` required by `shard.yml`
You might find useful information using the --verbose option
 ```
</details>

<details>
<summary>Output with `--verbose`:</summary>

```
➜  shard-test git:(main) ✗ shards --verbose
Resolving dependencies
git ls-remote --get-url origin
Fetching https://github.com/spider-gazelle/priority-queue.git
git fetch --all --quiet
git tag --list --column=never
Tags: v1.0.1, v1.1.0
Version tags (vX.Y): 1.0.1, 1.1.0

- `priority-queue (0.2.0)` required by `shard.yml`
```
</details>

### DIFF
--- a/src/resolvers/fossil.cr
+++ b/src/resolvers/fossil.cr
@@ -213,9 +213,16 @@ module Shards
     end
 
     protected def versions_from_tags
-      capture("fossil tag list -R #{Process.quote(local_fossil_file)}")
+      tags = capture("fossil tag list -R #{Process.quote(local_fossil_file)}")
         .split('\n')
-        .compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
+
+      Log.debug { "Tags: #{tags.reject(&.empty?).join(", ")}" }
+
+      version_tags = tags.compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
+
+      Log.debug { "Version tags (vX.Y): #{version_tags.join(", ")}" }
+
+      version_tags
     end
 
     def install_sources(version : Version, install_path : String)
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -204,9 +204,16 @@ module Shards
     end
 
     protected def versions_from_tags
-      capture("git tag --list #{GitResolver.git_column_never}")
+      tags = capture("git tag --list #{GitResolver.git_column_never}")
         .split('\n')
-        .compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
+
+      Log.debug { "Tags: #{tags.reject(&.empty?).join(", ")}" }
+
+      version_tags = tags.compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
+
+      Log.debug { "Version tags (vX.Y): #{version_tags.join(", ")}" }
+
+      version_tags
     end
 
     def install_sources(version : Version, install_path : String)
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -160,6 +160,6 @@ rescue ex : Shards::ParseError
   ex.to_s(STDERR)
   exit 1
 rescue ex : Shards::Error
-  Shards::Log.error { ex.message }
+  Shards::Log.error { ex.message.to_s + "\nYou might find useful information using the --verbose option" }
   exit 1
 end
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -160,6 +160,7 @@ rescue ex : Shards::ParseError
   ex.to_s(STDERR)
   exit 1
 rescue ex : Shards::Error
-  Shards::Log.error { ex.message.to_s + "\nYou might find useful information using the --verbose option" }
+  msg = Shards::Log.level == Log::Severity::Debug ? "" : "\nYou might find useful information using the --verbose option"
+  Shards::Log.error { ex.message.to_s + msg }
   exit 1
 end
--- a/src/resolvers/hg.cr
+++ b/src/resolvers/hg.cr
@@ -221,10 +221,17 @@ module Shards
     end
 
     protected def versions_from_tags
-      capture("hg tags --template #{Process.quote("{tag}\n")}")
+      tags = capture("hg tags --template #{Process.quote("{tag}\n")}")
         .lines
         .sort!
-        .compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
+
+      Log.debug { "Tags: #{tags.reject(&.empty?).join(", ")}" }
+
+      version_tags = tags.compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
+
+      Log.debug { "Version tags (vX.Y): #{version_tags.join(", ")}" }
+
+      version_tags
     end
 
     def install_sources(version : Version, install_path : String)


straight-shoota:
Having this additional information is certainly helpful.
However, I'm wondering if this is the best way to make it available.

The initial error message could already contain some diagnostic information to help the user directly, instead of just guiding to run again with `--verbose`. IMO "run again with --verbose" is a UX anti-pattern and should be avoided if possible. It requires manual interaction which is at best annoying in an interactive shell session, but can be really hard to deal with in automated environments (such as CI).
If we have the ability and knowledge to provide some helpful information directly, let's try to do that without another round trip.

We know which shard failed to install, so we could show some basic information such as available versions. Maybe filter/highlight ones that are similar to the missing version, to help against typos?

I expect the verbose tag output with `--verbose` could be quite polluted when you have a couple of dependencies with hundreds of tags (wether they are releases or not). And I'm not sure it would be that helpful to have this expansive information about *all* dependencies, when you're probably only interesting in one (which failed to resolve).
So perhaps a better way might be listing available versions per shard? This would fit well with a `shards info` command (#86).

beta-ziliani:
I talked to my past self, and I realized that the `--verbose` solution was a quick hack that I should've mentioned. Getting the relevant information required a bit more of work, which I did as a POC in e87c8c0.

Now it produces the following output:

```
Resolving dependencies
Fetching https://github.com/crystal-lang/crystal-molinillo.git

- `molinillo (0.3.0)` required by `shard.yml`
For molinillo the available tags are: [v0.1.0, v0.2.0]
```

There's currently no filtering, so for shards with a large list of tags it will print everything. Note also the following: this output doesn't help to the problem in #521:

```
- `molinillo (v0.2.0)` required by `shard.yml`
For molinillo the available tags are: [v0.1.0, v0.2.0]
```

I guess we need to filter the version before even reaching molinillo: if a requirement starts with `v`, fail altogether since it won't ever succeed.

straight-shoota:
I like moving in this direction 👍 

It might still be a bit much to always print *all* tags, though. An overload of information is probably not very helpful.

Maybe a first step could be to filter only tags that are valid versions.
Even that can be quite a lot. The maximum number of versions known to shardbox is [208](https://shardbox.org/shards/psykube/releases).

And then maybe some `Did you mean?` suggestions could be a good idea. Levensthein distance might actually work reasonably well to suggest similar versions. Although some tweaking might help to account for the special semantics of version numbers.

beta-ziliani:
> It might still be a bit much to always print all tags, though. An overload of information is probably not very helpful.

Yes, I mention as a problem that there's no filtering.

But, as said, this doesn't solve the original problem, and the output is as confusing as before.

beta-ziliani:
A bit hackish to say the least, but at least it produces a better output:

 * If the version is wrong in the .yml file (`v0.2.0` instead of `0.2.0`):

```
- `molinillo (v0.2.0)` required by `shard.yml`
For molinillo the closest available tag to vv0.2.0 is: v0.2.0
```

 * If the version is not listed (`0.3.0`):
 
 ```
 - `molinillo (0.3.0)` required by `shard.yml`
For molinillo the closest available tag to v0.3.0 is: v0.2.0
```

 * Is the version is not fixed (`>=0.3.0`):
 
 ```
 - `molinillo (>=0.3.0)` required by `shard.yml`
For molinillo the last available tags are v0.2.0, v0.1.0
 ```
 
 I'm certain this won't cover all possible scenarios, but at least it will improve the situation.

File: src/commands/command.cr [hunk 1]
```diff
@@ -73,8 +74,30 @@ module Shards
...
+          "For #{k} the last available tags are #{tags.reverse.first(5).join(", ")}"
+        end
+      end
+    end
+
+    def handle_resolver_errors(solver, &)
       yield
+    rescue e : Molinillo::VersionConflict(Shards::Dependency, Shards::Spec)
+      Log.error { e.message }
+      Log.error { log_available_tags(e.conflicts) }
```
straight-shoota, line 99: **polish:** This should probably be a single error message. The conflict information is just context for the version conflict error, not itself another error.

src/commands/command.cr
```diff
@@ -73,8 +74,30 @@ module Shards
...
+    private def log_available_tags(conflicts)
+      conflicts.join(separator: "\n") do |k, v|
+        req = v.requirement
+        tags = req.resolver.available_tags
+        req = req.requirement
+
+        if req.is_a?(Version) || (req.is_a?(VersionReq) && req.patterns.size == 1 && req.patterns[0] !~ /^(<|>|=)/)
+          req = "v" + req.to_s
+          found = Levenshtein.find(req, tags, 6)
+          "For #{k} the closest available tag to #{req} is: #{found}"
```
straight-shoota, line 86: **suggestion:** This branch is about version restrictions. We should not confuse that with tags.
It's an implementation detail that versions are represented as tags in the version control systems.

So the base set should be `available_releases` instead of `available_tags`. For a version restriction we don't care about all the tags, only the versions.

Also, I believe we need to explicitly handle the case that no versions are available.

src/resolvers/fossil.cr
```diff
@@ -212,10 +212,18 @@ module Shards
       end
     end
 
+    def available_tags : Array(String)
+      tags = capture("fossil tag list -R #{Process.quote(local_fossil_file)}")
+        .split('\n')
+
+      tags.reject(&.empty?)
```
straight-shoota, line 219: **suggestion:** We can directly remove empty entries from `split`.
```suggestion
        .split('\n', remove_empty: true)
```

Ditto below.

Btw. `String#lines` might be more appopriate here, but it doesn't have `remove_empty` (see crystal-lang/crystal#15149). We could still use this with `reject!(&.empty)` though.

src/commands/command.cr
beta-ziliani: Do you mean simply replacing tags for release in messages and function names?

straight-shoota: The main change would be to use `resolver.available_releases` instead of `resolve.available_tags` as source of available items for this branch. And then adapt the wording as well, of course.

I can make a patch against your branch to demonstrate what I mean.

beta-ziliani: But that won't work for what we want to do here. If you filter already the tags, then you don't get information to display the error

beta-ziliani: I'll think of an alternative

beta-ziliani: I use releases but also show tags

src/commands/command.cr
straight-shoota: If something depends on the full exact text of the error message, it should be expected to break.
We cannot possible guarantee stability on that.

src/resolvers/fossil.cr
```diff
@@ -212,10 +212,13 @@ module Shards
       end
     end
 
+    def available_tags : Array(String)
+      tags = capture("fossil tag list -R #{Process.quote(local_fossil_file)}")
```
Sija, line 216:
```suggestion
      capture("fossil tag list -R #{Process.quote(local_fossil_file)}")
```

src/resolvers/git.cr
```diff
@@ -203,10 +203,13 @@ module Shards
       end
     end
 
+    def available_tags : Array(String)
+      tags = capture("git tag --list #{GitResolver.git_column_never}")
```
Sija, line 207:
```suggestion
      capture("git tag --list #{GitResolver.git_column_never}")
```

src/resolvers/git.cr
```diff
@@ -203,10 +203,13 @@ module Shards
...
+        .split('\n', remove_empty: true)
+    end
+
     protected def versions_from_tags
-      capture("git tag --list #{GitResolver.git_column_never}")
-        .split('\n')
-        .compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
+      tags = available_tags.compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
```
Sija, line 212:
```suggestion
      available_tags.compact_map { |tag| Version.new($1) if tag =~ VERSION_TAG }
```

src/resolvers/fossil.cr
straight-shoota: This PR is still a draft while we're trying to figure out the intended behaviour. Commenting on superficial issues is just noise in this stage.
Let's focus on the contents first.

Sija: Good point, sorry for the noise.

src/commands/command.cr
beta-ziliani: Mmm... this is harder than thought:
 * Version is used in very specific cases, and I don't think a levenshtein diff will be useful
 * VersionReq has the "just this version" case which follows the patter of Version above. This is the only useful case.
 * Any other VersionReq is hard to comply to "close match". What is close match of ">0.3.0, <0.3.0"?
 * A failing branch fails at checkin, not at resolving, so won't be part of this case.
 * Tags never fail??

So I'm tempted to leave it as is for now.

beta-ziliani:
I wasn't considering properly the case when the failure is a conflict with different shards. I guess there we need to resort to the original error message, that is, filter the case when there's only one conflict.

straight-shoota:
I'm having a hard time wrapping my head around what's the changes are doing. And especially _why_ things are implemented the way they are.
Part of the reason is certainly that I'm not well versed in the codebase. And that we're missing specs for this behaviour entirely doesn't make it easier to understand the effects.

But it's inherently quite a challenge to be aware of all the invariants in the code. For example, keep track what are the conditions that lead to each of the variations of error messages.
I think it would be helpful and straightforward improvement to ease reviewing and future work on this code if we could code comments explaining the reasoning of the code. Context and rationale are quite often not self explanatory, unfortunately.

@beta-ziliani Let me know if you want me to take this.

Issue #628, opened 2024-04-24
`shards install --local` doesn't recognize cached shards


straight-shoota:
When a git repository is cached, shards should use that and only that with `--local` (no network requests). This does not seem to work, however. The cache contains bare repositories (only the git metadata without a worktree) which causes some issues with git commands.

In https://github.com/crystal-lang/shards/issues/611#issuecomment-2073641481 I noticed a particular issue with `GitResolver#origin_url` which runs `git ls-remote --get-url origin` in the cache directory. But git doesn't know its a bare repository and searches for the next `.git` folder up in the directory tree.
If the cache path is in the work tree of another git repository, it'll result in this repository's origin, which is obviously not what we're looking for.
If it's not in the work tree of another git repository (which should typically be the case with the default `SHARDS_CACHE_PATH`), the result is `origin`. This is a bit surprising and also not what what we're looking for.
So I'm wondering how (or if) `origin_url` even works in the first place?

This should technically be easy to solve by adding the flag `--bare` to let git now it's working on a bare repository. However we need to understand a bit more why this issue is happening and why it's not a bigger problem. I'm confused how other git commands seem to work with a bare repository just fine, even though they should be affected in a similar way.

I presume a similar issue might exist for other resolvers as well.

ysbaddaden:
See https://github.com/crystal-lang/shards/commit/adefcf3a725ec77961154510b1a4cb6c8ac96cbc

The origin changed, and goes to delete the cloned repository to clone the new origin. Hence delete + clone + fail because we specified `--local`.

ysbaddaden:
There are no issues with the bare repositories in the standard cache path (`~/.cache/shards`). The problem is that Git likely becomes confused by the presence of a `.git` folder in a parent folder when the cache is set to the

We probably need to specify `GIT_DIR` or `--git-dir` to specify the path.

straight-shoota:
Yeah, I figure we can even drop `run_in_folder` for git commands and just configure the git dir to point `local_path`.

straight-shoota:
> The problem is that Git likely becomes confused by the presence of a `.git` folder in a parent folder when the cache is set to be in the subdirectory of a Git project.

Yes, that seems to be it. Only if `git` cannot determine it's in a work tree (i.e. neither the curretn directory nor any of its parents have a `.git` folder) it tries to treat the current directory as a bare repo, even without `--bare`. That's super confusing because the implicit recognition as a bare repository depends on an unrelated fact.

[...]

PR #638, opened 2024-09-03
Ensure `--local` recognizes locally cached shards 

GrantBirki:
This pull request changes how the `origin_url` method works for the Git resolver. Rather than trying to run native git commands, this method now peers into the `local_path` dir for the cached shard and directly reads from the INI `config` file which contains the origin URL. Other methods in the Git resolver do exactly this so we should lean into it here as well.

resolves: https://github.com/crystal-lang/shards/issues/628

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -1,4 +1,5 @@
 require "uri"
+require "ini"
 require "./resolver"
 require "../versions"
 require "../logger"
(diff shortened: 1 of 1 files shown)


straight-shoota:
> Other methods in the Git resolver do exactly this so we should lean into it here as well.

We're only reading `config` file directly in `valid_repository?`. I don't know the reason for this, but it appears we could just use `git config` instead. Going through the CLI is easier and safer. The git configuration format is quite powerful and supports some forms of indirection that we would need to take into account. I'd prefer to avoid parsing `config` file directly if possible. Going through the CLI should always be better.

You mentioned in https://github.com/crystal-lang/shards/issues/628#issuecomment-2325707808 that you couldn't get the expected behaviour to work. But it's not quite clear to me what didn't work and why. It seems that it *should* be possible to do this without reading the `config` file directly.

In order to understand whether this is working as expected, we should add a spec to test the behaviour.

PR #639, opened 2024-09-04
Run `git config` instead of reading `config` file manually

straight-shoota:
I don't think there's a good reason to read the config file directly. Going through the CLI is easier and safer.

ref: https://github.com/crystal-lang/shards/pull/638#issuecomment-2328312139

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -351,10 +351,7 @@ module Shards
     end
 
     private def valid_repository?
-      File.each_line(File.join(local_path, "config")) do |line|
-        return true if line =~ /mirror\s*=\s*true/
-      end
-      false
+      capture("git config --get-regexp 'remote\\..+\\.mirror'").each_line.any?(&.==("true"))
     end
 
     private def origin_url


ysbaddaden (approved):
Oh my, why did I choose to read the config file 😱

GrantBirki (approved):

src/resolvers/git.cr
```diff
@@ -351,10 +351,7 @@ module Shards
...
 
     private def valid_repository?
-      File.each_line(File.join(local_path, "config")) do |line|
-        return true if line =~ /mirror\s*=\s*true/
-      end
-      false
+      capture("git config --get-regexp 'remote\\..+\\.mirror'").each_line.any?(&.==("true"))
     end
 
     private def origin_url
```
GrantBirki, line 357: Do you think we can do something similar to `origin_url` then to fix the problem related to https://github.com/crystal-lang/shards/issues/628?

straight-shoota: Yes, I think this should work similarly.

straight-shoota:
The CI failure due to #640 masked another failure. This patch seems to be broken on Windows.

Issue #640, opened 2024-09-04
`pip install mercurial` is broken on Windows CI


straight-shoota:
https://github.com/crystal-lang/shards/actions/runs/10661704235/job/29547633885

```
Run pip install mercurial
  pip install mercurial
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"
Collecting mercurial
  Downloading mercurial-6.8.1.tar.gz (8.3 MB)
     ---------------------------------------- 8.3/8.3 MB 74.0 MB/s eta 0:00:00
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'error'
  error: subprocess-exited-with-error
  
  Getting requirements to build wheel did not run successfully.
  exit code: 1
  
  [17 lines of output]
  Traceback (most recent call last):
    File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
      main()
    File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "C:\hostedtoolcache\windows\Python\3.9.13\x64\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 118, in get_requires_for_build_wheel
      return hook(config_settings)
    File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-f47lhb64\overlay\Lib\site-packages\setuptools\build_meta.py", line 332, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=[])
    File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-f47lhb64\overlay\Lib\site-packages\setuptools\build_meta.py", line 302, in _get_build_requires
      self.run_setup()
    File "C:\Users\runneradmin\AppData\Local\Temp\pip-build-env-f47lhb64\overlay\Lib\site-packages\setuptools\build_meta.py", line 503, in run_setup
      super().run_setup(setup_script=setup_script)
[...]

PR #642, opened 2024-09-18
Revert "Run `git config` instead of reading `config` file manually"

straight-shoota:
The change is broken on Windows. This was masked by the CI failure due to #640.

> E: Failed git config --get-regexp 'remote\..+\.mirror'.

Reverts crystal-lang/shards#639

### DIFF
--- a/src/resolvers/git.cr
+++ b/src/resolvers/git.cr
@@ -351,7 +351,10 @@ module Shards
     end
 
     private def valid_repository?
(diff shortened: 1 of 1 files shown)


beta-ziliani:
But why does that command fails? is that an issue with `git`?

HertzDevil:
I think it would work on Windows if the single quotation marks are dropped

PR #644, opened 2024-10-02
Allow `scripts` and `targets` in schema to be permissive

devnote-dev:
Geode makes use of custom properties in these fields which show as invalid in the editor. In future it would probably be a good idea to also make other parts of the schema permissive.

### DIFF
--- a/docs/shard.yml.schema.json
+++ b/docs/shard.yml.schema.json
@@ -230,7 +230,7 @@
           "description": "Will be run whenever the dependency is installed or upgraded in a project that requires it"
         }
       },
-      "additionalProperties": false
+      "additionalProperties": true
     },
     "targets": {
       "type": "object",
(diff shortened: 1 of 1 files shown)


ysbaddaden:
Well, they're custom extensions that don't conform to the SPEC. The schema error is right to outline these.

Maybe the SPEC could evolve to allow for some extensions (in a future compatible way).

devnote-dev:
Geode's primary use case for `targets` is allowing a `flags:` property to be set which is passed with the command when building a target, and for `scripts` is allowing custom scripts (platform-dependent or not) to be ran, including an extension of `postinstall` for platform support.

These are of course subject to change during development so I'm not sure how the SPEC could be changed to allow extensions, besides just labeling them unsupported by Shards.

PR #663, opened 2025-01-15
Vendor lib/molinillo

straight-shoota:
Vendor the dependency code into the repository, as suggested in https://github.com/crystal-lang/crystal/pull/15344#issuecomment-2592886590. We're already doing the same in https://github.com/crystal-lang/crystal.

This simplifies the build process. All code required for building shards is contained in this repository (except for Crystal stdlib). There's no need to bootstrap the initial installation of the molinillo dependency anymore.


ysbaddaden (approved):

straight-shoota:
In an additional commit, I've removed the make recipes related to dependency installation.

With dependencies vendored in we don't need automatic dependency installation anymore. We can be sure that lib/ is always present. Upon dependency upgrade we'll need to explicitly run `shards update` but there's no reason to have a make recipe for that.

ysbaddaden (approved):

PR #664, opened 2025-01-20
Fix handle global `--help` flag

### LINKED ISSUE #649: `shards --help` invokes `install` subcommand
After #631, if the command line invocation is exactly `shards --help`, or it is `shards --help --` followed by more arguments, then Shards behaves as if `install` is the selected command and ignores the `--help` entirely. This does not seem to be intended behavior.

straight-shoota:
Resolves  #649

### DIFF
--- a/src/cli.cr
+++ b/src/cli.cr
@@ -57,6 +57,10 @@ module Shards
       opts.on("-q", "--quiet", "Decrease the log verbosity, printing only warnings and errors.") { self.set_warning_log_level }
       opts.on("-h", "--help", "Print usage synopsis.") { display_help = true }
 
+      if cli_options[0]? == "--help"
+        display_help_and_exit(opts)
+      end
+
       opts.unknown_args do |args, options|
         case args[0]? || DEFAULT_COMMAND
         when "build"


ysbaddaden:
That won't catch `shards --local --help` for example, and `shards install --help` also runs the `install` command then prints the help :disappointed:

straight-shoota:
My intent was to do a simple fix for the patch release to address the most common use case `shards --help`.
But the behaviour of `shards install --help` certainly isn't acceptable either.

I think we have the following alternative steps forward:
* Fix the regression from #631
* Revert #631 for 0.19.1 (possibly followed by a reimplementation with fixed behaviour in 0.20.0)
* Refactor the entire CLI handling code to fix it once and for all (this should be based on parsing the command _first_, and then decide on the possible options, like the compiler CLI does), and release this as 0.20.0

ysbaddaden:
I think this works: https://github.com/crystal-lang/shards/pull/665

ysbaddaden:
Refactoring would indeed be :+1: 

And having a better support for sub-commands right into OptionParser would be :heart:

straight-shoota:
Resolved by https://github.com/crystal-lang/shards/pull/665

PR #665, opened 2025-01-20
Fix: --help option

### LINKED ISSUE #649: `shards --help` invokes `install` subcommand

ysbaddaden:
Restores the behavior where `--help` would always skip the command to be invoked, while still allowing to pass the `--help` to external subcommands:

The following commands print the help message and exit:

- `shards --help`
- `shards --local --help`
- `shards update --help`

While the following tries to call `shards-unknown --help` then falls back to print the help message and exit:

- `shards unknown --help` 

Relies on a fixed list of the builtin command names to avoid having each `when` case starting with `display_help_and_exit(opts) if display_help`.

closes #649

### DIFF
--- a/spec/integration/help_spec.cr
+++ b/spec/integration/help_spec.cr
@@ -3,10 +3,10 @@ require "./spec_helper"
 describe "--help" do
   it "prints help and doesn't invoke the command" do
     metadata = {
-      version: "1.0.0",
+      version:      "1.0.0",
       dependencies: {
-        mock: { git: git_path("mock") }
-      }
+        mock: {git: git_path("mock")},
+      },
     }
 
     [
--- a/spec/integration/help_spec.cr
+++ b/spec/integration/help_spec.cr
@@ -1,0 +1,29 @@
+require "./spec_helper"
+
+describe "--help" do
+  metadata = {
+    version: "1.0.0",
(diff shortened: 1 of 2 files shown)


straight-shoota:
I'm not sure if the hard coded list is worth it. Ideally, every command should have its own help message. So the currently duplicated `display_help_and_exit(opts) if display_help` is just a temporary stand-in because we don't have specific messages. Eventually, it might become something like `display_help_for_install_and_exit(opts) if display_help` etc.

straight-shoota (approved):

spec/integration/help_spec.cr
```diff
@@ -0,0 +1,29 @@
...
+      "shards update --help",
+    ].each do |command|
+      with_shard(metadata) do
+        output = run command
+
+        # it printed the help message
+        output.should contain("Commands:")
+        output.should contain("General options:")
+
+        # it didn't run the default command
```
ysbaddaden, line 24:
```suggestion
        # it didn't run the command (or default command)
```

PR #673, opened 2025-05-16
Dependency cli parsing

bcardiff:
Related to #672, this PR adds the logic to parse dependencies from the cli. Still not used, but worth discussing syntax and implementation.

The `Dependency` class unfortunately is not fit to generate the `shard.yaml`. Resolver key and source are normalized. I patched it to allow this, but the solution is not clean. We could use a separate type which would be cleaner, but is odd that the `Spec` would not use that new type when parsing the `shard.yaml`.

The added spec shows the various syntax supported.

To create a proper dependency, with the shard name, we need to actually fetch the shard to resolve that. This is done in `Dependency.from_cli` while the `Dependency.parts_from_cli` is just the parsing without side effects.

In general `{resolver}:{source}` syntax is supported, but there are some shorthands for convenience.

- `github:foo/bar`
- `github:Foo/Bar@1.2.3` (which will mean `~> 1.2.3`)
- `https://github.com/foo/bar` and other known hosts like gitlab, bitbucket, and codeberg
- `https://github.com/Foo/Bar/commit/000000`
- `https://github.com/Foo/Bar/tree/v1.2.3` (which is interpreted as a tag since it's prefixed with `v`)
- `https://github.com/Foo/Bar/tree/some/branch`
- relative and absolute paths

Some of these are maybe controversial, feel free to discuss. 

If you want to play around you can use the following code.

```crystal
require "./src/dependency"

def t(s)
  d = Shards::Dependency.from_cli(s)
  pp! d

  puts "\nshard.yaml"
  puts(YAML.build do |yaml|
    yaml.mapping do
      d.to_shard_yaml(yaml)
    end
  end)

  puts "\nshard.lock"
  puts(YAML.build do |yaml|
    yaml.mapping do
      d.to_yaml(yaml)
    end
  end)
end

t("github:crystal-lang/crystal-db")

t("github:crystal-lang/crystal-db@0.13.0")
```

which will output

```
d # => #<Shards::Dependency:0x10482be60
 @name="db",
 @requirement=*,
 @resolver=
  #<Shards::GitResolver:0x104823200
   @local_path=
    "/Users/bcardiff/.cache/shards/github.com/crystal-lang/crystal-db.git",
   @name="unknown",
   @origin_url="https://github.com/crystal-lang/crystal-db.git",
   @source="https://github.com/crystal-lang/crystal-db.git",
   @updated_cache=true>,
 @resolver_key="github",
 @source="crystal-lang/crystal-db">

shard.yaml
---
db:
  github: crystal-lang/crystal-db

shard.lock
---
db:
  git: https://github.com/crystal-lang/crystal-db.git
d # => #<Shards::Dependency:0x10482b9b0
 @name="db",
 @requirement=Shards::VersionReq(@patterns=["~> 0.13.0"]),
 @resolver=
  #<Shards::GitResolver:0x104823200
   @local_path=
    "/Users/bcardiff/.cache/shards/github.com/crystal-lang/crystal-db.git",
   @name="unknown",
   @origin_url="https://github.com/crystal-lang/crystal-db.git",
   @source="https://github.com/crystal-lang/crystal-db.git",
   @updated_cache=true>,
 @resolver_key="github",
 @source="crystal-lang/crystal-db">

shard.yaml
---
db:
  github: crystal-lang/crystal-db
  version: ~> 0.13.0

shard.lock
---
db:
  git: https://github.com/crystal-lang/crystal-db.git
  version: ~> 0.13.0
```

### DIFF
--- a/spec/unit/dependency_spec.cr
+++ b/spec/unit/dependency_spec.cr
@@ -71,6 +71,49 @@ module Shards
       parse_dependency({foo: {git: "", tag: "rc-1.0"}}).to_s.should eq("foo (tag rc-1.0)")
       parse_dependency({foo: {git: "", commit: "4478d8afe8c728f44b47d3582a270423cd7fc07d"}}).to_s.should eq("foo (commit 4478d8a)")
     end
+
+    it ".parts_from_cli" do
+      # GitHub short syntax
+      Dependency.parts_from_cli("github:foo/bar").should eq({resolver_key: "github", source: "foo/bar", requirement: Any})
+      Dependency.parts_from_cli("github:Foo/Bar@1.2.3").should eq({resolver_key: "github", source: "Foo/Bar", requirement: VersionReq.new("~> 1.2.3")})
+
+      # GitHub urls
+      Dependency.parts_from_cli("https://github.com/foo/bar").should eq({resolver_key: "github", source: "foo/bar", requirement: Any})
+      Dependency.parts_from_cli("https://github.com/Foo/Bar/commit/000000").should eq({resolver_key: "github", source: "Foo/Bar", requirement: GitCommitRef.new("000000")})
+      Dependency.parts_from_cli("https://github.com/Foo/Bar/tree/v1.2.3").should eq({resolver_key: "github", source: "Foo/Bar", requirement: GitTagRef.new("v1.2.3")})
+      Dependency.parts_from_cli("https://github.com/Foo/Bar/tree/some/branch").should eq({resolver_key: "github", source: "Foo/Bar", requirement: GitBranchRef.new("some/branch")})
+
+      # GitLab short syntax
+      Dependency.parts_from_cli("gitlab:foo/bar").should eq({resolver_key: "gitlab", source: "foo/bar", requirement: Any})
+
+      # GitLab urls
+      Dependency.parts_from_cli("https://gitlab.com/foo/bar").should eq({resolver_key: "gitlab", source: "foo/bar", requirement: Any})
+
+      # Bitbucket short syntax
+      Dependency.parts_from_cli("bitbucket:foo/bar").should eq({resolver_key: "bitbucket", source: "foo/bar", requirement: Any})
+
+      # bitbucket urls
+      Dependency.parts_from_cli("https://bitbucket.com/foo/bar").should eq({resolver_key: "bitbucket", source: "foo/bar", requirement: Any})
+
+      # Git convenient syntax since resolver matches scheme
+      Dependency.parts_from_cli("git://git.example.org/crystal-library.git").should eq({resolver_key: "git", source: "git://git.example.org/crystal-library.git", requirement: Any})
+
+      # Local paths
+      local_absolute = File.join(tmp_path, "local")
+      local_relative = File.join("spec", ".repositories", "local") # rel_path is relative to integration spec
+      Dir.mkdir_p(local_absolute)
+
+      # Path short syntax
+      Dependency.parts_from_cli(local_absolute).should eq({resolver_key: "path", source: local_absolute, requirement: Any})
+      Dependency.parts_from_cli(local_relative).should eq({resolver_key: "path", source: local_relative, requirement: Any})
+
+      # Path resolver syntax
+      Dependency.parts_from_cli("path:#{local_absolute}").should eq({resolver_key: "path", source: local_absolute, requirement: Any})
+      Dependency.parts_from_cli("path:#{local_relative}").should eq({resolver_key: "path", source: local_relative, requirement: Any})
+
+      # Other resolvers short
+      Dependency.parts_from_cli("git:git://git.example.org/crystal-library.git").should eq({resolver_key: "git", source: "git://git.example.org/crystal-library.git", requirement: Any})
+    end
   end
 end
 
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -7,8 +7,29 @@ module Shards
     property name : String
     property resolver : Resolver
     property requirement : Requirement
+    # resolver's key and source are normalized. We preserve the key and source to be used
+    # in the shard.yml file in these field. This is used to generate the shard.yml file
+    # in a more human-readable way.
+    # A Dependency can still be created without them, but it will not be possible to
+    # generate the shard.yml file.
+    property! resolver_key : String
+    property! source : String
 
-    def initialize(@name : String, @resolver : Resolver, @requirement : Requirement = Any)
+    def initialize(@name : String, @resolver : Resolver, @requirement : Requirement = Any, @resolver_key : String? = nil, @source : String? = nil)
+    end
+
+    # Parse a dependency from a CLI argument
+    def self.from_cli(value : String) : Dependency
+      parts = parts_from_cli(value)
+
+      # We need to check the actual shard name to create a dependency.
+      # This requires getting the actual spec file from some matching version.
+      resolver = Resolver.find_resolver(parts[:resolver_key], "unknown", parts[:source])
+      version = resolver.versions_for(parts[:requirement]).first || raise Shards::Error.new("No versions found for dependency: #{value}")
+      spec = resolver.spec(version)
+      name = spec.name || raise Shards::Error.new("No name found for dependency: #{value}")
+
+      Dependency.new(name, resolver, parts[:requirement], parts[:resolver_key], parts[:source])
     end
 
     # :nodoc:
@@ -120,11 +141,22 @@ module Shards
       end
     end
 
+    # Used to generate the shard.lock file.
     def to_yaml(yaml : YAML::Builder)
       yaml.scalar name
       yaml.mapping do
         yaml.scalar resolver.class.key
         yaml.scalar resolver.source
+        requirement.to_yaml(yaml)
+      end
+    end
+
+    # Used to generate the shard.yml file.
+    def to_shard_yaml(yaml : YAML::Builder)
+      yaml.scalar name
+      yaml.mapping do
+        yaml.scalar resolver_key
+        yaml.scalar source
         requirement.to_yaml(yaml)
       end
     end
--- a/src/dependency.cr
+++ b/src/dependency.cr
@@ -11,6 +11,82 @@ module Shards
     def initialize(@name : String, @resolver : Resolver, @requirement : Requirement = Any)
     end
 
+    # :nodoc:
+    #
+    # Parse the dependency from a CLI argument
+    # and return the parts needed to create the proper dependency.
+    #
+    # Split to allow better unit testing.
+    def self.parts_from_cli(value : String) : {resolver_key: String, source: String, requirement: Requirement}
+      resolver_key = nil
+      source = ""
+      requirement = Any
+
+      if File.directory?(value)
+        resolver_key = "path"
+        source = value
+      end
+
+      if value.starts_with?("https://github.com")
+        resolver_key = "github"
+        uri = URI.parse(value)
+        source = uri.path[1..-1] # drop first "/""
+
+        components = source.split("/")
+        case components[2]?
+        when "commit"
+          source = "#{components[0]}/#{components[1]}"
+          requirement = GitCommitRef.new(components[3])
+        when "tree"
+          source = "#{components[0]}/#{components[1]}"
+          requirement = if components[3].starts_with?("v")
+                          GitTagRef.new(components[3])
+                        else
+                          GitBranchRef.new(components[3..-1].join("/"))
+                        end
+        end
+      end
+
+      if value.starts_with?("https://gitlab.com")
+        resolver_key = "gitlab"
+        uri = URI.parse(value)
+        source = uri.path[1..-1] # drop first "/""
+      end
+
+      if value.starts_with?("https://bitbucket.com")
+        resolver_key = "bitbucket"
+        uri = URI.parse(value)
+        source = uri.path[1..-1] # drop first "/""
+      end
+
+      if value.starts_with?("git://")
+        resolver_key = "git"
+        source = value
+      end
+
+      unless resolver_key
+        Resolver.resolver_keys.each do |key|
+          key_schema = "#{key}:"
+          if value.starts_with?(key_schema)
+            resolver_key = key
+            source = value.sub(key_schema, "")
+
+            # narrow down requirement
+            if source.includes?("@")
+              source, version = source.split("@")
+              requirement = VersionReq.new("~> #{version}")
+            end
+
+            break
+          end
+        end
+      end
+
+      raise Shards::Error.new("Invalid dependency format: #{value}") unless resolver_key
+
+      {resolver_key: resolver_key, source: source, requirement: requirement}
+    end
+
     def self.from_yaml(pull : YAML::PullParser)
       mapping_start = pull.location
       name = pull.read_scalar
--- a/src/resolvers/resolver.cr
+++ b/src/resolvers/resolver.cr
@@ -103,6 +103,10 @@ module Shards
     private RESOLVER_CLASSES = {} of String => Resolver.class
     private RESOLVER_CACHE   = {} of ResolverCacheKey => Resolver
 
+    def self.resolver_keys
+      RESOLVER_CLASSES.keys
+    end
+
     def self.register_resolver(key, resolver)
       RESOLVER_CLASSES[key] = resolver
     end


spec/unit/dependency_spec.cr
```diff
@@ -71,6 +71,49 @@ module Shards
...
+      # Path short syntax
+      Dependency.parts_from_cli(local_absolute).should eq({resolver_key: "path", source: local_absolute, requirement: Any})
+      Dependency.parts_from_cli(local_relative).should eq({resolver_key: "path", source: local_relative, requirement: Any})
+
+      # Path resolver syntax
+      Dependency.parts_from_cli("path:#{local_absolute}").should eq({resolver_key: "path", source: local_absolute, requirement: Any})
+      Dependency.parts_from_cli("path:#{local_relative}").should eq({resolver_key: "path", source: local_relative, requirement: Any})
+
+      # Other resolvers short
+      Dependency.parts_from_cli("git:git://git.example.org/crystal-library.git").should eq({resolver_key: "git", source: "git://git.example.org/crystal-library.git", requirement: Any})
```
ysbaddaden, line 115: **Thought**: I wouldn't invent a new syntax, but reuse existing ones such as `git+<url>`. I'd still prefer have an explicit `--git` arg but :shrug:

ysbaddaden: Oh, I only realize it follows the shorthand syntax...

src/dependency.cr
```diff
@@ -7,8 +7,105 @@ module Shards
...
+
+      unless resolver_key
+        Resolver.resolver_keys.each do |key|
+          key_schema = "#{key}:"
+          if value.starts_with?(key_schema)
+            resolver_key = key
+            source = value.sub(key_schema, "")
+
+            # narrow down requirement
+            if source.includes?("@")
```
ysbaddaden, line 96: What about a SSH URL? `git@host.com/user/repo.git` ?

ysbaddaden:
I think I'd still go the other way around: from a CLI arg that directly targets a resolver and have it parse the URL into a dependency, and fallback to iterate each resolver to parse the URL for known hosts when no CLI arg is specified, and fail asking for an explicit resolver when we can't know what the remote is.

straight-shoota:
I would go the other way: encode the full information in a single URI.
This should work well. Most of the time, the natural formats are already URI-compatible.

That makes the implementation of different strategies quite simple: We parse the value as a URI and route it to a resolver-specific parser based on the scheme (e.g. `git`, `github`, `git+https`, etc. go to `GitResolver`).

If the URI scheme does not indicate a specific resolver (e.g. `https`), we fall back to heuristics based on the full URI (e.g. based on known hostname, or file extension).

src/dependency.cr
```diff
@@ -7,8 +7,105 @@ module Shards
...
+          source = "#{components[0]}/#{components[1]}"
+          requirement = GitCommitRef.new(components[3])
+        when "tree"
+          source = "#{components[0]}/#{components[1]}"
+          requirement = if components[3].starts_with?("v")
+                          GitTagRef.new(components[3])
+                        else
+                          GitBranchRef.new(components[3..-1].join("/"))
+                        end
+        end
```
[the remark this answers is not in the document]
straight-shoota: If you mean to not implement any advanced features, I agree. It might be best to focus on an MVP implementation first.
We can then introduce enhancements incrementally. And make sure they're consistent across resolvers.
We should take extensibility into account from the beginning, but don't need to specify and implement all details.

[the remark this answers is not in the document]
bcardiff: Ok, dropping them for now. I don't mind adding gitlab and others. For self-hosted ones if there is a way to identify the kind of service we can even support those.

spec/unit/dependency_spec.cr
bcardiff: Yeah, it's the `{resolver}:{source}` case, for git it looks ugly. So actually `git://...` is the short syntax.

I prefer single string so the add command will be simpler to use.

src/dependency.cr
```diff
@@ -7,8 +7,105 @@ module Shards
...
+    # Parse the dependency from a CLI argument
+    # and return the parts needed to create the proper dependency.
+    #
+    # Split to allow better unit testing.
+    def self.parts_from_cli(value : String) : {resolver_key: String, source: String, requirement: Requirement}
+      resolver_key = nil
+      source = ""
+      requirement = Any
+
+      if File.directory?(value)
```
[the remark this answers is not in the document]
bcardiff: Ok, I will leave the following
- `./`  (it's more common to have relative paths than absolute probably either way)
- `file://...`

resolver:source will work
- `path:...`

src/dependency.cr
bcardiff: Is `git@` always the prefix?

bcardiff:
For GitHub, from the clone menu we should support

- https://github.com/owner/repo.git
- git@github.com:owner/repo.git

I want to support the most common format the user can copy paste directly.

And I want to preserve as much as possible user intent. That's why I don't want the `Resolver` to offer a canonical representation of the dependency for the shard.yaml. Eg: urls are downcased in the normalization and that will look odd.

I saw mentions of `git+ssh` somewhere. IIUC those will be `git@github.com:owner/repo.git`, am I missing another pattern?

Thanks for the feedback.

bcardiff:
Oh and should I move parsing and rendering to a DependencyDefinition type to avoid having these changes in dependency but not used in all places?

src/dependency.cr
straight-shoota: The username can be anything. `git` is just a common name used with key-based authentication.

straight-shoota: The fully-qualified URI would be `git:user@host/path`.
We can use the `git@` prefix as a heuristic for the git resolver, though.

straight-shoota:
We could probably turn the named tuple return type from `.parts_from_cli` into a dedicated type which holds this information.

bcardiff:
I think I incorporated all of the feedback.

src/dependency_definition.cr
```diff
@@ -0,0 +1,114 @@
...
+        resolver_key = "gitlab"
+        uri = URI.parse(value)
+        source = uri.path[1..-1].rchop(".git") # drop first "/""
+      end
+
+      if value.starts_with?("https://bitbucket.com")
+        resolver_key = "bitbucket"
+        uri = URI.parse(value)
+        source = uri.path[1..-1] # drop first "/""
+      end
```
straight-shoota, line 79: **issue:** These conditions are insufficient because they also accept a partial matches on the host name (example: `https://github.com.ua`).
I think it would be easier to use `URI.parse` and then apply simple mappings from hostname to resolver.

```cr
KNOWN_PROVIDERS = { # this could be merged with GitResolver::KNOWN_PROVIDERS
      "www.github.com" => "github",
      "github.com" => "github",
      "www.bitbucket.com" => "bitbucket",
      "bitbucket.com" => "bitbucket",
      "www.gitlab.com" => "gitlab",
      "gitlab.com" => "gitlab",
      "www.codeberg.org" => "codeberg",
      "codeberg.org" => "codeberg",
}

uri = URI.pase(value)
if resolver_key = KNOWN_PROVIDERS[uri.host]?
  source = uri.path
end
```

spec/unit/dependency_definition_spec.cr
```diff
@@ -0,0 +1,61 @@
...
+
+module Shards
+  describe DependencyDefinition do
+    it ".parts_from_cli" do
+      # GitHub short syntax
+      expect_parses("github:foo/bar", "github", "foo/bar", Any)
+      expect_parses("github:Foo/Bar@1.2.3", "github", "Foo/Bar", VersionReq.new("~> 1.2.3"))
+
+      # GitHub urls
+      expect_parses("https://github.com/foo/bar", "github", "foo/bar", Any)
```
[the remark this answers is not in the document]
bcardiff: For me closer to the intent is to have `github: foo/bar`, because I'm assuming the user is copy-pasting the url in the browser. We do preserve the format with training .git as those are copied from the clone popup.

At some point maybe it's worth configuring shards to map all github source to be resolved in some specific way, but is a separate story.

spec/unit/dependency_definition_spec.cr
```diff
@@ -0,0 +1,61 @@
...
+
+      # Local paths
+      local_absolute = "/an/absolute/path"
+      local_relative = "an/relative/path"
+
+      # Path short syntax
+      expect_parses("./#{local_relative}", "path", "./#{local_relative}", Any)
+      expect_parses("../#{local_relative}", "path", "../#{local_relative}", Any)
+      expect_parses(".\\relative\\windows", "path", ".\\relative\\windows", Any)
+      expect_parses("..\\relative\\windows", "path", "..\\relative\\windows", Any)
```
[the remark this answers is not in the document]
bcardiff: I am not sure how shards work on windows I just didn't want to ignore it.

Can a shard be shared between windows and linux at all? Yeah, maybe we need to normalize it.

So `Path[input].to_posix` is what we should do here?
Or do you prefer to leave windows out for now?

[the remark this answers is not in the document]
straight-shoota: > Can a shard be shared between windows and linux at all? 

Of course. Why wouldn't it?
`path` resolver is often used only locally (e.g. for overrides), where interop doesn't matter much.
But there are use cases for sharing `path` configurations (e.g. https://github.com/athena-framework/athena/blob/c8aa585218db5bc64fa3cdb61624637e0af9a6a5/shard.dev.yml).
So we should try to make them portable, which means using posix-style paths when possible.

Windows and Posix paths have lots of incompatibilities (e.g. absolute paths), so there will be issues. But most of the type the paths should be expected to be relative, so those are rather edge cases.
Maybe an ideal solution would check if a Windows path can be safely turned into a posix path. But a blunt conversion should be fine as well.
We can just document that `crystal add` does not support some edge cases such as absolute paths on Windows.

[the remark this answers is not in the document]
bcardiff: Ok. I missed that posix path works on windows. It's been a while.

I think the current state is enough. Let me know if not.

straight-shoota:
Following up on https://github.com/crystal-lang/shards/pull/673#issuecomment-2886164527, I propose an alternative implementation, fully based on URI parsing in #678

straight-shoota:
For reference, VCS dependencies in PIP: https://pip.pypa.io/en/stable/topics/vcs-support/

straight-shoota:
https://github.com/crystal-lang/shards/pull/678 is a patch against the branch of this PR. We should merge that in and then continue here.

bcardiff:
Commit pushed into this branch

straight-shoota:
The windows test is failing because for relative URIs we expect `.starts_with?("./"), .starts_with?("../")` to recognize a relative path by prefix.
If we want to support path's with backslash separators as well, we need to add `.starts_with?(".\\"), .starts_with?("..\\")`

src/dependency_definition.cr
```diff
@@ -0,0 +1,95 @@
...
+          Parts.new("git", uri.to_s)
+        else
+          Parts.new("git", uri.path)
+        end
+      when "git+https"
+        uri.scheme = "https"
+        Parts.new("git", uri.to_s)
+      else
+        if resolver_class = Resolver::RESOLVER_CLASSES[scheme]?
+          uri.scheme = nil
```
straight-shoota, line 80: **suggestion:** We can further generalize and simplify the implementation:
```suggestion
      else
        scheme, _, subscheme = scheme.partition('+')
        subscheme = subscheme.presence
        if Resolver.find_class(scheme)
          if uri.host.nil? || subscheme
            uri.scheme = subscheme
          end
```

straight-shoota: The current implementation is incomplete. It misses other combinations, such as `git+ssh`, or `fossil+https`. You could expand all of these as individual branches. But it should be much easier to generalize it. This will automatically support any transport provided by a resolver.

bcardiff: Ok. Merged

src/dependency_definition.cr
```diff
@@ -0,0 +1,96 @@
...
+          if uri.host.nil? || subscheme
+            uri.scheme = subscheme
+          end
+          source = uri.to_s
+          # narrow down requirement
+          requirement = Any
+          if source.includes?("@")
+            source, version = source.split("@")
+            requirement = VersionReq.new("~> #{version}")
+          end
```
straight-shoota, line 88: **issue:** Splitting on `"@"` over the entire URL is error prone. It would also match the user info portion of a URI.
`git:user@host:repo` parses into source `user` with version requirement `~> host:repo`.
This mechanism must be restricted to the path of the URI.

Even then, there's a chance for misinterpretation because `@` is a completely valid character in a URI path. Might not be likely to encounter shards at such a path, but I'm wondering if it's worth it. IMO the `@` syntax is not essential.
We could consider a different format for expressing version constraints concisely in the URL.

Nix flake uses query parameters to encode version information (e.g. `git+https://example.com/foo/bar?branch=master`)
https://nix.dev/manual/nix/2.28/command-ref/new-cli/nix3-flake#examples

I think this is not entirely elegant because query parameters have a different purpose. They're supposed to be resolved by the remote party.

The URI fragment is more elegant for this. It represents resource-internal information resolved by the local party.
Our URIs point to shard repositories, that's the resource. And navigating to a specific portion of a resource is the purpose of a fragment.

So URLs would look like this: `github:Foo/Bar#1.2.3`, `git+https://example.com/foo/bar#1.2.3`.

We can further enhance this by adding support for parameters to the fragment, thus being able to resolve fragments like `#branch=master`, or `#tag=1.2.3-rc1`.

bcardiff: Wouldn't that # force the need of quoting in the shell?

straight-shoota: No. You can use URLs with fragments perfectly fine in shell commands.

The shell considers `#` as a comment only if it's preceded by whitespace.
```console
$ echo foo#bar
foo#bar
$ echo foo #bar
foo
```

bcardiff: Using fragments for version numbers with the same semantics currently have @ seems good to me.

Tag and branch (and revision/commit) can come later imo.

ysbaddaden: It may work, but other package managers (cargo, bundler, ...) use `@version`, so using something else like `#version` is confusing, and `#` has a special meaning in shells (comment), which only increases the confusion.

luislavena: The only issue I see with using `@` for indicating the version is when using HTTP authentication for the repository (not a parsing issue, but a visual one):

```crystal
URI.parse("https://user:pass@server:8080/path@version") # =>  #<URI:0xffff87009f60 @scheme="https", @host="server", @port=8080, @path="/path@version", @query=nil, @user="user", @password="pass", @fragment=nil>
```

OT: this is a problem I found Using Bundler/RubyGems (and moving the user credentials to environment variables to avoid having those in the Gemfile and lockfiles).

URI fragment shouldn't present any issue and could make things easier to parse:

```crystal
URI.parse("https://github.com/luislavena/drift.git#branch=master") #=>  #<URI:0xffff87009f00 @scheme="https", @host="github.com", @port=nil, @path="/luislavena/drift.git", @query=nil, @user=nil, @password=nil, @fragment="branch=master">
```

Which then requires parsing for the fragment.

But I would agree with @ysbaddaden comment: the usage of `@` for the version is the one of least surprise: I want project X _at_ version 1.2.3, and in that case, `@` translate cleanly into that.

bcardiff: I like more `@` than `#` but I don't feel strongly. The only thing that makes me doubt is that `@version` could be expected to be `= version` instead of `~> version`, because the _at_ read.

`#` is more tidy for treating things as url. but then `github:repo/owner` is not exactly a url. I find `github:repo/owner@version` more intuitive than `github:repo/owner#version` but we need to pick one symbol.

It seems that `#` is a cleaner implementation than `@` because we can rely on URI parsing, but there is no ambiguity, we can split `@` in the last path component. I don't think we need to choose based on the implementation but rather by the UX, but again there is not much difference I think in that here.

Any final calls?

bcardiff: I think we do. I rather support github:owner/repo with version rather than all the other combinations we've been discussing. It seems that it will be more used IMO.

python pip uses `@` for versioning also, and fragment for some additional thing.

I would stick with `@` as the simplest syntax and if we need branch or other we can add `#` as an alternative later.

straight-shoota: I'm not happy about jumping to a syntax for one use case without proper consideration of other related applications. Even if it's for the most common use case.
There are many important details to consider.
For example, the format used in pip and others references a concrete version: `foo@1.2.3` means package `foo` at exactly version 1.2.3. In this proposal, however, `foo@1.2.3` implicitly means package `foo` with a restriction of `~>1.2.3`. Adopting an existing syntax but applying different semantics can be confusing to users.
I'd like us to be sure that's what we want, and we see no reasonable, better alternative.

bcardiff: The fact shards encourage semver, calver seems good enough for me to make @ or whatever syntax that mentions just a version to mean ~>.

I would leave out for now any other version operator.

I don't mean to jump. I mean to discuss and decide. I don't want to not support ~> version restriction out in this first cut.

straight-shoota: The point is not about whether it's generally reasonable to interpret a plain version as `~>`.
My point is that in all other applications that use a similar reference scheme, a plain version means exactly that version.
Introducing different semantics than what people are used from other tools causes confusion. I think we should try to avoid that.

spec/unit/dependency_definition_spec.cr
```diff
@@ -0,0 +1,63 @@
...
+
+      # Path short syntax
+      expect_parses("./#{local_relative}", "path", "./#{local_relative}", Any)
+      expect_parses("../#{local_relative}", "path", "../#{local_relative}", Any)
+      {% if flag?(:windows) %}
+        expect_parses(".\\relative\\windows", "path", "./relative/windows", Any)
+        expect_parses("..\\relative\\windows", "path", "../relative/windows", Any)
+      {% end %}
+      # Path file schema
+      expect_parses("file://#{local_relative}", "path", local_relative, Any)
```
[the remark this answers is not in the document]
beta-ziliani: If possible, we can send a warning about the preferred form, but still accept it.

Let's be nice with our users. I've used that format before myself.

beta-ziliani:
I have my opinion on the high level of it. I think it's loable to center on the user, in aspects like being able to copy-paste from the browser. Thanks Branch!

src/dependency_definition.cr
bcardiff: Ok. So for ~> it should be `github:foo/bar#~>1.2.3` ?

I think is important to support ~> so we can have a good copy paste instruction with version restriction and we should encourage ~> as much as possible.

straight-shoota: > we should encourage ~> as much as possible.

This might be diverting this discussion even more than it already is. But I'm not sure I can agree on that.
What's the reason?
I don't think neither we nor any shard author recommends version restrictions anywhere in installation instructions for any shard.

PR #676, opened 2025-05-19
Add dependency name in the logs

miry:
In certain cases, it's unclear which package causes an error during shard installation. These changes improves log messages by adding explicit package context. 
The log will now indicate which package triggered the event.

## Tophat

```
Fetching https://github.com/sija/retriable.cr.git
[retriable] git config --get remote.origin.mirror
[nbchannel] git fetch --all --quiet
[splay_tree_map] git fetch --all --quie
```

grafik

### DIFF
--- a/src/package.cr
+++ b/src/package.cr
@@ -58,15 +58,20 @@ module Shards
     end
 
     def install
-      cleanup_install_directory
+      Log.with_context do
+        Log.context.set package: name
+        Log.context.set version: report_version
 
-      # install the shard:
-      resolver.install_sources(version, install_path)
+        cleanup_install_directory
 
-      # link the project's lib path as the shard's lib path, so the dependency
-      # can access transitive dependencies:
-      unless resolver.is_a?(PathResolver)
-        install_lib_path
+        # install the shard:
+        resolver.install_sources(version, install_path)
+
+        # link the project's lib path as the shard's lib path, so the dependency
+        # can access transitive dependencies:
+        unless resolver.is_a?(PathResolver)
+          install_lib_path
+        end
       end
 
       Shards.info.installed[name] = self
--- a/src/package.cr
+++ b/src/package.cr
@@ -115,24 +115,27 @@ module Shards
 
       Dir.mkdir_p(Shards.bin_path)
 
-      spec.executables.each do |name|
-        exe_name = find_executable_file(Path[install_path], name)
-        unless exe_name
-          raise Shards::Error.new("Could not find executable #{name.inspect}")
-        end
-        Log.debug { "Install #{exe_name}" }
-        source = File.join(install_path, exe_name)
-        destination = File.join(Shards.bin_path, File.basename(exe_name))
+      Log.with_context do
+        Log.context.set package: name
+        spec.executables.each do |name|
+          exe_name = find_executable_file(Path[install_path], name)
+          unless exe_name
+            raise Shards::Error.new("Could not find executable #{name.inspect}")
+          end
+          Log.debug { "Install #{exe_name}" }
+          source = File.join(install_path, exe_name)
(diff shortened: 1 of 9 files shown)


src/package.cr
```diff
@@ -58,15 +58,20 @@ module Shards
     end
 
     def install
-      cleanup_install_directory
+      Log.with_context do
+        Log.context.set package: name
+        Log.context.set version: report_version
```
miry, line 63: I added this for debugging purposes. I haven’t identified a use case for regular users yet, but I plan to use it to detect discrepancies between the expected and installed versions.

Let me know if I should keep it or not.

straight-shoota: Not sure if it's relevant.
But if we keep it, we should merge both into a single context assignment for efficiency.

Also, is there any reason you're using explicit `set` calls instead of passing the data to `with_context`? That'd be more concise.

src/resolvers/resolver.cr
```diff
@@ -65,11 +65,14 @@ module Shards
     end
 
     def spec(version : Version) : Spec
-      if spec = load_spec(version)
-        spec.version = version
-        spec
-      else
-        Spec.new(name, version, self)
+      Log.with_context do
+        Log.context.set package: name
```
straight-shoota, line 69: **thought:** This looks odd. The package context should be set when entering the scope of a specific package (e.g. in each interation when iterating over some packages).
But this is just an instance method on a resolver. I figure we should perhaps move this to the caller scope.

straight-shoota (commented):
Some of these context assignment locations look odd.

The package context should be set when entering the scope of a specific package (e.g. in each interaction when iterating over some packages).
But often they're just an instance method on a resolver. I figure we should perhaps move some of these to the caller scopes.

src/package.cr
miry: > Also, is there any reason you're using explicit set calls instead of passing the data to with_context? That'd be more concise.

Good call - because the `with_context(**kwargs)` was introduced after crystal 1.0.0. I found there are still tests are runing against Crystal 1.0.0 that are failing. [matrix](https://github.com/crystal-lang/shards/blob/master/.github/workflows/ci.yml#L19-L20)

I thought to introduce patch for older version to support `with_context(**kwargs)`.

https://github.com/crystal-lang/crystal/pull/11517

straight-shoota: Okay, that's a good reason 😆 
I like the idea of adding a polyfill for Crystal < 1.3.
However, it would be fine to keep a separate `set` call either.

miry:
Yeah, I have introduced some `with_context` for cases where there are currently no `Log.debug` statements in place, but those helped me debug using my custom Log statements.

It made me wonder whether I should introduce more logging steps for debugging purposes or remove those extra context statements.

src/resolvers/git.cr
```diff
@@ -363,7 +366,10 @@ module Shards
 
     private def valid_repository?
       command = "git config --get remote.origin.mirror"
-      Log.debug { command }
+      Log.with_context do
+        Log.context.set package: name
+        Log.debug { command }
```
miry, line 371: My small discovery.

I found `emit` could add context inplace (e.g `Log.warn exception: e, &.emit("Oh no!", user_id: 42)`).
The normal Log does not have this feature.

For example would be nice to have consistency:
```crystal
  Log.debug(exception: e, package: name) { "Oh no!" }
  Log.debug(exception: e, package: name) &.emit("Oh no!")
```

miry:
@straight-shoota I have updated code a bit. It should have less context calls.

straight-shoota (approved):

straight-shoota:
There's now a merge conflict which needs resolution.

miry:
> There's now a merge conflict which needs resolution.

done

straight-shoota (approved):

PR #678, opened 2025-05-20
URI-based dependency parsing

straight-shoota:
This proposes an alternative implementation for #673

### DIFF
--- a/spec/unit/dependency_definition_spec.cr
+++ b/spec/unit/dependency_definition_spec.cr
@@ -35,6 +35,11 @@ module Shards
       # bitbucket urls
       expect_parses("https://bitbucket.com/foo/bar", "bitbucket", "foo/bar", Any)
 
+      # unknown https urls
+      expect_raises Shards::Error, "Cannot determine resolver for HTTPS URI" do
+        Shards::DependencyDefinition.parts_from_cli("https://example.com/foo/bar")
+      end
+
       # Git convenient syntax since resolver matches scheme
       expect_parses("git://git.example.org/crystal-library.git", "git", "git://git.example.org/crystal-library.git", Any)
       expect_parses("git@example.org:foo/bar.git", "git", "git@example.org:foo/bar.git", Any)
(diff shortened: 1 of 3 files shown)


src/dependency_definition.cr
```diff
@@ -45,59 +45,51 @@ module Shards
...
-              requirement = VersionReq.new("~> #{version}")
-            end
 
-            break
+      case scheme = uri.scheme
+      when Nil
+        case value
+        when .starts_with?("./"), .starts_with?("../")
+          Parts.new("path", Path[value].to_posix.to_s)
+        when .starts_with?("git@")
```
ysbaddaden, line 55: Edge case: it's the username to connect with, it doesn't have to be `git`.

spec/unit/dependency_definition_spec.cr
```diff
@@ -51,13 +56,19 @@ module Shards
...
+      expect_parses("file:#{local_relative}", "path", local_relative, Any)
+      expect_parses("file:#{local_absolute}", "path", local_absolute, Any)
       expect_parses("file://#{local_absolute}", "path", local_absolute, Any)
       # Path resolver syntax
       expect_parses("path:#{local_absolute}", "path", local_absolute, Any)
       expect_parses("path:#{local_relative}", "path", local_relative, Any)
       # Other resolvers short
       expect_parses("git:git://git.example.org/crystal-library.git", "git", "git://git.example.org/crystal-library.git", Any)
+      expect_parses("git+https://example.org/foo/bar", "git", "https://example.org/foo/bar", Any)
+      expect_parses("git:https://example.org/foo/bar", "git", "https://example.org/foo/bar", Any)
```
ysbaddaden, line 71: There are no specs for mercurial/hg or fossil URL.

src/dependency_definition.cr
```diff
@@ -45,59 +45,51 @@ module Shards
...
+        else
+          raise Shards::Error.new("Cannot determine resolver for HTTPS URI: #{value}")
+        end
+      when "git"
+        if uri.host
+          Parts.new("git", uri.to_s)
+        else
+          Parts.new("git", uri.path)
+        end
+      when "git+https"
```
ysbaddaden, line 75: Why treat git particularly, instead of `<resolver>+` for fossil and mercurial?

Edge cases: what about other schemes and resolver combinations, like `git+file:`, `fossil+http:`, or `hg+ssh://`?

src/dependency_definition.cr
straight-shoota: Yes of course. This is just an example for a heuristic. If the username is `git`, it's likely a git URL.

We can do similar things for other URI components. For example, if the hostname starts with `fossil.`, it's likely a fossil URL. If the path ends with `.git` it's likely a git URL.

I'm not entirely sure if we should do all those things when it's relatively simple to just provide a fully qualified URI. But it seems easy enough, shouldn't cause much issues and might improve UX in some cases.

src/dependency_definition.cr
straight-shoota: Yes, of course. That all sounds good. This patch was really just intended to demonstrate the concept as an alternative to the original approach in https://github.com/crystal-lang/shards/pull/673. It certainly isn't production ready.

straight-shoota:
Yes, the `git+https` etc. schemes are well established (cf. https://pip.pypa.io/en/stable/topics/vcs-support/, https://nix.dev/manual/nix/2.28/command-ref/new-cli/nix3-flake#url-like-syntax). We should support them all.

URIs combining a resolver name with an opaque url is also coherent with SPEC. This does not apply only to shorthands.  It's nice that you can always use the same notation as in `shard.yml` (minus the whitespace) for `shards add`.

We can easily generalize the mapping implementation and get rid of any special cases (see https://github.com/crystal-lang/shards/pull/673/files#r2102181386).
The only explicit URI handlers are for heuristics when the scheme is missing or inconclusive, and for mapping `file` to `path`.

PR #693, opened 2026-02-10
Drop spec `debug` helper

straight-shoota:
This looks like a development artifact and seems very confusing.


ysbaddaden:
Indeed. It shouldn't be used, but has been _very_ useful to understand why a command fails.

ysbaddaden (approved):

PR #726, opened 2026-09-12
Add Tangled support for dependencies

LaskaIDK:
Added [Tangled](https://tangled.org) support + tests for it


spec/unit/git_resolver_spec.cr
```diff
@@ -40,13 +40,18 @@ module Shards
       GitResolver.normalize_key_source("gitlab", "rEpo/pAth").should eq({"git", "https://gitlab.com/repo/path.git"})
       GitResolver.normalize_key_source("gitlab", "REPO/PATH").should eq({"git", "https://gitlab.com/repo/path.git"})
       GitResolver.normalize_key_source("codeberg", "REPO/PATH").should eq({"git", "https://codeberg.org/repo/path.git"})
+      GitResolver.normalize_key_source("tangled", "repo/path").should eq({"git", "https://tangled.org/repo/path.git"})
+      GitResolver.normalize_key_source("tangled", "rEpo/pAth").should eq({"git", "https://tangled.org/repo/path.git"})
+      GitResolver.normalize_key_source("tangled", "REPO/PATH").should eq({"git", "https://tangled.org/repo/path.git"})
+
```
Sija, line 46:
```suggestion
```

straight-shoota:
I don't see a justification for this. 

Named resolvers for single code hosts is a special feature that we provide as a short cut for some exceptionally common and popular services.
Tangled does not appear to be in that league (at this time).

If we add named keys for just any public code host, we'd be overwhelmed by complexity.

This features doesn't block access to tangled repositories.
It's perfectly fine to use regular `git: https://tangled.org/foo/bar` as resolver.

PR #729, opened 2026-09-16
docs(lock): add explanatory comment header to generated shard.lock

### LINKED ISSUE #195: Add explanatory comment in shard.lock
I just stumbled upon #130 which was kind of resolved by adding a note to the README.

Having a lock file is quite common with other dependency managers as well, but people who are not familiar with the concept will probably have difficulties understanding what this file is for. Even if the README and man page explain this, it would probably be a nice addition if the shard.lock file had a comment about this as well. It's a YAML file, so it's shouldn't be an issue putting a comment at the top.

renich:
Resolves #195

This PR adds a short, standard explanatory comment to the top of newly generated `shard.lock` files to explicitly notify users that the file is automatically generated and should not be manually edited. This prevents confusion for developers coming from languages or ecosystems without lockfiles.


renich:
I updated the integration test assertions in `spec/integration/install_spec.cr` and `spec/integration/update_spec.cr` to include the generated comment header. All integration and unit specs pass locally.

src/lock.cr
```diff
@@ -65,6 +65,8 @@ module Shards
     end
 
     def self.write(packages : Array(Package), override_path : String?, io : IO)
+      io << "# This file is generated by Shards.\n"
```
Sija:
```suggestion
      io << "# This file is automatically generated by Shards.\n"
```

renich: Updated to 'automatically generated' as suggested. Thanks!

src/lock.cr
```diff
@@ -65,6 +65,8 @@ module Shards
     end
 
     def self.write(packages : Array(Package), override_path : String?, io : IO)
+      io << "# This file is automatically generated by Shards.\n"
+      io << "# Please don't edit this file manually.\n"
```
straight-shoota, line 69: **question:** What's the reason for using `<<` instead of `#puts` here?

straight-shoota, line 69: **suggestion:** No need to be polite or wordy. I'd suggest to copy the phrase we use in autogenerated files in the Crystal repo:

```suggestion
      io << "# DO NOT EDIT\n"
```

Sija: For consistency with the calls below, I reckon.

